PR #4, opened 2023-09-12
Updates README.md

aleskandro:


README.md
```diff
@@ -1,34 +1,55 @@
...
-  subset of architectures that the container images support. The gate is then removed and the pod is scheduled as normal and based on the affinity scheduling predicates,
-  will land on a node with the supported architecture
+
+- **Architecture aware Pod Placement**: The pod placement operand aims to automate the 
+  inspection of the container images, derive a set of architectures supported by a pod and use it to
+  automatically define strong predicates based on the `kubernetes.io/arch` label in the pod's nodeAffinity. 
+  This operand is based on the [KEP-3521](https://github.com/kubernetes/enhancements/blob/afad6f270c7ac2ae853f4d1b72c379a6c3c7c042/keps/sig-scheduling/3521-pod-scheduling-readiness/README.md) and
+  [KEP-3838](https://github.com/kubernetes/enhancements/blob/afad6f270c7ac2ae853f4d1b72c379a6c3c7c042/keps/sig-scheduling/3521-pod-scheduling-readiness/README.md), as
+  described in the [Openshift EP](https://github.com/kubernetes/enhancements/blob/afad6f270c7ac2ae853f4d1b72c379a6c3c7c042/keps/sig-scheduling/3838-pod-mutable-scheduling-directives/README.md) introducing it.
```
Prashanth684: links are incorrect. need to be fixed

README.md
```diff
@@ -1,34 +1,55 @@
...
+  adds it as node affinity requirement to the pod spec, and remove the scheduling gate.
 
 ## Getting Started
-You’ll need a Kubernetes cluster to run against. You can use [KIND](https://sigs.k8s.io/kind) to get a local cluster for testing, or run against a remote cluster.
-**Note:** Your controller will automatically use the current context in your kubeconfig file (i.e. whatever cluster `kubectl cluster-info` shows).
 
-### Running on the cluster
-1. Install Instances of Custom Resources:
+The aim of this operator will be to run on any Kubernetes cluster, although the main focus of development and testing
+will be carried out on Openshift clusters.
```
Prashanth684, line 22: i wonder if we should even say this now. the operator is tightly coupled with Openshift as of now

aleskandro: yeah, I was thinking of that, but I wanted to add it anyway to have this target/aim in mind :).

Users of non-Openshift clusters could set the location of the previous two config maps as of a [recent commit](https://github.com/openshift/multiarch-manager-operator/commit/bb06729d906ea924f185355e053eb80383725053), and we will be able to disable the OCP-specific controllers soon.

So yes, the operator will include deps from the Openshift Eco-system, it is still untested outside of OCP and we cannot guarantee things like ICSPs/IDMSes/ITMSes on non-OCP clusters, but the feasibility/aim could be kept (with something non automated as for ocp).

Although things will probably be more complicated while moving forward, if we could just be heads up on not compromising the compatibility with non-OCP clusters, we could keep the "aim of running on any k8s".

wdyt?

README.md
aleskandro: Thanks! fixed

README.md
Prashanth684: The aim is fine - i would not advertise it until we have proved and tested it works, so maybe a cautionary note that this is not tested outside of OCP and that it is still a wip would be good

Prashanth684: > although the main focus of development and testing
will be carried out on Openshift clusters

ok you have this so we should be ok

PR #5, opened 2023-09-14
MIXEDARCH-328: Change ICache implementation to use and review storeGlobalPullSecret indirection

aleskandro:
The current caching mechanism comes from the PoC implementation but is not ready yet to be used because (1) it misses a cache eviction policy, (2) it may need a few further fixes to avoid leaking images' metadata when the inspection happens with different pull secrets.

This PR skips the `cacheProxy`'s use to rely solely on the remote registry inspector.

Moreover, the first commit improves the storeGlobalPullSecret indirection (through the facade) implementation to easily allow other ICache implementation to be used by the Facade (e.g., for unit tests).

### DIFF
--- a/pkg/image/facade.go
+++ b/pkg/image/facade.go
@@ -30,16 +30,23 @@ var (
 )
 
 type Facade struct {
-	inspectionCache ICache
+	inspectionCache       ICache
+	storeGlobalPullSecret func(pullSecret []byte)
 }
 
 func (i *Facade) GetCompatibleArchitecturesSet(ctx context.Context, imageReference string, secrets [][]byte) (architectures sets.Set[string], err error) {
 	return i.inspectionCache.GetCompatibleArchitecturesSet(ctx, imageReference, secrets)
 }
 
+func (i *Facade) StoreGlobalPullSecret(pullSecret []byte) {
+	i.storeGlobalPullSecret(pullSecret)
+}
+
 func newImageFacade() *Facade {
+	inspectionCache := newCache()
 	return &Facade{
-		inspectionCache: newCache(),
+		inspectionCache:       inspectionCache,
+		storeGlobalPullSecret: inspectionCache.storeGlobalPullSecret,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/image/interfaces.go
```diff
@@ -33,5 +33,5 @@ type IRegistryInspector interface {
 	// StoreGlobalPullSecret takes a pull secret and stores it in the ImageFacade. It will be used by the controller
```
Prashanth684: nit: change function name in comment (lower case s)

pkg/image/facade.go
```diff
@@ -30,16 +30,24 @@ var (
...
+	i.storeGlobalPullSecret(pullSecret)
+}
+
 func newImageFacade() *Facade {
+	// TODO: restore to use the cacheProxy once an eviction policy is implemented
+	inspectionCache := newRegistryInspector()
 	return &Facade{
-		inspectionCache: newCache(),
+		inspectionCache:       inspectionCache,
+		storeGlobalPullSecret: inspectionCache.storeGlobalPullSecret,
```
Prashanth684, line 50: any reason why we need this and not just directly reference as `i.inspectionCache.storeGlobalPullSecret`

Prashanth684:
lgtm

pkg/image/facade.go
aleskandro: Hi @Prashanth684, the reason is that the `inspectionCache` field is of type `ICache`. Instead, the `inspectionCache` variable is its real implementation (implementing `IRegistryInspector` that also embeds `ICache`).

Therefore, `i.inspectionCache.storeGlobalPullSecret` will not be possible. I'm delegating to the newFacade method the creation and decision about what the storeGalobalPullSecret method should do.

The idea here is that I would avoid setting public the `storeGlobalPullSecret` method in the remote image inspector and at the same time avoid defining a k8s-specific controller in that package to provide separation of concerns. 

I know I'm not there 100% yet, and this can still change in time.... 

What's more important, finally, is that we can have a specific implementation of ICache and IRegistryInspector for the unit tests.

PR #8, opened 2023-11-08
Unit tests for the pod embedding

aleskandro:
This PR introduces initial unit tests for the pod_model embedding. 

It also implements the ICache fake Facade and registryInspector structs, to allow the unit tests not to depend on the external registry.

### DIFF
--- a/controllers/podplacement/pod_model_test.go
+++ b/controllers/podplacement/pod_model_test.go
@@ -1,0 +1,771 @@
+package podplacement
+
+import (
+	"context"
+	"log"
+	"multiarch-operator/pkg/image/fake"
+	"sort"
+	"testing"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	. "github.com/onsi/gomega"
+)
+
+var ctx context.Context
+
+func init() {
+	ctx = context.TODO()
+	// Replace the facade singleton with a fake one. The init() function is called after the initialization of the
+	// facade singleton, so replacing it here is safe.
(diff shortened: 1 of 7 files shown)


pkg/image/fake/inspector.go
```diff
@@ -0,0 +1,62 @@
...
+	ArchitectureAmd64   = "amd64"
+	ArchitectureArm64   = "arm64"
+	ArchitecturePpc64le = "ppc64le"
+	ArchitectureS390x   = "s390x"
+)
+
+const (
+	SingleArchAmd64Image = "docker.io/library/single-arch-amd64-image:latest"
+	SingleArchArm64Image = "docker.io/library/single-arch-arm64-image:latest"
+	MultiArchImage       = "docker.io/library/multi-arch-image:latest"
```
Prashanth684: nit: maybe instead of `docker.io` use `myregistry.io` or something to indicate that these images are fake

aleskandro: done

controllers/podplacement/pod_model_test.go
```diff
@@ -0,0 +1,771 @@
...
+	tests := []struct {
+		name               string
+		pod                v1.Pod
+		pullSecretDataList [][]byte
+		// Be aware that the values in the want.Values slice must be sorted alphabetically
+		want    v1.NodeSelectorRequirement
+		wantErr bool
+	}{
+		{
+			name: "pod with several containers using multi-arch images",
```
Prashanth684, line 389: also probably want to to add one for images without the right permissions

PR #10, opened 2023-11-14
Integration test suite

aleskandro:
This PR adds minor fixes and an initial implementation of the integration suite that includes a registry spawn before the tests run.

The registry is based on the one in distribution/distribution, but a custom auth provider is implemented, based on the upstream htpasswd one, also to provide authorization capabilities.

Additional methods are implemented in the fake/registry package to seed the registry with fake images having different media types (OCI v1 index, Docker Manifest List, etc..), no data layers, and different authorization requirements (public, global pull secret, local, ...).

The current initial suite runs the k8s envtest, the registry, seeds the registry with fake images, seeds the fake cluster with namespaces and initial configuration. Finally, it starts the manager and runs the tests. Only an example test is added now.

Business logic changes include:
- logs text in the main
- using env vars when available to catch the paths of the system config files
- fix a data race by using a shared lock in the global pull secret handler
- added an annotation to the pod after the scheduling gate is removed: this is helpful for testing, but also debugging and to inform the users that our operator did some processing on the pod spec ~~(this needs further checks to verify it doesn't break reconciliation from other controllers)~~.


controllers/podplacement/pod_model.go
```diff
@@ -27,7 +27,8 @@ import (
 )
 
 const (
-	archLabel = "kubernetes.io/arch"
+	archLabel                = "kubernetes.io/arch"
+	schedulingGateAnnotation = "multiarch.openshift.io/scheduling-gate"
 )
```
Prashanth684, line 37: this could also be potentially useful in telemetry to find out how many pods have been through the gate. I assume this includes pods which have not been mutated as well

Prashanth684: maybe also have the label be `scheduling -gate-evaluated`

pkg/image/inspector.go
```diff
@@ -36,13 +36,16 @@ import (
...
-	mutex sync.Mutex
+	mutex sync.RWMutex
 }
 
 func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, imageReference string, secrets [][]byte) (supportedArchitectures sets.Set[string], err error) {
 	// Create the auth file
 	log := ctrllog.FromContext(ctx, "imageReference", imageReference)
-	authFile, err := i.createAuthFile(log, append([][]byte{i.globalPullSecret}, secrets...)...)
+	i.mutex.RLock()
+	globalPullSecret := i.globalPullSecret
```
Prashanth684, line 46: so this is when there is a concurrent read and write?

Prashanth684:
should the `fake/` packages go into a testing sub folder rather than with the actual logic?

pkg/image/fake/registry/mock_image.go
```diff
@@ -0,0 +1,211 @@
+package registry
```
Prashanth684, line 1: nice work with this!!

controllers/podplacement/pod_model.go
aleskandro: Right. Maybe it's worth it to add labels rather than annotations to support telemetry and some queries with `oc` for the uninstallation procedure.

pkg/image/inspector.go
aleskandro: here we read the globalPullSecret. The podReconciler usually calls this. Instead, the global pull secret syncer (Runnable) will write it. Since they run concurrently, there is a potential data race, albeit unlikely if you think the global pull secret doesn't change so often.

The syncer was already using a mutex to synchronize writes.

aleskandro:
> should the `fake/` packages go into a testing subfolder rather than with the actual logic?

Yes, I thought of that, but I would do it as a follow-up because I want to move some other packages before we continue. I want this basic unit and integration testing to move forward with further changes, reducing risks of breaking stuff.

controllers/podplacement/pod_model.go
Prashanth684: yes that sounds better. hopefully adding labels won't affect any custom controllers watching a pod

pkg/image/inspector.go
Prashanth684: understood. yes the case is rare but can happen

controllers/podplacement/pod_model.go
aleskandro: yeah, that was my concern. I see other labels and annotations being added in pods. Labels are well-known (like pod-template-hash) so some custom controllers could opt-out them specifically, whereas annotations are less common (e.g., ovn-specific annotations). The replicaset controller doesn't affect them btw (I expect the same for others core controllers).

I changed to labels for now but can revert to annotations if we are unsure of labels... See commit https://github.com/openshift/multiarch-manager-operator/pull/10/commits/c143e50bae347f01e1096b721339c89ef4da628b

Prashanth684: let us have it on by default for now. we could even think about exposing a "marked flag" as part of the configuration options when setting up the CR

aleskandro:
I excluded some testing files raising alerts in the security job. I also asked a question about how to exclude specific CWE rules for specific files in #forum-testplatform. We could change it later to unblock some wip followups if the response will require time.

/approve

aleskandro:
Merging as no clear path is yet available to exclude only some rules on specific files.

Issue #12, opened 2023-12-14
Implement  Operator and controllers as distinct entities with the operator-controller model


Prashanth684:
Currently the operator is implemented just as a p-o-c style controller to validate the functionality of adding node affinities on scheduling gated pods. But this is actually just the pod placement controller. We need the operator to manage various controllers such as:

- PodPlacement controller (which is built in to the operator today)
- PodPlacementConfig controller - managing and reconciling the config CR
-  Webhook controller - managing the webhook deployments

Other potential controllers to implement:
-  ICSP/IDMS controller - controller to reconcile and store/update ICSP/IDMS config
-  ImageRegistryConfig controller - update registry configuration and update policies.json

In addition to this, the PodPlacementController should run as an active-standby deployment using leader election. The webhook will be implemented as a replicaset (3 replicas ideally)  to intercept pods and gate them as per config.

This would involve the operator/controller each having their own deployment specs with command line options for each controller to be associated with the respective deployments

- Code would have to be restructured to have an operator package and a package for each controller
- Have deployment yamls for the operator and each controller

aleskandro:
I think we can close this as #21 finalized this work. #23 will be a follow-up regarding the definition of a fine-grained FSM, also helpful for the ordered uninstallation.

PR #17, opened 2024-01-04
MIXEDARCH-325: Add flags to enable a subset of the controllers

aleskandro:
This PR adds three flags to allow enabling a subset of the controllers:
- `--enable-operator` enables the operator controllers (the PodPlacementConfig now)
- `--enable-ppc-webhook` will register the webhook to mutate pods
- `--enable-ppc-controllers` will enable the controllers related to the arch-aware pod placement

Currently, the manifests are changed to enable all the controllers. Once we go concluding #12 , we will disable the PPC-related controllers from running in operator pod.

This is related to #12 and closes [MIXEDARCH-325](https://issues.redhat.com//browse/MIXEDARCH-325)

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,10 +17,17 @@ limitations under the License.
 package main
 
 import (
+	"errors"
 	"flag"
+	"fmt"
 	"os"
 
+	"github.com/openshift/multiarch-manager-operator/controllers/operator"
 	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -32,9 +39,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/healthz"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	//+kubebuilder:scaffold:imports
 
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -63,40 +84,16 @@ func init() {
...
-	_ = flag.Set("alsologtostderr", "true")
-
-	opts.BindFlags(flag.CommandLine)
-	flag.Parse()
-
-	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
-
-	// TODO: verify after bumping the controller-runtime version
-
 	webhookServer := webhook.NewServer(webhook.Options{
```
Prashanth684, line 97: shouldn't webhook server initialization go inside the `if enablePodPlacementConfigOperandWebHook {` below ?

main.go
```diff
@@ -110,7 +107,7 @@ func main() {
 		WebhookServer:          webhookServer,
```
Prashanth684, line 107: hmm..should there be different managers for webhook and the others ?

main.go
```diff
@@ -123,94 +120,109 @@ func main() {
...
-	err = mgr.Add(podplacement.NewICSPSyncer(mgr))
-	if err != nil {
-		setupLog.Error(err, "unable to add the ICSPSyncer Runnable to the manager")
-		os.Exit(1)
-	}
+	must(mgr.Add(podplacement.NewICSPSyncer(mgr)),
+		unableToAddRunnable, runnableKey, "ICSPSyncer")
+	must(mgr.Add(podplacement.NewIDMSSyncer(mgr)),
+		unableToAddRunnable, runnableKey, "IDMSSyncer")
+	must(mgr.Add(podplacement.NewITMSSyncer(mgr)),
```
Prashanth684, line 176: hmm..what is ITMS ? :)

main.go
```diff
@@ -123,94 +120,109 @@ func main() {
...
-		os.Exit(1)
+func bindFlags() {
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&certDir, "cert-dir", "/var/run/manager/tls", "The directory where the TLS certs are stored")
+	// TODO: Change the defaults to match a local secret; the OCP specific settings will be provided by the operator
+	flag.StringVar(&globalPullSecretNamespace, "global-pull-secret-namespace", "openshift-config", "The namespace where the global pull secret is stored")
```
Prashanth684, line 204: the bindflags options also possibly need to change for each controller...at least the globalpullsecrets, registry info is not required for the webhook and the operator

main.go
aleskandro: ImageTagMirrorSet (ITMS) and ImageDigestMirrorSet (IDMS)

main.go
aleskandro: hmm, we can validate and error out if a parameter is passed and required by a controller, but the parameter to enable that controller is not set.

main.go
aleskandro: As far as I understand, each manager will need to have a webhook, in order to respond at `healthz` and `ready` for the liveness and readiness probes. What we can do is only to register the mutating webhook handler at need (based on the parameter)

main.go
aleskandro: related to https://github.com/openshift/multiarch-manager-operator/pull/17#discussion_r1444458189... I also believe we need to keep the custom webhook we instantiate in `webhookServer` to comply with the OCP certs chain.

main.go
Prashanth684: that sounds good to me.

main.go
Prashanth684: ah..ok got it...lgtm

main.go
Prashanth684: ok..let's skip validation since we have defaults.

PR #21, opened 2024-01-16
MIXEDARCH-352: Manage the lifecycle of the pod placement operand

### LINKED ISSUE #12: Implement  Operator and controllers as distinct entities with the operator-controller model

aleskandro:


apis/multiarch/v1alpha1/podplacementconfig_types.go
```diff
@@ -30,6 +30,21 @@ const (
 	LogVerbosityLevelTraceAll LogVerbosityLevel = "TraceAll"
 )
 
+func (verbosity LogVerbosityLevel) ToZapLevelInt() int {
```
Prashanth684: instead, can we use a map here? that maps the verbosity level to the int and use it directly

apis/multiarch/v1alpha1/podplacementconfig_types.go
```diff
@@ -30,6 +30,21 @@ const (
...
+	switch verbosity {
+	case LogVerbosityLevelNormal:
+		return 3
```
Prashanth684: even if it is small, i would suggest moving logging stuff to a separate helper file for improved readability

controllers/operator/objects.go
```diff
@@ -0,0 +1,287 @@
...
+								"--v=0",
+							},
+							Ports: []corev1.ContainerPort{
+								{
+									ContainerPort: 8443,
+									Protocol:      corev1.ProtocolTCP,
+									Name:          "https",
+								},
+							},
+							Resources: corev1.ResourceRequirements{
```
Prashanth684, line 222: for future considerations, cpu and memory level would need to be tuneable

controllers/operator/objects.go
```diff
@@ -0,0 +1,287 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/openshift/multiarch-manager-operator/apis/multiarch/v1alpha1"
+	"github.com/openshift/multiarch-manager-operator/pkg/utils"
+)
+
+func buildMutatingWebhookConfiguration(podPlacementConfig *v1alpha1.PodPlacementConfig) *admissionv1.MutatingWebhookConfiguration {
```
Prashanth684, line 17: is it too much overkill to use the[ deploymentcontroller](https://github.com/openshift/library-go/blob/master/pkg/operator/deploymentcontroller/deployment_controller.go) in library-go? it gives us more fine grained control that we might not need right now but maybe in the future

Prashanth684: we could use optional [hook functions](https://github.com/openshift/library-go/blob/master/pkg/operator/deploymentcontroller/deployment_controller.go#L59) to validate each part of the deployment

controllers/operator/objects.go
aleskandro: Yeah, I had the same thought while writing this. I opened #22 to remember discussing this.

apis/multiarch/v1alpha1/podplacementconfig_types.go
aleskandro: Done. For now, I'm storing the map within the function to make the map itself 'constant/read-only'. Users requiring this conversion should still call a method in the type IMHO to follow the principle of encapsulation.

apis/multiarch/v1alpha1/podplacementconfig_types.go
aleskandro: done

controllers/operator/objects.go
aleskandro: I also see this as useful as it helps defining the FSM for the ordered uninstallation in the future. Should we consider this as a follow-up and give priority to other stuff for now?

Prashanth684: let's follow up in a separate ticket.

aleskandro: #23

PR #25, opened 2024-01-17
MIXEDARCH-352: Updates README.md

aleskandro:


README.md
```diff
@@ -52,6 +52,45 @@ make docker-buildx IMG=<some-registry>/multiarch-manager-operator:tag
...
 ```
 
+### Deploy the Pod Placement Operand
+
+After the operator is running, you can deploy the Pod Placement Operand on the cluster through the PodPlacementConfig CR.
+It is expected to be a singleton, cluster-scoped resource, named `cluster`.
+
+The following is an example of a PodPlacementConfig CR that sets the log verbosity level to `Normal` and 
+will watch and setup CPU architecture node affinities on all the pods, except the ones in namespaces labeled with 
+`multiarch.openshift.io/exclude-pod-placement`.
```
Prashanth684, line 62: also should probably mention that ocp and kube namespaces will be excluded by default

aleskandro: done.

PR #29, opened 2024-02-06
fix exclude pod placement label

lwan-wanglin:
fix exclude pod placement label.

@aleskandro could you help to review?thanks


aleskandro:
/lgtm
/approve

THanks @lwan-wanglin

PR #32, opened 2024-02-07
MIXEDARCH-330: E2E test suite initial implementation

aleskandro:
This PR adds the initial implementation for a e2e test suite and 1 test case that verifies the operands are deployed and available after the creation of the pod placement config.

### DIFF
--- a/hack/ci-test.sh
+++ b/hack/ci-test.sh
@@ -44,6 +44,10 @@ if [ "$SKIP_COVERAGE" != "true" ] && [ -n "$ARTIFACT_DIR" ] && [ -d "$ARTIFACT_D
   GINKGO_ARGS="${GINKGO_ARGS} --cover --coverprofile=test-unit-coverage.out"
 fi
 
+if [ "${TEST_LABEL}" == "e2e" ]; then
+  export NAMESPACE="openshift-multiarch-operator"
+fi
+
 # Print the command we are going to run as Make would.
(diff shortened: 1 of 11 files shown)


hack/ci-test.sh
```diff
@@ -44,6 +44,10 @@ if [ "$SKIP_COVERAGE" != "true" ] && [ -n "$ARTIFACT_DIR" ] && [ -d "$ARTIFACT_D
   GINKGO_ARGS="${GINKGO_ARGS} --cover --coverprofile=test-unit-coverage.out"
 fi
 
+if [ "${TEST_LABEL}" == "e2e" ]; then
+  export NAMESPACE="openshift-multiarch-operator"
```
Prashanth684, line 48: can this not be passed in from the yaml? to set the namespace?

aleskandro: This is expected to be used by the E2e suite, for example to check if the operator's pods are running in the remote cluster. Not sure about other ways to achieve it. I mostly reused the same facility to get the namespace from either the downward api -> env vars mapping or via the ones set by the makefile targets

Prashanth684: ah..ok so this is used outside the cluster

Prashanth684: yeah i see that the namespace is already exported inside the operator here: https://github.com/openshift/multiarch-manager-operator/blob/main/config/default/default/manager_auth_proxy_patch.yaml#L53 . i was just curious why it was not done in the main container

PR #44, opened 2024-02-22
MIXEDARCH-442: Add test for nodeAffinity conflict or non conflict

lwan-wanglin:
Add test for nodeAffinity conflict or non conflict

- Add one testcase for nodeAffinity non-conflicting test
- Add one testcase for nodeAffinity conflicting test
- Add one testcase for nodeSeletor test
- Add one testcase for multiple MatchExpressions test, one has conflicting, another doesn't

### DIFF
--- a/pkg/e2e/podplacement/pod_placement_test.go
+++ b/pkg/e2e/podplacement/pod_placement_test.go
@@ -4,15 +4,14 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
-	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/multiarch-manager-operator/pkg/e2e"
+	. "github.com/openshift/multiarch-manager-operator/pkg/testing/builder"
 	"github.com/openshift/multiarch-manager-operator/pkg/testing/framework"
 	"github.com/openshift/multiarch-manager-operator/pkg/utils"
 )
@@ -28,39 +27,24 @@ var _ = Describe("The Pod Placement Operand", func() {
 			Expect(err).NotTo(HaveOccurred())
 			//nolint:errcheck
 			defer client.Delete(ctx, ns)
-			d := appsv1.Deployment{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:      "test-deployment",
-					Namespace: ns.Name,
-				},
-				Spec: appsv1.DeploymentSpec{
-					Replicas: utils.NewPtr(int32(1)),
-					Selector: &metav1.LabelSelector{
-						MatchLabels: map[string]string{
-							"app": "test",
-						},
-					},
-					Template: corev1.PodTemplateSpec{
-						ObjectMeta: metav1.ObjectMeta{
-							Labels: map[string]string{
-								"app": "test",
-							},
-						},
-						Spec: corev1.PodSpec{
-							Containers: []corev1.Container{
-								{
-									Name:  "test",
-									Image: helloOpenshiftPublicMultiarchImage,
-								},
-							},
-						},
-					},
-				},
-			}
+			ps := NewPodSpec().
+				WithContainersImages(helloOpenshiftPublicMultiarchImage).
+				Build()
+			d := NewDeployment().
+				WithSelectorAndPodLabels("app", "test").
+				WithPodSpec(ps).
+				WithReplicas(utils.NewPtr(int32(1))).
+				WithName("test-deployment").
+				WithNamespace(ns.Name).
+				Build()
 			err = client.Create(ctx, &d)
 			Expect(err).NotTo(HaveOccurred())
-			verifyPodNodeAffinity(ns, "app", "test", utils.ArchitectureAmd64,
-				utils.ArchitectureArm64, utils.ArchitectureS390x, utils.ArchitecturePpc64le)
+			archLabelNSR := NewNodeSelectorRequirement().
+				WithKeyAndValues(utils.ArchLabel, corev1.NodeSelectorOpIn, utils.ArchitectureAmd64,
+					utils.ArchitectureArm64, utils.ArchitectureS390x, utils.ArchitecturePpc64le).
+				Build()
+			expectedNSTs := NewNodeSelectorTerm().WithMatchExpressions(&archLabelNSR).Build()
+			verifyPodNodeAffinity(ns, "app", "test", expectedNSTs)
 		})
 		It("should set the node affinity on privileged deployments", func() {
 			var err error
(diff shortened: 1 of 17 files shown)


pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,135 @@ var _ = Describe("The Pod Placement Operand", func() {
 				g.Expect(pod.Annotations).NotTo(BeEmpty())
 				g.Expect(pod.Annotations).To(HaveKeyWithValue("openshift.io/scc", "privileged"))
 			})
-			verifyPodNodeAffinity(ns, "app", "test", utils.ArchitectureAmd64,
+			verifyPodNodeAffinity(ns, "app", "test", utils.ArchLabel, corev1.NodeSelectorOpIn, utils.ArchitectureAmd64,
 				utils.ArchitectureArm64, utils.ArchitectureS390x, utils.ArchitecturePpc64le)
 		})
+		It("should set the node affinity when users node affinity non conflict", func() {
+			var err error
```
Prashanth684, line 110: nit: "should set the node affinity when users node affinity non conflict"
to
"should set the node affinity when users node affinity do not conflict"

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,135 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+					},
+				},
+			}
+			err = client.Create(ctx, &d)
+			Expect(err).NotTo(HaveOccurred())
+			verifyPodNodeAffinity(ns, "app", "test", utils.OSLabel, corev1.NodeSelectorOpIn, utils.OSLinux)
+			verifyPodNodeAffinity(ns, "app", "test", utils.ArchLabel, corev1.NodeSelectorOpIn, utils.ArchitectureAmd64,
+				utils.ArchitectureArm64, utils.ArchitectureS390x, utils.ArchitecturePpc64le)
+		})
+		It("should not set the node affinity when users node affinity is conflict", func() {
```
Prashanth684: nit: "should not set the node affinity when users node affinity is conflict"
to
"should not set the node affinity when users node affinity conflicts"

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,135 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+						},
+					},
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{
+								"app": "test",
+							},
+						},
+						Spec: corev1.PodSpec{
+							Affinity: &corev1.Affinity{
```
Prashanth684: do we want to have a test case with just "nodeSelector"? in that case I would assume nodeAffinity should also be added as both can exist

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,135 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+								NodeAffinity: &corev1.NodeAffinity{
+									RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+										NodeSelectorTerms: []corev1.NodeSelectorTerm{
+											{
+												MatchExpressions: []corev1.NodeSelectorRequirement{
+													{
```
Prashanth684: should we maybe also add a test case to test with two match expressions? one with archlabel and one without? and it should add the archlabel affinity to the match expression without the pre-existing label

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: sure, let me add it

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: @Prashanth684 sorry, I didn't undestand, could you explain more?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,135 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+														Key:      utils.OSLabel,
+														Operator: corev1.NodeSelectorOpIn,
+														Values: []string{
+															utils.OSLinux,
+														},
+													},
+												},
+											},
```
```suggestion
											{
												MatchExpressions: []corev1.NodeSelectorRequirement{
													{
														Key:      utils.OSLabel,
														Operator: corev1.NodeSelectorOpIn,
														Values: []string{
															utils.OSLinux,
														},
													},
												},
											},
											{
												{
														Key:      utils.ArchLabel,
														Operator: corev1.NodeSelectorOpIn,
														Values: []string{
															utils.ArchitectureS390X,
														},
													},
												},
											},
											
											
```

Regarding https://github.com/openshift/multiarch-manager-operator/pull/44/files#r1498687311, in this case, we expect that the first NodeSelectorTerm gets the automated arch-aware affinity.

The second (already including an arch-specific matchExpression) should not as we don't want to override any spec provided by users.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,135 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+								NodeAffinity: &corev1.NodeAffinity{
+									RequiredDuringSchedulingIgnoredDuringExecution: &corev1.NodeSelector{
+										NodeSelectorTerms: []corev1.NodeSelectorTerm{
+											{
+												MatchExpressions: []corev1.NodeSelectorRequirement{
+													{
+														Key:      utils.OSLabel,
+														Operator: corev1.NodeSelectorOpIn,
+														Values: []string{
+															utils.OSLinux,
```
aleskandro: Is there any other label we can use other than the OS one? It's ok to use it, but in the future, it might be good for us to include the OS matchExpression too (e.g., windows container images shouldn't risk landing into a linux node).

lwan-wanglin: got it, let me find one OS and Arch unrelated label

pkg/e2e/podplacement/pod_placement_test.go
Prashanth684: yes this is what i meant. sorry if I wasn't clear

pkg/e2e/podplacement/pod_placement_test.go
Prashanth684: @aleskandro you mean in the future we detect os metadata from image and add that affinity as well?

aleskandro: yeah, we already have that information in our objects (as well as manifest list images do), but I didn't want to generate that node affinity to keep the scope constrained to the architecture. However, if you remember some of the initial discussions about the design of it, it is a good piece of information that is also worth considering for the scheduler predicates. The current official recommendation (for OCP) is to leverage runtime classes that instruct the scheduler as a kind of 'side-effect'/additional feature, so it's just a "possible" future thing that we can onboard for free to better consider the platform information in the manifest list images we inspect.

Prashanth684: that's a good idea. let's create a ticket for it as a future add on

lwan-wanglin:
@aleskandro @Prashanth684 could you help review again?

for this pr, I was drafting the automation cases first, but I will update polarion case after the pr aprrove.
for others, I will draft polarion testcases first and request review, then automating.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,281 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+							"app": "test",
+						},
+					},
+					Template: corev1.PodTemplateSpec{
+						ObjectMeta: metav1.ObjectMeta{
+							Labels: map[string]string{
+								"app": "test",
+							},
+						},
+						Spec: corev1.PodSpec{
```
aleskandro: Hi @lwan-wanglin, while we're still implementing the initial cases, I wonder if it's possible to implement a factory for PodSpecs that we can use when creating workload objects (we can also do one for deployments that provides a withPodSpec method to chain with the PodSpec one).

We already have an example Pod factory at https://github.com/openshift/multiarch-manager-operator/blob/main/pkg/testing/framework/pod_factory.go

If we implement this, the amount of redundant code along the test specs will reduce a lot IMO.

lwan-wanglin: sure, let me try it.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -174,13 +174,281 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+								},
+							},
+						},
+					},
+				},
+			}
+			err = client.Create(ctx, &d)
+			Expect(err).NotTo(HaveOccurred())
+			// uncomment this check after https://issues.redhat.com/browse/MIXEDARCH-430, MMO adds an empty affinity now.
+			//verifyPodNodeAffinity(ns, "app", "test", utils.ArchLabel, corev1.NodeSelectorOpIn, utils.ArchitectureAmd64,
```
aleskandro: I don't think we will be able to verify this after MIXEDARCH430 lands. What we expect here is an empty affinity today and a nil one in the future.

So,we should not use verifyPodNodeAffinity and run something like:

```
Consistently(func(g Gomega) {
  pods := &corev1.PodList{}
..... 
  g.Expect(pods.Items).To(HaeEach(WithTransform(func (p *Pod) *Affinity {
     return p.Spec........Affinity
   }, BeEmpty())
   // TODO: Change to BeNil when MIXEDARCH430 is done.
}, e2e.WaitMedium())
```

aleskandro: Even better, I would check
1. Eventually (WaitShort) the pods are labeled with `scheduling-gate: removed`
2. The Affinity is set to BeEmpty (in future BeNil)

We could change verifyPodAffinity to achieve the above rules: check for the removed gate label first and then check the affinity. If you change the `nodeAffinityKey` and `operator` to be pointers, you can check if they are nil before running the matcher at line 463

pkg/utils/const.go
```diff
@@ -18,6 +18,7 @@ const (
 	NodeAffinityLabel               = "multiarch.openshift.io/node-affinity"
 	NodeAffinityLabelValueSet       = "set"
 	NodeAffinityLabelValueUnset     = "unset"
+	HostNameLabel                   = "kubernetes.io/hostname"
```
aleskandro:
```suggestion
	HostnameLabel                   = "kubernetes.io/hostname"
```

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: ok,  will fix

pkg/e2e/podplacement/pod_placement_test.go
aleskandro: Opened #46 and will make a jira card too once the migration of the boards is completed.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -20,6 +19,10 @@ import (
 const helloOpenshiftPublicMultiarchImage = "quay.io/openshifttest/hello-openshift:1.2.0"
 
 var _ = Describe("The Pod Placement Operand", func() {
+	var (
+		optIn     = corev1.NodeSelectorOpIn
+		archLabel = "kubernetes.io/arch"
```
aleskandro: we should have a constant around that has this value in const. 

Can we assign the value to this var from it?

pkg/testing/framework/podspec_factory.go
```diff
@@ -0,0 +1,127 @@
...
+	ps.podspec.Containers = make([]v1.Container, len(images))
+
+	for i, image := range images {
+		// compute hash of the image name using SHA-256
+		hasher := sha256.New()
+		hasher.Write([]byte(image))
+
+		ps.podspec.Containers[i] = v1.Container{
+			Image: image,
+			Name:  hex.EncodeToString(hasher.Sum(nil))[:63], // hash of the image name (40 characters, 63 is max)
```
aleskandro: is the final string longer than 63 chars?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -100,87 +85,221 @@ var _ = Describe("The Pod Placement Operand", func() {
...
-					Selector: &metav1.LabelSelector{
-						MatchLabels: map[string]string{
-							"app": "test",
-						},
-					},
-					Template: corev1.PodTemplateSpec{
-						ObjectMeta: metav1.ObjectMeta{
-							Labels: map[string]string{
-								"app": "test",
+			ps := corev1.PodSpec{
```
aleskandro: can't we use the factory here?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -100,87 +85,221 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+			var err error
+			ns := framework.NewEphemeralNamespace()
+			err = client.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer client.Delete(ctx, ns)
+			ps := framework.NewPodSpec().
+				WithContainersImages(helloOpenshiftPublicMultiarchImage).
+				WithNodeSelectorTermsMatchExpressions(
+					[]corev1.NodeSelectorRequirement{
```
aleskandro: Do you think it's worth it having a factory (or a few more) to handle this?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -28,38 +31,20 @@ var _ = Describe("The Pod Placement Operand", func() {
...
-							},
-						},
-					},
-				},
-			}
+			ps := framework.NewPodSpec().
+				WithContainersImages(helloOpenshiftPublicMultiarchImage).
+				Build()
+			d := framework.NewDeployment().
+				WithPodLabels("app", "test").
```
aleskandro: I think we can merge WithPodLabels and WithSelector, can't we?

aleskandro:
Great job @lwan-wanglin! Just a few other comments to check if we can anticipate some other future needs

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: if having a factory, the code would be more clear, let me do it

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: yes, will fix

pkg/testing/framework/podspec_factory.go
lwan-wanglin: yes, it's longer than 63 chars

pkg/testing/framework/deployment_factory.go
```diff
@@ -0,0 +1,65 @@
+package framework
+
+import (
+	//#nosec G505: Blocklisted import crypto/sha1: weak cryptographic primitive
```
aleskandro:
```suggestion
```

Do we still need this comment?

pkg/testing/framework/pod_spec_factory.go
```diff
@@ -0,0 +1,134 @@
+package framework
+
+import (
+	//#nosec G505: Blocklisted import crypto/sha1: weak cryptographic primitive
```
aleskandro:
```suggestion
```

Do we still need this comment?

pkg/testing/framework/volume_factory.go
```diff
@@ -0,0 +1,31 @@
+package framework
```
aleskandro: Is moving all the factories in a dedicated package, e.g., testing/factory, easy and worth it?

lwan-wanglin: that makes sense, thanks

aleskandro:
Thanks @lwan-wanglin

PR #52, opened 2024-03-07
MULTIARCH-4171: Multiarch builds w/ Konflux

aleskandro:
This PR set up the multiarch PipelineRun for building both the bundle and the operator as multiarch manifest-lists.

An additional layer is added to the bundle.Dockerfile image as Konflux has a bug with single-stage images extending `FROM scratch`. See https://issues.redhat.com/browse/KFLUXBUGS-10


Prashanth684:
where can i see the image that was built by konflux? i tried going into the pipeline for the job run, but there was nothing

aleskandro:
> where can i see the image that was built by konflux? i tried going into the pipeline for the job run, but there was nothing

i got it from the output logs in the konflux UI for now (the build-container taskrun output)

PR #56, opened 2024-03-13
Enable source build and fork the Dockerfile to build the enterprise image with Konflux

aleskandro:
Continuing with the Konflux builds onboarding. The release plan and the pipeline changes required for replacing the images refs in the bundle will follow up.

As Konflux cannot replace build contexts yet, this PR includes a temporary commit to revert when Konflux will make us able to replace base images without manual changes to the Dockerfile: https://github.com/openshift/multiarch-tuning-operator/pull/56/commits/4075e32c6181cece370ee7e12b69518524adcea7

### DIFF
--- a/.tekton/multiarch-tuning-operator-pull-request.yaml
+++ b/.tekton/multiarch-tuning-operator-pull-request.yaml
@@ -113,7 +113,7 @@ spec:
         description: Image tag expiration time, time values could be something like
           1h, 2d, 3w for hours, days, and weeks, respectively.
         name: image-expires-after
-      - default: "false"
+      - default: "true"
         description: Build a source image.
         name: build-source-image
         type: string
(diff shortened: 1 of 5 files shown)


.tekton/multiarch-tuning-operator-pull-request.yaml
```diff
@@ -19,7 +19,8 @@ metadata:
 spec:
   params:
     - name: dockerfile
-      value: Dockerfile
+      value: konflux.Dockerfile
+      # TODO: change to Dockerfile when https://issues.redhat.com/browse/KONFLUX-2361
```
Prashanth684, line 23: so konflux will build only an enterprise image now? not upstream?

aleskandro: yes. For the upstream we can use the Prow arm64/amd64 only ones, or we can even create a GitHub action to cross-build them (not sure if this is possible from this organization).

aleskandro:
/override ci/prow/e2e-gcp-multi-operator

(the failure is unrelated and we are verifying with the -olm job in any case).

PR #58, opened 2024-03-14
Fix gomod

aleskandro:


aleskandro:
/override  @openshift-ci
ci/prow/e2e-gcp-multi-operator

PR #59, opened 2024-03-14
MULTIARCH-4539: Fix system config synchronization in the allowedRegistries scenarios

aleskandro:
When the image.config.openshift.io/cluster object is configured with a set of allowedRegistries, the policy.json is configured by default not to accept any image's registry except those in the allowed registries set.

The operand was not setting the system context files to reflect this scenario properly, and the inspector didn't perform the signature verification (which processes the policy.json file's rules).

This PR fixes the issues caused by this scenario: the images from registries not in the allowed registries should not be inspected. 

**Note**: we import an additional package  (`github.com/containers/v5/signature`) to execute the signature verification. This allows us to refrain from writing our own structs to marshal/unmarshal the policy.json file. However, the package depends on the `gpgme` C library, and this implies (1) having CGO enabled and (2) the inability to cross-build from an amd64 builder by exploiting the Golang cross-compilation features. Therefore CGO is now enabled in the Dockerfile and I changed the docker-buildx will now build on per-arch container images (instead of cross-building on the amd64 builder and leveraging qemu only for the finally stage). Finally, as gpgme-devel (libgpgme-dev) is not available in docker.io/library/golang:1.21, I filed an issue to address as a follow-up to allow users with no RH creds to build the operator out of the box. https://github.com/openshift/multiarch-tuning-operator/issues/61


pkg/systemconfig/syncer.go
```diff
@@ -56,22 +56,24 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
 	defer s.unlockAndSync()
 	// Ensure the previous state is reset
 	for _, rc := range s.registriesConfContent.Registries {
-		rc.Allowed = nil
 		rc.Blocked = nil
 		rc.Insecure = nil
 	}
-	s.policyConfContent.resetTransports()
+	s.policyConfContent.reset()
```
Prashanth684: i'm surprised the image apis don't have a default registry config setting and we have to set these ourselves

Prashanth684: gives more room for error is what I'm afraid of

aleskandro: Yeah, this is on the podamn/containers side, actually. The idea here is that it's easier to rebuild the entire structure instead of computing the diff on each event. In fact, some events may be lost, and the deletion of some objects in the APIs can be impossible to track as their content is lost when we process them.

aleskandro: So, reset rebuild the default configuration and from that this method builds the system configuration based on the cached information we gather from the involved objects.

Prashanth684: ok..sounds fair..my question was - doesn't the containers API library already do that? why did we have to put all this together

aleskandro: I remember I had a quick look at the beginning of this implementation, maybe the related types and methods are not exported by the lib, or for some reasons it wasn't worth it to import another lib for this work (we'd still implement the transformation logic between the resources we watch and the /etc/containers conf).

aleskandro: Found https://github.com/containers/image/blob/main/signature/policy_types.go for the policy.json file. We will need to import this package in any case, so I'll check if this can be ok to replace the current type we own and remove ours if possible..

Prashanth684: great, but maybe let's get this in first and then continue the investigation

aleskandro: If seems we need it here for the verification of the signature policy (used to block/allow the registries)

lwan-wanglin:
the issue is fixed, now PPO does not deal with image which are not in AllowedRegistries list. and I also do a regression test, no more issue found.

aleskandro:
/override Red Hat Konflux / Red Hat Konflux / multiarch-tuning-operator-jjs7z / multiarch-tuning-operator-enterprise-contract

aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-jjs7z / multiarch-tuning-operator-enterprise-contract"

aleskandro:
FBC or SQLite based: get index image labels: error pulling image quay.io/operator-framework/opm:latest: failed to copy: httpReaderSeeker: failed open: unexpected status code https://quay.io/v2/operator-framework/opm/blobs/sha256:802b6fc123ece12a10e08193fb1ca29241727b7838a132e606bfc0552a0f110e: 502 Bad Gateway\n"

trying to test the non bundle too 

/test e2e-gcp-multi-operator

PR #63, opened 2024-03-20
Konflux build bundle with image ref replacement

aleskandro:
The bundle Dockerfile is re-generated each time make bundle is run. Until the full support for OLM operators is available in konflux, we must use an ad-hoc dockerfile. Therefore we shouldn't modify it for the workarounds due to the Konflux missing features.

Also, Konflux doesn't support setting the build-arg in the buildah(-remote) tasks (see https://issues.redhat.com/browse/KONFLUX-268). This PR adds a task for each arch for each pipelinerun (for each branch?) to build the bundle and refer the correct operator image.

### DIFF
--- a/.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
+++ b/.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
@@ -19,7 +19,7 @@ metadata:
 spec:
   params:
     - name: dockerfile
-      value: bundle.Dockerfile
+      value: bundle.konflux.Dockerfile
     - name: git-url
       value: '{{source_url}}'
     - name: image-expires-after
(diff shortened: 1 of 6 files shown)


.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
```diff
@@ -134,6 +136,41 @@ spec:
...
+              command:
+                - /bin/bash
+                - -c
+              args:
+                - |
+                  #!/bin/bash
+                  COMMIT=$(tasks.clone-repository.results.commit)
+                  for d in /workspace/*; do
+                    echo $d
+                    sed -i "s/^ARG IMG=''/ARG IMG='$(params.operator-image)'/" /workspace/$d/bundle.konflux.Dockerfile
```
Prashanth684: sad but necessary..sigh..

aleskandro: Yeah 😔

aleskandro:
:fireworks:

aleskandro:
(merged to unblock #57

PR #67, opened 2024-03-29
MULTIARCH-4312: Add more e2e test for ppo

lwan-wanglin:
- Add e2e for  [Single/Multi container images](https://issues.redhat.com/browse/MULTIARCH-4318) and [high-level resources ](https://issues.redhat.com/browse/MULTIARCH-4312).
- Change func `WithSelectorAndPodLabels` to recive map[string]string instand of []string to make code more match the name
-  fix the description typo for nodeSelector test

### DIFF
--- a/pkg/e2e/podplacement/pod_placement_test.go
+++ b/pkg/e2e/podplacement/pod_placement_test.go
@@ -16,9 +16,15 @@ import (
 	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
 )
 
-const helloOpenshiftPublicMultiarchImage = "quay.io/openshifttest/hello-openshift:1.2.0"
+const (
+	helloOpenshiftPublicMultiarchImage = "quay.io/openshifttest/hello-openshift:1.2.0"
+	helloOpenshiftPublicArmImage       = "quay.io/openshifttest/hello-openshift:arm-1.2.0"
+	helloOpenshiftPublicArmAmdImage    = "quay.io/openshifttest/hello-openshift:arm-amd-1.2.0"
+	helloOpenshiftPublicArmPpcImage    = "quay.io/openshifttest/hello-openshift:arm-ppc64le-1.2.0"
+)
 
 var _ = Describe("The Pod Placement Operand", func() {
+	var podLabel = map[string]string{"app": "test"}
 	Context("When a deployment is deployed with a single container and a public image", func() {
 		It("should set the node affinity", func() {
 			var err error
@@ -31,7 +37,7 @@ var _ = Describe("The Pod Placement Operand", func() {
 				WithContainersImages(helloOpenshiftPublicMultiarchImage).
 				Build()
 			d := NewDeployment().
-				WithSelectorAndPodLabels("app", "test").
+				WithSelectorAndPodLabels(podLabel).
 				WithPodSpec(ps).
 				WithReplicas(utils.NewPtr(int32(1))).
 				WithName("test-deployment").
@@ -89,7 +95,7 @@ var _ = Describe("The Pod Placement Operand", func() {
 				WithVolumes(v).
 				Build()
 			d := NewDeployment().
-				WithSelectorAndPodLabels("app", "test").
+				WithSelectorAndPodLabels(podLabel).
 				WithPodSpec(ps).
 				WithReplicas(utils.NewPtr(int32(1))).
 				WithName("test-deployment").
@@ -122,7 +128,7 @@ var _ = Describe("The Pod Placement Operand", func() {
 				WithNodeSelectorTerms(hostnameLabelNSTs).
 				Build()
 			d := NewDeployment().
-				WithSelectorAndPodLabels("app", "test").
+				WithSelectorAndPodLabels(podLabel).
 				WithPodSpec(ps).
 				WithReplicas(utils.NewPtr(int32(1))).
 				WithName("test-deployment").
@@ -152,7 +158,7 @@ var _ = Describe("The Pod Placement Operand", func() {
 				WithContainersImages(helloOpenshiftPublicMultiarchImage).
 				WithNodeSelectorTerms(archLabelNSTs).Build()
 			d := NewDeployment().
-				WithSelectorAndPodLabels("app", "test").
+				WithSelectorAndPodLabels(podLabel).
 				WithPodSpec(ps).
 				WithReplicas(utils.NewPtr(int32(1))).
 				WithName("test-deployment").
@@ -181,7 +187,7 @@ var _ = Describe("The Pod Placement Operand", func() {
 				WithContainersImages(helloOpenshiftPublicMultiarchImage).
 				WithNodeSelectorTerms(archLabelNSTs, hostnameLabelNSTs).Build()
 			d := NewDeployment().
-				WithSelectorAndPodLabels("app", "test").
+				WithSelectorAndPodLabels(podLabel).
 				WithPodSpec(ps).
 				WithReplicas(utils.NewPtr(int32(1))).
 				WithName("test-deployment").
@@ -196,8 +202,9 @@ var _ = Describe("The Pod Placement Operand", func() {
 			expectedHostnameNST := NewNodeSelectorTerm().WithMatchExpressions(&hostnameLabelNSR, &expectedArchLabelNSR).Build()
 			verifyPodNodeAffinity(ns, "app", "test", expectedHostnameNST, archLabelNSTs)
 		})
-		It("should set the node affinity when nodeSelector exist", func() {
+		It("should not set the node affinity when nodeSelector exist", func() {
 			var err error
+			var nodeSelectors = map[string]string{utils.ArchLabel: utils.ArchitectureAmd64}
 			ns := framework.NewEphemeralNamespace()
 			err = client.Create(ctx, ns)
 			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -279,7 +279,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) GOFLAGS='' go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) GOFLAGS='' go install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240208111015-5923139bc5bd
```
aleskandro: hey, see #69. I'll go through the other code later.

lwan-wanglin:
/pj-rehearse periodic-ci-openshift-openshift-tests-private-release-4.16-multi-nightly-baremetal-upi-ovn-ipv4-arm-mixarch-f14-day2-64k-pagesize

controllers/podplacement/pod_reconciler_test.go
```diff
@@ -20,7 +20,7 @@ import (
 
 var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 	When("Handling Single-container Pods", func() {
-		Context("with different image types", func() {
+		Context("with different image types", Serial, func() {
```
aleskandro: Why is this serial?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -44,6 +51,7 @@ var _ = Describe("The Pod Placement Operand", func() {
 					utils.ArchitectureArm64, utils.ArchitectureS390x, utils.ArchitecturePpc64le).
 				Build()
 			expectedNSTs := NewNodeSelectorTerm().WithMatchExpressions(&archLabelNSR).Build()
+			verifyPodlabels(ns, "app", "test", "multiarch.openshift.io/scheduling-gate", "removed")
```
aleskandro: This is unrelated to the PR, right? If yes, could you split the single commit into more commits describing what we are achieving in each set of changes?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -194,9 +206,174 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+				WithNamespace(ns.Name).
+				Build()
+			err = client.Create(ctx, &d)
+			Expect(err).NotTo(HaveOccurred())
+			//Should not set Node Affinity
+			verifyPodlabels(ns, "app", "test", "multiarch.openshift.io/scheduling-gate", "removed")
+			verifyPodNodeAffinity(ns, "app", "test")
+		})
+	})
+	Context("PPO should respect to an opt-out namespaceSelector in PodPlacementConfig CR", Serial, func() {
```
aleskandro: This should be tested in the operator suite iinw.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -194,9 +206,174 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+				WithKeyAndValues(utils.ArchLabel, corev1.NodeSelectorOpIn, utils.ArchitectureAmd64,
+					utils.ArchitectureArm64, utils.ArchitectureS390x, utils.ArchitecturePpc64le).
+				Build()
+			expectedNSTs := NewNodeSelectorTerm().WithMatchExpressions(&archLabelNSR).Build()
+			//should handle the namespace
+			verifyPodlabels(ns, "app", "test", "multiarch.openshift.io/scheduling-gate", "removed")
+			verifyPodNodeAffinity(ns, "app", "test", expectedNSTs)
+		})
+	})
+	Context("PPO should respect to an opt-in namespaceSelector in PodPlacementConfig CR", Serial, func() {
```
aleskandro: same for this

pkg/testing/builder/daemonset_builder.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if d.daemonset.Spec.Selector == nil {
+		d.daemonset.Spec.Selector = &metav1.LabelSelector{}
+	}
+	if d.daemonset.Spec.Selector.MatchLabels == nil {
+		d.daemonset.Spec.Selector.MatchLabels = make(map[string]string)
+	}
+	if len(kv)%2 != 0 {
+		panic("the number of arguments must be even")
+	}
```
aleskandro:
```suggestion
	if len(kv)%2 != 0 {
		panic("the number of arguments must be even")
	}
	if d.daemonset.Spec.Template.Labels == nil {
		d.daemonset.Spec.Template.Labels = make(map[string]string)
	}
	if d.daemonset.Spec.Selector == nil {
		d.daemonset.Spec.Selector = &metav1.LabelSelector{}
	}
	if d.daemonset.Spec.Selector.MatchLabels == nil {
		d.daemonset.Spec.Selector.MatchLabels = make(map[string]string)
	}
```

aleskandro: Thinking at it, could we rather use a dedicated struct or a map[string]string? We'd just reassign it to Labels and MatchLabels probably, without for loop, panic, etc...

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -272,3 +749,50 @@ func verifyPodAnnotations(ns *corev1.Namespace, labelKey string, labelInValue st
...
+		err := client.List(ctx, pods, &runtimeclient.ListOptions{
+			Namespace:     ns.Name,
+			LabelSelector: labelSelector,
+		})
+		g.Expect(err).NotTo(HaveOccurred())
+		g.Expect(pods.Items).NotTo(BeEmpty())
+		for i := 0; i < len(pods.Items); i++ {
+			pod := pods.Items[i]
+			g.Expect(pod.Labels).NotTo(BeEmpty())
+			for j := 0; j < len(kv); j += 2 {
```
aleskandro: Maybe there is a 'ginkgo-native' way of doing this?

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: yes, I missed this part of lable verification in the previous prs. 
sure, will split to more commits to make the single pr clear.

pkg/e2e/podplacement/pod_placement_test.go

controllers/podplacement/pod_reconciler_test.go
lwan-wanglin: @aleskandro When we run parallelly, the registy is setting at the SynchronizedBeforeSuite block which is running in process 1, only the testcase running in the process 1 can get the docker images, for case running in other process, it will fail, but I don't know how to fix it now.

lwan-wanglin: can I just set it Setial and fix it in another pr?

aleskandro: uhm, is this because the data structure having the docker images is empty in the children threads?

aleskandro: If yes and if you think this requires more investigation time, I prefer that we decide what this PR targets should be and do more PRs:
1. Adding test cases (higher in pri)
2. Making them run in parallel (we set this as a low priority for now)
3. Fixing the verification of the labels (medium...)
4. ...

I understand you were working on this PR with more sets of changes as some can depend on the previous ones, but I continue to suggest at least having separate commits and then deciding which commit(s) to land first and move others to another branch (to rebase after the previous commits' PRs merge): see https://graphite.dev/blog/stacked-prs as an interesting pattern regarding this.

lwan-wanglin: the `SynchronizedBeforeSuite` first func(){} is running in process 1 before all children thread start , and the images should already push to reigstry. so I am not really sure.

lwan-wanglin: @aleskandro  yes, I have updated this pr's goal in Description, we can enable parallel in current pr and  just set Serail for controller/operator, then the all others will run in parallel.
the verification of the labels and namespaceSelector test will split to more PRs.

aleskandro: @lwan-wanglin based on what you say, I would get either full enablement of parallel or none: the risk is that we will have more and more serial test cases that shouldn't be such because we are still missing the enablement of a 'shared memory' to be used by the test cases.

I prefer we focus on the remaining test cases and the integration with the Openshift E2e (tests-private, origin, ...) to have quality in the release of the operator.


Also, as above, `parallel` is lower in pri.

lwan-wanglin: got you ,that make sense, let me remove parallel part from this pr.

aleskandro: When you'll work on this, I'd start by checking what's the value of `registryAddress` in the children processes (e.g., at line 26).

aleskandro:
/lgtm

/approve

/override "Red Hat Konflux / multiarch-tuning-operator-bundle-on-pull-request" "Red Hat Konflux / multiarch-tuning-operator-on-pull-request"

aleskandro:
Thanks @lwan-wanglin

PR #68, opened 2024-03-29
MULTIARCH-4551: Add logo to CSV file for operator

AnnaZivkovic:
Screenshot from 2024-04-01 14-56-39

PR #66, opened 2024-03-28
MULTIARCH-4546: Add supported arch labels to the CSV

AnnaZivkovic:

PR #62, opened 2024-03-19
MULTIARCH-4306:  avoid the possibility of deleting the operand togeth…

AnnaZivkovic:
…er with the uninstallation of the operator.

PR #69, opened 2024-04-02
Pin setup-envtest

aleskandro:
setup-envtest is installed in the automatically generated makefile target.
Pinning its version to avoid updates to its repo like the minimum golang version in the go.mod can break compatibility and CI.

See kubernetes-sigs/kubebuilder#2480

Blocks #62 #66 #67 #68


aleskandro:

PR #70, opened 2024-04-05
NO-JIRA: Fixes to the CSV, Godoc

aleskandro:
This PR proposes minor fixes to adjust how the operator's pages are shown in the OperatorHub sections of the console.
- Improved Golang doc for the CRD 
- Improved description of the operator and operand
- Removed some labels from the CR (@Prashanth684 we were unsure of this, but checked other operators and they don't ship those labels either)
- Added an annotation to declare the minimum openshift version we can install this operator on: unfortunately, this needed an hack in the Makefile `bundle` target as operator-sdk doesn't consider custom annotations in the `bundle/metadata/annotations.yaml` file (xref threads in the internal slack)
- Adds suggested namespace to deploy the operator in the default `openshift-multiarch-tuning-operator` namespace.

~Finally, when deploying the operator through OLM or OperatorHub in a namespace different than the default one (`openshift-multiarch-tuning-operator`), the operand's serviceaccount are not bound to the expected Roles and ClusterRoles properly. This is because OLM doesn't patch the subject's namespace in the ClusterRoleBinding and RoleBinding manifests shipped by the bundle.~

~I like neither this solution, nor alternatives:~
~1. The operator should be able to create ad-hoc cluster role, roles and clusterolebindings. (And maybe the operator's service account should be allowed to perform the same actions that the role or clusterrole is meant to perform, plus the ones for managing clusterroles, clusterrolebindings, roles and rolebindings)~
~2. Create a single service account and use it for both the operator and the operands~
~3. Deploy all the operands, with 0 replicas, and scale them based on the CR config~

~IMO, the best would be that OLM patches the subjects' namespaces (perhaps with some condition/selector) like Kustomize does.~

Here some screenshots showing the changing in the console.

Screenshot_20240405_155513
Screenshot_20240405_155530
Screenshot_20240405_162034
Screenshot_20240405_162128

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,22 +2,27 @@
 
 The Multiarch Tuning Operator enhances the user experience for administrators of Openshift clusters with
 multi-architecture compute nodes or Site Reliability Engineers willing to migrate from single-arch to multi-arch
-OpenShift. When diverse CPU architectures coexist within a cluster, the Multiarch Tuning Operator operator stands out as a pivotal tool to
+OpenShift. When diverse CPU architectures coexist within a cluster, the Multiarch Tuning Operator stands out as a pivotal tool to
 enhance efficiency and streamline operations such as architecture-aware scheduling of workloads.
 
 The development work is still ongoing and there is no official, general available, release of it yet.
 
-## Description
+## Operands
 
-- **Architecture aware Pod Placement**: The pod placement operand aims to automate the 
-  inspection of the container images, derive a set of architectures supported by a pod and use it to
-  automatically define strong predicates based on the `kubernetes.io/arch` label in the pod's nodeAffinity. 
+- **Architecture aware Pod Placement**: The pod placement operand consists of
+  the `PodPlacementController` and the `PodPlacementWebhook` and is managed through
+  a singleton custom resource - `podplacementconfigs.multiarch.openshift.io`.
+  Its aim is to automate the set up of strong predicates based on the
+  `kubernetes.io/arch` label in the pod's _nodeAffinity_ by inspecting the container
+  images in each pod and deriving a set of architectures supported by
+  the pod. When a pod is created, the `PodPlacementWebhook` will add the
+  `multiarch.openshift.io/scheduling-gate` scheduling gate.
+  It will prevent the pod from being scheduled until the `PodPlacementController`
+  computes a predicate for the `kubernetes.io/arch` label, adds it as a node affinity
+  requirement to the pod spec, and removes the scheduling gate.
   This operand is based on the [KEP-3521](https://github.com/kubernetes/enhancements/blob/afad6f270c7ac2ae853f4d1b72c379a6c3c7c042/keps/sig-scheduling/3521-pod-scheduling-readiness/README.md) and
   [KEP-3838](https://github.com/kubernetes/enhancements/blob/afad6f270c7ac2ae853f4d1b72c379a6c3c7c042/keps/sig-scheduling/3838-pod-mutable-scheduling-directives/README.md), as
   described in the [Openshift EP](https://github.com/openshift/enhancements/blob/6cebc13f0672c601ebfae669ea4fc8ca632721b5/enhancements/multi-arch/multiarch-manager-operator.md) introducing it.
-  When a pod is created, the mutating webhook will add the `multiarch.openshift.io/scheduling-gate` scheduling gate, that will
-  prevent the pod from being scheduled until the controller computes a predicate for the `kubernetes.io/arch` label,
-  adds it as node affinity requirement to the pod spec and removes the scheduling gate.
 
 ## Getting Started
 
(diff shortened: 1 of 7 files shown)


config/manifests/bases/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -57,4 +84,5 @@ spec:
   minKubeVersion: 1.27.0
   provider:
     name: Red Hat
+    url: https://github.com/openshift/cluster-nfd-operator
```
Prashanth684: why is nfd specified here?

aleskandro: Thanks

aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-Operator" "Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-Operator-bundle"

README.md
```diff
@@ -2,22 +2,27 @@
...
-  inspection of the container images, derive a set of architectures supported by a pod and use it to
-  automatically define strong predicates based on the `kubernetes.io/arch` label in the pod's nodeAffinity. 
+## Operands
+
+- **Architecture aware Pod Placement**: The pod placement operand consists of
+  the `PodPlacementController` and the `PodPlacementWebhook` and is managed through
+  a singleton custom resource - `podplacementconfigs.multiarch.openshift.io`.
+  Its aim is to automate the set up of strong predicates based on the
+  `kubernetes.io/arch` label in the pod's _nodeAffinity_ by inspecting the container
+  images in each cluster's pod and deriving a set of architectures supported by
```
Prashanth684: `in each pod` rather than cluster's pod ?

apis/multiarch/v1alpha1/podplacementconfig_types.go
```diff
@@ -22,56 +22,35 @@ import (
...
-	//         "prod",
-	//         "staging"
-	//       ]
-	//     }
-	//   ]
-	// }
+	// {"namespaceSelector":{"matchExpressions":[{"key":"multiarch.openshift.io/exclude-pod-placement","operator":"DoesNotExist"}]}}
+	//
+	// The operand will set the node affinity requirement in all the pods created in namespaces missing
+	// the `multiarch.openshift.io/exclude-pod-placement` label.
```
Prashanth684, line 40: "set the node affinity requirement for all pods created in namespaces which do not have the `multiarch.openshift.io/exclude-pod-placement` label" ?

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -187,7 +207,7 @@ spec:
             metadata:
               annotations:
                 kubectl.kubernetes.io/default-container: manager
-                multiarch.openshift.io/image: registry.ci.openshift.org/origin/multiarch-tuning-operator:main
+                multiarch.openshift.io/image: quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator/multiarch-tuning-operator:on-pr-921a2a96a68f18a04b504146c74b6ebbd49b49e8
```
Prashanth684: should this still be the upstream prow image?

config/manifests/bases/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -51,10 +75,11 @@ spec:
   - multiarch-tuning-operator
   - pod-placement
   links:
-  - name: Multiarch Tuning Operator
-    url: https://github.com/openshift/multiarch-tuning-operator
+  - name: Multiarch Documentation
+    url: https://docs.openshift.com/container-platform/4.14/post_installation_configuration/configuring-multi-arch-compute-machines/multi-architecture-configuration.html
```
Prashanth684: hmm..should we wait to set  this till we have a docs page?

aleskandro: Yeah, or maybe we can try to add the url that we expect to use for the final doc (I think that one link will be available only after we release, so we need something in advance)

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
aleskandro: good catch! Thanks

config/manifests/bases/multiarch-tuning-operator.clusterserviceversion.yaml
aleskandro: I set it to this page: https://docs.openshift.com/container-platform/4.16/post_installation_configuration/configuring-multi-arch-compute-machines/multiarch-tuning-operator.html

@kelbrown20 is this a valid future url for the documentation of the operator? Can we set this in advance and then you let the URL be this one in the future doc?

aleskandro:
/retest-required 

/hold

(For a last ack from Prashanth)

aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-bundle-on-pull-request" "Red Hat Konflux / multiarch-tuning-operator-on-pull-request"

PR #72, opened 2024-04-11
MULTIARCH-4304: Add validation for the name of the PodPlacementConfig…

AnnaZivkovic:
… so there can only be one cluster

Unfortunately, make manifests overrides manually adding validation (example [here](https://github.com/openshift/aws-ebs-csi-driver-operator/pull/29))  in multiarch.openshift.io_podplacementconfigs.yaml. 

The only issue with using  +kubebuilder:validation:XValidation:rule is the error message is not as nicely formatted as shown 
Screenshot from 2024-04-11 13-37-40


aleskandro:
The error doesn't look so bad to me and the solution is quick enough to cover the problem for now. 

The alternative is to hack the Makefile target so that the CRD yaml is patched after all the other steps execute: we're needing it in https://github.com/openshift/multiarch-tuning-operator/pull/70/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R291 (as per suggestions in some operator-framework threads internally), but I like that even less. 

What's the `field` the API is considering to be `<nil>`? Any way we can change it?

apis/multiarch/v1alpha1/podplacementconfig_types.go
```diff
@@ -85,6 +85,7 @@ type PodPlacementConfigStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=podplacementconfigs,scope=Cluster
+// +kubebuilder:validation:XValidation:rule="self.metadata.name == \"cluster\"",message="metadata.name must be cluster"
```
aleskandro: Perhaps, we are still missing some others?

image

https://book.kubebuilder.io/reference/markers/crd

AnnaZivkovic: The <nil> field is coming from where if finds the validation. If you look at where +kubebuilder:validation:XValidation:rule is placed in the [yaml](https://github.com/openshift/multiarch-tuning-operator/blob/d824e8ed5cb5c2d48d9b1821c45c9d32a4da11e1/config/crd/bases/multiarch.openshift.io_podplacementconfigs.yaml#L187)  you can see that it is outside of openAPIV3Schema. Where as all of the other validation is nested in there. The solution in this [pr](https://github.com/openshift/aws-ebs-csi-driver-operator/pull/29/commits/93a4711b17971b668bf68269e7a5a1af3869fa48) gives us a nice error message

aleskandro: mmm, I see.

Can we also check if we can patch the CRD via [kustomization](https://github.com/openshift/multiarch-tuning-operator/blob/main/config/crd/kustomization.yaml)? There's a patch folder to store patches and refer to the patch files in the kustomization for the crd.

Ideally, using a strategicMerge patch would be great, but it may be impossible because the path to the object we want to patch contains a list (`versions`). You can try that first and then rely on the JSON patch if it doesn't work.

At first take, this solution looks better as it allows us to set the nicer message and is not as bad as in patching from the Makefile target.

aleskandro:
@AnnaZivkovic, remember that we also have to remove any previous code that was checking the name in the pod placement config controller. 

Also, let's check the integration test cases, patch any that were already verifying the previous work, or add at least one integration/unit test that verifies CR with a wrong name cannot be created

AnnaZivkovic:
@aleskandro are you talking about removing the logic [here](https://github.com/openshift/multiarch-tuning-operator/blob/b51a51518087223fa04a863bfc9ec2614350478b/controllers/operator/podplacementconfig_controller.go#L91C50-L91C77) and the stuff below it?

aleskandro:
From line 100 in particular. But yes, we could even delete that if statement that you linked as we can assume the name of the object will always be the correct one

apis/multiarch/v1alpha1/podplacementconfig_types.go
```diff
@@ -85,6 +85,7 @@ type PodPlacementConfigStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:path=podplacementconfigs,scope=Cluster
+
```
aleskandro:
```suggestion
```

controllers/operator/podplacementconfig_controller_test.go
```diff
@@ -176,23 +176,6 @@ var _ = Describe("Controllers/PodPlacementConfig/PodPlacementConfigReconciler",
 				Expect(err).NotTo(HaveOccurred(), "failed to delete PodPlacementConfig", err)
 				validateDeletion()
 			})
```
aleskandro: Can we rather adjust the test case according to the new behavior?

```suggestion
			})
			It("should refuse to create a PodPlacementConfig with an invalid name", func() {
				ppc := newPodPlacementConfig().WithName("invalid-name").Build()
				err := k8sClient.Create(ctx, ppc)
				Expect(err).To(HaveOccurred(), "failed to create PodPlacementConfig", err)
			})
```

This should already work, can we also validate that the error is specifically the one we expect?

controllers/operator/podplacementconfig_controller_test.go
```diff
@@ -179,7 +179,7 @@ var _ = Describe("Controllers/PodPlacementConfig/PodPlacementConfigReconciler",
 			It("should refuse to create a PodPlacementConfig with an invalid name", func() {
 				ppc := newPodPlacementConfig().WithName("invalid-name").Build()
 				err := k8sClient.Create(ctx, ppc)
-				Expect(err).NotTo(HaveOccurred(), "failed to create PodPlacementConfig", err)
+				Expect(err).NotTo(HaveOccurred(), "fcc", err)
```
aleskandro: should an error rather occur? If we try to create a PodPlacementConfig with a wrong name, the API should deny the request, right?

controllers/operator/podplacementconfig_controller_test.go
```diff
@@ -188,9 +188,7 @@ var _ = Describe("Controllers/PodPlacementConfig/PodPlacementConfigReconciler",
 							Namespace: utils.Namespace(),
 						},
 					}), ppc)
-					g.Expect(err).To(HaveOccurred(), "An error should have been detected")
-					g.Expect(crclient.IgnoreNotFound(err)).NotTo(HaveOccurred(),
-						"The error should not be different than a NotFound error")
+					g.Expect(err).NotTo(HaveOccurred(), "An error should not have been detected")
```
aleskandro: And this shouldn't be needed anymore once the API reports a failure and the object is not created.

controllers/operator/podplacementconfig_controller_test.go
AnnaZivkovic: The patches dont seem to get applied until we bundle it. So it seems to be applied at a later time

aleskandro: https://github.com/aleskandro/multiarch-manager-operator/blob/main/controllers%2Fpod_placement%2Fsuite_test.go#L55

I think this needs some update as it can't consider patches as of now.

controllers/operator/clusterpodplacementconfig_controller_test.go
```diff
@@ -179,19 +179,7 @@ var _ = Describe("Controllers/ClusterPodPlacementConfig/ClusterPodPlacementConfi
...
-						ObjectMeta: metav1.ObjectMeta{
-							Name:      "invalid-name",
-							Namespace: utils.Namespace(),
-						},
-					}), ppc)
-					g.Expect(err).To(HaveOccurred(), "An error should have been detected")
-					g.Expect(crclient.IgnoreNotFound(err)).NotTo(HaveOccurred(),
-						"The error should not be different than a NotFound error")
-				}).Should(Succeed())
+				Expect(err).To(HaveOccurred(), "The creation of the ClusterPodPlacementConfig with a wrong name failed", err)
```
aleskandro:
```suggestion
				Expect(err).To(HaveOccurred(), "The creation of the ClusterPodPlacementConfig with a wrong name did not fail", err)
```

controllers/operator/suite_test.go
```diff
@@ -17,12 +17,20 @@ limitations under the License.
...
+	"bytes"
 	"context"
 	"net/http"
 	"path/filepath"
 	"testing"
 	"time"
 
+	"k8s.io/klog/v2"
+
+	"k8s.io/apimachinery/pkg/api/errors"
```
aleskandro:
```suggestion
	"k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/klog/v2"
```

controllers/operator/suite_test.go
```diff
@@ -95,6 +103,18 @@ func startTestEnv() {
...
+	resMap, err := kustomizer.Run(filesys.MakeFsOnDisk(), filepath.Join("..", "..", "config", "crd"))
+	Expect(err).NotTo(HaveOccurred())
+
+	// Write the resulting resources to the buffer
+	for _, res := range resMap.Resources() {
+		yamlData, err := res.AsYAML()
+		Expect(err).NotTo(HaveOccurred())
+		_, err = buffy.Write(yamlData)
+		Expect(err).NotTo(HaveOccurred())
+	}
```
aleskandro:
```suggestion
```

It seems we can avoid putting all together in `buffy` and splitting again. I'd move this code a bit below.

controllers/operator/suite_test.go
```diff
@@ -112,6 +132,62 @@ func startTestEnv() {
...
+	Expect(err).NotTo(HaveOccurred())
+}
+
+func applyResources(buffer *bytes.Buffer) error {
+	// Create a universal decoder for deserializing the resources
+	decoder := scheme.Codecs.UniversalDeserializer()
+
+	// Split the buffer into individual YAML documents
+	docs := bytes.Split(buffer.Bytes(), []byte("\n---[ ]*\n"))
+	for _, doc := range docs {
```
aleskandro:
```suggestion
	klog.Info("Applying CRDs to the test environment")
	kustomizer := krusty.MakeKustomizer(krusty.MakeDefaultOptions())
	resMap, err := kustomizer.Run(filesys.MakeFsOnDisk(), filepath.Join("..", "..", "config", "crd"))
	Expect(err).NotTo(HaveOccurred())
	err = applyResources(resMap)
	Expect(err).NotTo(HaveOccurred())
}

func applyResources(resources resmap.ResMap) error {
	// Create a universal decoder for deserializing the resources
	decoder := scheme.Codecs.UniversalDeserializer()

	for _, res := range resources.Resources() {
		raw, err := res.AsYAML()
		Expect(err).NotTo(HaveOccurred())
```

We can build the kustomize-d resources and directly use the returned `resMap.Resources()` as a list of documents to apply

controllers/operator/suite_test.go
```diff
@@ -112,6 +132,62 @@ func startTestEnv() {
...
+			}
+		} else {
+			// Resource exists, update it
+			obj.(client.Object).SetResourceVersion(existingObj.GetResourceVersion())
+			err = k8sClient.Update(context.TODO(), obj.(client.Object))
+			if err != nil {
+				return err
+			}
+		}
+	}
```
aleskandro:
```suggestion
		if err != nil && !errors.IsNotFound(err) {
			// Return error if it's not a "not found" error
			return err
		}
		if err == nil {
			// Resource exists, update it
			obj.(client.Object).SetResourceVersion(existingObj.GetResourceVersion())
			err = k8sClient.Update(ctx, obj.(client.Object))
			if err != nil {
				return err
			}
		} else { // errors.IsNotFound(err) is true
			// Resource does not exist, create it
			err = k8sClient.Create(ctx, obj.(client.Object))
			if err != nil {
				return err
			}
		}
```

More about taste here: to avoid many nested if statements.

aleskandro (commented):
Hey @AnnaZivkovic thanks!

Leaving some comments. In general, it looks good to me. I wonder if we can simplify the kustomize to buffer process by using resMap (see suggestions).

controllers/operator/suite_test.go
aleskandro: Also, consider if we really need this code now (we should always hit the Create branch, right?)

AnnaZivkovic: Ill test it

AnnaZivkovic: I dont think I can remove the update resource segment. When I run `make test` without it I get the following error

```
[BeforeSuite] 
/home/user/Working/multiarch-manager-operator/controllers/operator/suite_test.go:79
  Enter [BeforeSuite] TOP-LEVEL - /home/user/Working/multiarch-manager-operator/controllers/operator/suite_test.go:79 @ 06/07/24 15:26:57.682
  STEP: bootstrapping test environment - /home/user/Working/multiarch-manager-operator/controllers/operator/suite_test.go:99 @ 06/07/24 15:26:57.683
  2024-06-07T15:26:57-07:00	DEBUG	controller-runtime.test-env	starting control plane
  2024-06-07T15:27:00-07:00	DEBUG	controller-runtime.test-env	installing CRDs
  2024-06-07T15:27:00-07:00	DEBUG	controller-runtime.test-env	reading CRDs from path	{"path": "../../config/crd/bases"}
  2024-06-07T15:27:00-07:00	DEBUG	controller-runtime.test-env	read CRDs from file	{"file": "multiarch.openshift.io_clusterpodplacementconfigs.yaml"}
  2024-06-07T15:27:00-07:00	DEBUG	controller-runtime.test-env	installing CRD	{"crd": "clusterpodplacementconfigs.multiarch.openshift.io"}
  2024-06-07T15:27:00-07:00	DEBUG	controller-runtime.test-env	adding API in waitlist	{"GV": "multiarch.openshift.io/v1alpha1"}
  2024-06-07T15:27:00-07:00	DEBUG	controller-runtime.test-env	installing webhooks
I0607 15:27:00.543666 1966467 suite_test.go:124] Applying CRDs to the test environment
  [FAILED] Unexpected error:
      <*errors.StatusError | 0xc0007c0280>: 
      customresourcedefinitions.apiextensions.k8s.io "clusterpodplacementconfigs.multiarch.openshift.io" already exists
      {
        ...
              Status: "Failure",
              Message: "customresourcedefinitions.apiextensions.k8s.io \"clusterpodplacementconfigs.multiarch.openshift.io\" already exists",
              Reason: "AlreadyExists",
              ...
  Full Stack Trace
    github.com/openshift/multiarch-tuning-operator/controllers/operator.startTestEnv()
    	/home/user/Working/multiarch-manager-operator/controllers/operator/suite_test.go:129 +0xd54
    github.com/openshift/multiarch-tuning-operator/controllers/operator.glob..func2()
    	/home/user/Working/multiarch-manager-operator/controllers/operator/suite_test.go:85 +0x2ae
```

aleskandro:
/lgtm
/approve

Thanks @AnnaZivkovic

PR #80, opened 2024-04-19
Remove annotation to force scheduling the operator on amd64 nodes

aleskandro:
Now, Prow builds multiarch images in the presubmits too


aleskandro:
For this to work we need the prow test to run on build10 too.

PR #84, opened 2024-04-24
Fix vendoring issue when building the container image with the internal base

aleskandro:
vendor should be copied before go mod download to avoid:
> module is explicitly required in go.mod, but not marked as explicit in vendor/modules.txt


lwan-wanglin:
/pj-rehearse periodic-ci-openshift-multiarch-tuning-operator-release-4.16-e2e-gcp-multi-operator

aleskandro:
Merging as the failing tests is not required and failed for other reasons.

PR #87, opened 2024-04-25
NO-JIRA: Makefile target to verify make bundle doesn't change the current code

aleskandro:
This will run in CI to check that there is no need for running `make bundle` in the changes from a Pull Request.


aleskandro:
/override ci/prow/e2e-gcp-multi-operator

/override ci/prow/e2e-gcp-multi-operator-olm

(unneeded as unrelated to the changes)

PR #94, opened 2024-05-03
[KFLUX] Update operator ref in the bundle

aleskandro:


aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-bundle-on-pull-request"

PR #95, opened 2024-05-03
Enable konflux prefetch

ralphbean:
This is a pre-requisite for hermetic builds, but it has other benefits. It will enable populating a more accurate SBOM for you and enable the production of source containers: https://access.redhat.com/articles/3410171

### DIFF
--- a/.tekton/multiarch-tuning-operator-pull-request.yaml
+++ b/.tekton/multiarch-tuning-operator-pull-request.yaml
@@ -102,7 +102,7 @@ spec:
         description: Execute the build with network isolation
         name: hermetic
         type: string
-      - default: ""
+      - default: "gomod"
         description: Build dependencies to be prefetched by Cachi2
         name: prefetch-input
         type: string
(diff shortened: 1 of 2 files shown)


.tekton/multiarch-tuning-operator-pull-request.yaml
```diff
@@ -102,7 +102,7 @@ spec:
         description: Execute the build with network isolation
         name: hermetic
         type: string
-      - default: ""
+      - default: "gomod"
```
aleskandro, line 105: Would this cover https://github.com/openshift/multiarch-tuning-operator/blob/main/Makefile#L280 too?

aleskandro: This shouldn't pass for the build of the bundle until we fix the first stage not to use controller-gen (or other). xref: https://redhat-internal.slack.com/archives/C3VS0LV41/p1714483831438239?thread_ts=1714392810.215579&cid=C3VS0LV41

aleskandro:
/lgtm

/approve

we can work on the bundle later and have value from the changes to the operator build here.

Thanks @ralphbean

PR #108, opened 2024-05-10
:warning: Rename PodPlacementConfig to ClusterPodPlacementConfig

aleskandro:
Follow a pattern for which cluster-scoped resources/objects start with '[cC]luster'.


aleskandro:
Cc @lwan-wanglin

PR #111, opened 2024-05-17
Add mto testing for registry configuration

lwan-wanglin:
for [MULTIARCH-4311](https://issues.redhat.com/browse/MULTIARCH-4311)

Add e2e for registry configuration
- Create a proxy registry container in test namespace
- run test using image from that registry
- firstly adding three cases: the first one is adding registry url to insecureRegistries list, the second one is doing nothing, the third one is adding registry cert in the trusted anchors
- updating image.config will cause node rescheduling, extend test spec timeout to 30m
- need to get machineconfigpool resource to check node update finish, update `github.com/openshift/api` version


pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+	tmpDir := filepath.Join(os.TempDir(), fmt.Sprintf("mto-test-%s", getRandomString()))
+	err = os.MkdirAll(tmpDir, 0755)
+	if err != nil {
+		return "", "", err
+	}
+	caPath = filepath.Join(tmpDir, caFileName)
+	keyPath = filepath.Join(tmpDir, keyFileName)
+	Args := []string{"req", "-newkey", "rsa:4096", "-nodes", "-sha256", "-keyout", keyPath, "-x509", "-days", "3600",
+		"-out", caPath, "-addext", "subjectAltName=DNS:" + dns, "-subj", "/OU=MultiArch QE/CN="}
+	cmd := exec.Command("openssl", Args...)
```
aleskandro: Is there any way we can generate this certificate without using external tools?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+	cmd.Dir = tmpDir
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		fmt.Print(string(output))
+		return "", "", err
+	}
+	return keyPath, caPath, nil
+}
+
+func getRandomString() string {
```
aleskandro: Let's use a standard way for generating random strings: https://pkg.go.dev/github.com/google/uuid#UUID.String

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+		return nil, err
+	}
+	contentBytes := []byte(fileContent)
+	return contentBytes, nil
+}
+
+func waitForMCPComplete() {
+	var err error
+	// wait for 10s to make machineconfig working
+	time.Sleep(30 * time.Second)
```
aleskandro: I would use another Eventually here, that checks until the first MCP start rolling out (is not ready and updated)

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+// createRegistryCertificate creates a certificate and returns: (path to the key used to sign the CA, path to the CA, error)
+func createRegistryCertificate(dns string) (keyPath, caPath string, err error) {
+	var (
+		caFileName  = "registry.crt"
+		keyFileName = "registry.key"
+	)
+	if _, err := os.Stat(os.TempDir()); os.IsNotExist(err) {
+		//nolint:errcheck
+		//#nosec:G301 (CWE-276): Expect directory permissions to be 0750 or less (Confidence: HIGH, Severity: MEDIUM)
+		os.Mkdir(os.TempDir(), 1777)
```
aleskandro: 1770 should work too, but if there's some other code doing this already (iirc, yes), let's keep the same permission bits as that one.

pkg/testing/builder/service_port_builder.go
```diff
@@ -0,0 +1,41 @@
...
+	return s.WithProtocol("TCP")
+}
+
+func (s *ServicePortBuilder) WithPort(value int32) *ServicePortBuilder {
+	s.servicePort.Port = value
+
+	return s
+}
+
+func (s *ServicePortBuilder) WithTargetPortInt32(value int32) *ServicePortBuilder {
```
aleskandro:
```suggestion
func (s *ServicePortBuilder) WithTargetPort(value int32) *ServicePortBuilder {
```

Any reason to make the int32 type explicitly stated in the name?

pkg/e2e/podplacement/pod_placement_test.go
aleskandro: Something like https://stackoverflow.com/questions/43822945/golang-can-i-create-x509keypair-using-rsa-key or https://go.dev/src/crypto/tls/generate_cert.go

pkg/testing/builder/container_builder.go
```diff
@@ -40,6 +40,36 @@ func (c *ContainerBuilder) WithVolumeMounts(volumeMounts ...v1.VolumeMount) *Con
 	return c
 }
 
+func (c *ContainerBuilder) WithEnv(values ...*v1.EnvVar) *ContainerBuilder {
```
aleskandro: If you don't use a pointer here and at line 48, and then you Build directly into the WithEnv method (without using env1, env2, env3), would it work?

You would then use the builder as in `WithEnv(NewContainerEnv()...., NewContainerEnv()....)` or `WithEnv(NewContainerEnv()....Build()).WithEnv(NewContainerEnv()....Build)`

pkg/testing/builder/container_builder.go
```diff
@@ -40,6 +40,36 @@ func (c *ContainerBuilder) WithVolumeMounts(volumeMounts ...v1.VolumeMount) *Con
...
+	for i := range values {
+		if values[i] == nil {
+			panic("nil value passed to WithEnv")
+		}
+		c.container.Env = append(c.container.Env, *values[i])
+	}
+	return c
+}
+
+func (c *ContainerBuilder) WithPorts(values ...*v1.ContainerPort) *ContainerBuilder {
```
aleskandro: Same discussion about the pointers.... What's the reason? Can we avoid it?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
 		}
 	}, e2e.WaitShort).Should(Succeed())
 }
+
+func createRegistryContainer(ns *corev1.Namespace, registryProxyUrl, registryProxyUser, registryProxyPasswd string) (registry string, err error) {
```
aleskandro: `s/createRegistryContainer/createRegistry/`

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+		WithNameSpace(ns.Name).
+		WithPorts(&servicePort).
+		WithSelector(registryLabel).
+		Build()
+	err = client.Create(ctx, &service)
+	if err != nil {
+		return "", err
+	}
+	// Create route
+	route := NewRoute().
```
aleskandro: Can we use an Ingress instead of a Route? (If I remember correctly, it will then generate the Route in OCP, but we keep compatibility with non ocp clusters)

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +814,186 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+		WithNameSpace(ns.Name).
+		Build()
+	err = client.Create(ctx, &secret)
+	if err != nil {
+		return "", err
+	}
+	// Create registry pod
+	vm1 := NewVolumeMount().WithName("registry-storage").WithMountPath("/var/lib/registry").Build()
+	vm2 := NewVolumeMount().WithName("registry-secret").WithMountPath("/etc/secrets").Build()
+	env1 := NewContainerEnv().WithName("REGISTRY_HTTP_ADDR").WithValue(":5001").Build()
```
aleskandro: See discussion below on whether to use pointers or not.

pkg/testing/builder/service_port_builder.go
lwan-wanglin: no specific reason, I will remove it.

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: sure

pkg/e2e/podplacement/pod_placement_test.go

pkg/e2e/podplacement/pod_placement_test.go
Prashanth684: that's a good point - but maybe k8s tests live in another suite? we might eventually be  adding tests related to icsp/idms etc...which are ocp specific anyways

aleskandro: Yes, I was thinking of adding some skips on those OCP-specific test cases.

aleskandro: But yeah, we might eventually decide to diverge some bits here to keep things clearaer against both OCP and non-OCP k8s distros.

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: yes, we have it in [registry.go](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/testing/image/fake/registry/registry.go#L134) and with 1777 permission.

hack/ci-test.sh
```diff
@@ -23,7 +23,7 @@ REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
 OPENSHIFT_CI=${OPENSHIFT_CI:-""}
 ARTIFACT_DIR=${ARTIFACT_DIR:-""}
 GINKGO=${GINKGO:-"go run ${REPO_ROOT}/vendor/github.com/onsi/ginkgo/v2/ginkgo"}
-GINKGO_ARGS=${GINKGO_ARGS:-"-vv --randomize-all --randomize-suites -race -trace --keep-going --timeout=10m "}
+GINKGO_ARGS=${GINKGO_ARGS:-"-vv --randomize-all --randomize-suites -race -trace --keep-going --timeout=30m "}
```
aleskandro, line 26: Do you know which test is taking so long to require 20m more in the deadline?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -618,6 +634,214 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+				registryName string = "untrusted-registry"
+				err          error
+			)
+			ns := framework.NewEphemeralNamespace()
+			err = client.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer client.Delete(ctx, ns)
+			// Create registry
+			dns := ocpconfigv1.DNS{}
```
aleskandro: I see a lot of redundant code, any way we can move it to some functions to reduce this redundancy

aleskandro: Also, what's the purpose of the DNS object?

hack/ci-test.sh
aleskandro: Oh, maybe the MCP updates?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +945,255 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
 		}
 	}, e2e.WaitShort).Should(Succeed())
 }
+
+func createRegistry(ns *corev1.Namespace, registryName, registryHost, keyPath, caPath, registryProxyUrl, registryProxyUser, registryProxyPasswd string) {
```
aleskandro: can this be moved to a testing/registry package?

Also, maybe this can be a struct such that we can:
1. create the registry object (as in creating a new ref to the object `&Registry{ns: ..., ...}`
2. deploy it (as in calling a `registry.Deploy(&client)` method`
3. destroy it `registry.Deprovision(&client)`
4. include any internal and non-exported methods (e.g. create cert) keeping them for internal use.

aleskandro: Having shorter functions would also be nicer.

pkg/e2e/podplacement/pod_placement_test.go
aleskandro: If it is to find the base domain, can we do this in the beforeSuite and store it in a variable to be used later by each case instead of doing this get during the execution of every test case?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -618,6 +634,214 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+				err = client.Update(ctx, &image)
+				Expect(err).NotTo(HaveOccurred())
+				waitForCOComplete()
+			}()
+			waitForCOComplete()
+			d := NewDeployment().
+				WithSelectorAndPodLabels(podLabel).
+				WithPodSpec(
+					NewPodSpec().
+						WithContainersImages(fmt.Sprintf(`%s/multi-arch/tuning-test-local:1.1.0`, registryHost)).
```
aleskandro: Should we use a method `getMirroredImageURI(registry, image string)` to generalize this and store `multi-arch/tuning-test-local:1.1.0` in a constant?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +945,255 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+		WithOpaqueType().
+		WithName(registryName).
+		WithNameSpace(ns.Name).
+		Build()
+	err = client.Create(ctx, &secret)
+	Expect(err).NotTo(HaveOccurred())
+	// Create ingress
+	tlsAnnotations := map[string]string{
+		"route.openshift.io/destination-ca-certificate-secret": secret.Name,
+		"route.openshift.io/termination":                       "reencrypt",
```
aleskandro: Can we get this from https://github.com/openshift/api/blob/b9885cc0c986b493f6ea5f4b6fc2583fe7182cd1/route/v1/types.go#L496?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -721,3 +945,255 @@ func verifyPodLabels(ns *corev1.Namespace, labelKey string, labelInValue string,
...
+	err = client.Create(ctx, &ingress)
+	Expect(err).NotTo(HaveOccurred())
+	// Create registry
+	registryD := NewDeployment().
+		WithSelectorAndPodLabels(registryLabel).
+		WithPodSpec(
+			NewPodSpec().
+				WithContainers(
+					NewContainer().
+						WithImage("registry:2").
```
aleskandro: Can we use one an image from Quay?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -618,6 +634,214 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			defer func() {
+				image := ocpconfigv1.Image{}
+				err = client.Get(ctx, runtimeclient.ObjectKeyFromObject(&ocpconfigv1.Image{
+					ObjectMeta: metav1.ObjectMeta{
+						Name: "cluster",
+					},
+				}), &image)
+				Expect(err).NotTo(HaveOccurred(), "failed to get image.config cluster", err)
+				image.Spec = imageForRemove.Spec
```
aleskandro, line 658: uhm, this is not concurrent safe. I know it's a bit odd to remove by linear search, but we should implement it with a retry mechanism as well as in calling a method `removeFromImageConfig(registryHost string)` that:

1. Get the image config
2. filter out the registryHost from any lists (insecure,allowed,...)
3. try to update the image config
4. if the update fail (because something has updated it in the meantime), retry possibly using an exponential backoff (eg., retry after 20ms, then 40ms, 80ms, ....) and a relatively low maximum amount of retries (5/10) after which we panic

hack/ci-test.sh
lwan-wanglin: yes, mcp updating needs long time

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: thanks, that's a good suggestion.

pkg/e2e/podplacement/pod_placement_test.go

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: sure, thanks

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: ok, will fix it.

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: I'll think about it.

lwan-wanglin: according to my analysis, it's complex for image.config based testcases to run in Parallel. 

1. if two testcases update image.config CR within 5 min, when the second testcases run, the mcp will start the second reconcile and node start rescheduling, it might cause the first testcase mcp status check hit timeout, and it's also hard to choose a suitable timeout for mcp reconcile.
2. allowedRegistry and BlockedRegistry are conflict, but there is no error throw when we set the two field at once, and the cluster will ignore the second config.

lwan-wanglin:
@aleskandro could you help review again when you are available?thanks

pkg/e2e/podplacement/pod_placement_test.go
aleskandro: > allowedRegistry and BlockedRegistry are conflict, but there is no error throw when we set the two field at once, and the cluster will ignore the second config.

This actually looks like an OCP bug to me.

I'm ok with this implementation for now.

aleskandro (commented):
Hi @lwan-wanglin , there are still some bits I would simplify, but I'm ok to merge this now and unblock the other PR. We can have a chat later or a story that links this PR and do some changes later, if we'll think they are worth it.

Thanks for all this work!

/lgtm
/approve

PR #115, opened 2024-05-21
MULTIARCH-4635: Re-enable hermetic builds

aleskandro:
This PR enables hermetic builds and fixes the prefetch-input parameters. 

As even the bundle will be built on a disconnected environment and since the Makefile `bundle` target generated by operator-sdk strongly relies on ephemeral binaries downloaded from the internet that cachi2 cannot address, this PR also changes the way we rebuild the image references in the bundle from the `bundle` Makefile target to a regex/sed statement.


aleskandro:

fixing in #118

PR #117, opened 2024-05-21
MULTIARCH-4598: Use a single service account for operator and operands

AnnaZivkovic:
Have a single service account "multiarch-tuning-operator-controller-manager" for both operands and operator.  This is a step in allowing the operator to be installed in namespaces other than default (openshift-multiarch-tuning-operator)  as discovered by [on-boring work for konflux](https://github.com/openshift/multiarch-tuning-operator/pull/70).


aleskandro:
>  Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-operator Failing after 47s — Failed 

Can you rebase on the current main (just merged an update from the konflux bot)

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -149,12 +150,10 @@ func (r *ClusterPodPlacementConfigReconciler) reconcile(ctx context.Context, clu
 	log := ctrllog.FromContext(ctx)
 	objects := []client.Object{
 		buildDeployment(clusterPodPlacementConfig, PodPlacementControllerName, 2,
-			"multiarch-tuning-operator-podplacement-controller",
 			"--leader-elect",
```
Prashanth684, line 154: shouldn't the `serviceAccountName` be used here?

aleskandro:
> > Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-operator Failing after 47s — Failed
> 
> Can you rebase on the current main (just merged an update from the konflux bot)

Fixing in #118

aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-operator Failing after 47s — Failed"

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: We were discussing yesterday about this and decided to go for not continuing to have the serviceAccountName as a parameters as it currently is always the same and we can have a simplified signature for the method until possible.

aleskandro: See the move to a constant for now at line 58

aleskandro:
/approve

if we are good with https://github.com/openshift/multiarch-tuning-operator/pull/117#discussion_r1609018917, I'm ok to merge. Just want both the olm and non-olm tests to succeed first due to the specificity of the PR.

aleskandro:

aleskandro:
/hold

one case failing in the e2e-gcp-multi-operator test: Multiarch Tuning Operator Suite (PodPlacementOperand E2E): [It] The Pod Placement Operand PodPlacementOperand works with several high-level resources owning pods should set the node affinity on Build owning pod

aleskandro:
@AnnaZivkovic did you also try deploying the operator in a different namespace than the default one now?

Prashanth684:
> @AnnaZivkovic did you also try deploying the operator in a different namespace than the default one now?

this is also a good use case I would add to QE's list of tests.

aleskandro:
[guessing] I believe we are missing some SCC rolebinding because build pods should be privileged. I don't remember if we even have a dedicated test for privileged pods (@lwan-wanglin do we?)

lwan-wanglin:
> [guessing] I believe we are missing some SCC rolebinding because build pods should be privileged. I don't remember if we even have a dedicated test for privileged pods (@lwan-wanglin do we?)

I think we have one case for privileged pod testing https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/e2e/podplacement/pod_placement_test.go#L72

Prashanth684:
> [guessing] I believe we are missing some SCC rolebinding because build pods should be privileged. I don't remember if we even have a dedicated test for privileged pods (@lwan-wanglin do we?)

Does the deplyment's [permissions](https://github.com/openshift/multiarch-tuning-operator/pull/117/files#diff-b49b727fb9ff793be2a463efc100d397b8194bfad75ab2fa55a1454e10c02273R409) need to have scc included?

Prashanth684:
was able to debug this a bit further - the apigroup is incorrect in the spec and the seccomp profile for the manager pod needs to be added:

```
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -105,12 +105,6 @@ rules:
   - get
   - list
   - watch
-- apiGroups:
-  - ""
-  resources:
-  - securitycontextconstraints
-  verbs:
-  - use
 - apiGroups:
   - ""
   resources:
@@ -163,3 +157,9 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  verbs:
+  - use
```

```
--- a/controllers/podplacement/pod_reconciler.go
+++ b/controllers/podplacement/pod_reconciler.go
@@ -42,7 +42,7 @@ type PodReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;update;patch
 //+kubebuilder:rbac:groups=core,resources=pods/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
-//+kubebuilder:rbac:groups=core,resources=securitycontextconstraints,verbs=use
+//+kubebuilder:rbac:groups=security.openshift.io,resources=securitycontextconstraints,verbs=use
 //+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
```


```
--- a/config/default/manager_auth_proxy_patch.yaml
+++ b/config/default/manager_auth_proxy_patch.yaml
@@ -8,6 +8,10 @@ metadata:
 spec:
   template:
     spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:

```

This fixes the affinity, but the test itself for me was failing because the build pod failed with:

```
/bin/sh: /usr/libexec/s2i/assemble: not found
          error: build error: building at STEP "RUN /usr/libexec/s2i/assemble": while running runtime: exit status 127

```
but not sure if it was something I was missing. @AnnaZivkovic if you could  make these changes, we can see how the test runs.

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -133,12 +133,76 @@ spec:
...
+        - apiGroups:
+          - ""
+          resources:
+          - secrets
+          verbs:
+          - get
+          - list
+          - watch
+        - apiGroups:
+          - ""
```
Prashanth684, line 201: change api group here as well?

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -44,7 +44,7 @@ metadata:
     operatorframework.io/suggested-namespace: openshift-multiarch-tuning-operator
     operators.openshift.io/valid-subscription: '["OpenShift Kubernetes Engine", "OpenShift
       Container Platform", "OpenShift Platform Plus"]'
-    operators.operatorframework.io/builder: operator-sdk-v1.31.0
+    operators.operatorframework.io/builder: operator-sdk-v1.33.0
```
aleskandro:
```suggestion
    operators.operatorframework.io/builder: operator-sdk-v1.31.0
```

the bundle job uses the downstream version of operator-sdk.

PR #121, opened 2024-05-22
MULTIARCH-4315: Implements workloads partitioning annotations

aleskandro:
See https://github.com/openshift/enhancements/blob/c5b9aea25e2a7afd5f460c6a0eca7584b1685ce3/enhancements/workload-partitioning/management-workload-partitioning.md

Related to https://github.com/openshift/release/pull/51275

### DIFF
--- a/bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
@@ -26,7 +26,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: OpenShift Optional, Other
     console.openshift.io/disable-operand-delete: "true"
-    createdAt: "2024-05-10T17:21:55Z"
+    createdAt: "2024-05-22T19:17:23Z"
     features.operators.openshift.io/cnf: "false"
     features.operators.openshift.io/cni: "false"
     features.operators.openshift.io/csi: "false"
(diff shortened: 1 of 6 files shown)


Prashanth684:
Do we need e2e for these ? i would like to see some performance related testing on SNO and the like to really qualify this use case.

aleskandro:
> Do we need e2e for these ? i would like to see some performance related testing on SNO and the like to really qualify this use case.

I think the operator still qualifies for the use case of sno + multiple workers in the edge scenarios in particular. I also want to read some data asap to better tune some choices we've made at first.

aleskandro:

(we just need new snapshots)

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -270,9 +272,6 @@ spec:
                   initialDelaySeconds: 5
                   periodSeconds: 10
                 resources:
-                  limits:
```
Prashanth684, line 273: why remove the limits?

aleskandro: it's a requirement from origin to have all the pods in openshift-* to be in the `burstable` QoS class

Prashanth684: right right..or is it besteffort? but yes, got it

PR #144, opened 2024-06-20
Fix tekton pipelines' step deprecated-image-check

aleskandro:
The deprecated image check step doesn't process each single arch manifest when using multiarch builds


aleskandro:
/override ci/prow/ci-index-multiarch-tuning-operator-bundle ci/prow/images

PR #177, opened 2024-07-17
MULTIARCH-4694: Enable insecure registries for mirror repositories

AnnaZivkovic:
This is part of the disconnected work

### DIFF
--- a/pkg/systemconfig/syncer.go
+++ b/pkg/systemconfig/syncer.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"os"
-	"strings"
 	"sync"
 
 	"github.com/containers/image/v5/signature"
@@ -94,10 +93,7 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string, 
 	// Propagate insecure registries to the mirrors of any previously defined registry set by ICSP/ITMS/IDMS.
 	for _, rc := range s.registriesConfContent.Registries {
 		for _, mirror := range rc.Mirrors {
-			mirror.Insecure = nil
-			if mrc, ok := s.registriesConfContent.getRegistryConf(strings.Split(mirror.Location, "/")[0]); ok && mrc.Insecure != nil {
-				mirror.Insecure = mrc.Insecure
-			}
+			mirror.Insecure = s.registriesConfContent.checkLocationInsecurity(mirror.Location)
 		}
 	}
 
@@ -116,25 +112,11 @@ func (s *SystemConfigSyncer) StoreRegistryCerts(registryCertTuples []registryCer
 	return nil
 }
 
-func (s *SystemConfigSyncer) mirrorsFor(locations []string, pullType PullType) []*Mirror {
-	var mirrors []*Mirror
-	var insecure *bool
-
-	for _, location := range locations {
-		insecure = nil
-		if mrc, ok := s.registriesConfContent.getRegistryConf(strings.Split(location, "/")[0]); ok && mrc.Insecure != nil {
-			insecure = mrc.Insecure
-		}
-		mirrors = append(mirrors, mirrorFor(location, pullType, insecure))
-	}
-	return mirrors
-}
-
 func (s *SystemConfigSyncer) UpdateRegistryMirroringConfig(registry string, mirrors []string, pullType PullType) error {
 	s.mu.Lock()
 	defer s.unlockAndSync()
 	rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
-	rc.Mirrors = s.mirrorsFor(mirrors, pullType)
+	rc.Mirrors = s.registriesConfContent.newMirrorRegistries(mirrors, pullType)
 	return nil
 }
 
--- a/pkg/systemconfig/syncer.go
+++ b/pkg/systemconfig/syncer.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 
 	"github.com/containers/image/v5/signature"
@@ -89,6 +90,17 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string, 
 		rc.Insecure = &trueValue
 	}
 	s.registriesConfContent.cleanupAllRegistryConfIfEmpty()
+
+	// Propagate insecure registries to the mirrors of any previously defined registry set by ICSP/ITMS/IDMS.
+	for _, rc := range s.registriesConfContent.Registries {
+		for _, mirror := range rc.Mirrors {
+			mirror.Insecure = nil
+			if mrc, ok := s.registriesConfContent.getRegistryConf(strings.Split(mirror.Location, "/")[0]); ok && mrc.Insecure != nil {
+				mirror.Insecure = mrc.Insecure
+			}
+		}
+	}
+
 	return nil
 }
 
@@ -104,11 +116,25 @@ func (s *SystemConfigSyncer) StoreRegistryCerts(registryCertTuples []registryCer
 	return nil
 }
 
+func (s *SystemConfigSyncer) mirrorsFor(locations []string, pullType PullType) []*Mirror {
+	var mirrors []*Mirror
+	var insecure *bool
+
+	for _, location := range locations {
+		insecure = nil
+		if mrc, ok := s.registriesConfContent.getRegistryConf(strings.Split(location, "/")[0]); ok && mrc.Insecure != nil {
+			insecure = mrc.Insecure
+		}
+		mirrors = append(mirrors, mirrorFor(location, pullType, insecure))
+	}
+	return mirrors
+}
+
 func (s *SystemConfigSyncer) UpdateRegistryMirroringConfig(registry string, mirrors []string, pullType PullType) error {
 	s.mu.Lock()
 	defer s.unlockAndSync()
 	rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
-	rc.Mirrors = mirrorsFor(mirrors, pullType)
+	rc.Mirrors = s.mirrorsFor(mirrors, pullType)
 	return nil
 }
 
--- a/pkg/systemconfig/types.go
+++ b/pkg/systemconfig/types.go
@@ -19,6 +19,7 @@ package systemconfig
 import (
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -163,6 +164,25 @@ func (rsc *registriesConf) cleanupAllRegistryConfIfEmpty() {
 	}
 }
 
+func (rsc *registriesConf) newMirrorRegistries(locations []string, pullType PullType) []*Mirror {
+	var mirrors []*Mirror
+
+	for _, location := range locations {
+		mirrors = append(mirrors, mirrorFor(location, pullType, rsc.checkLocationInsecurity(location)))
+	}
+	return mirrors
+}
+
+func (rsc *registriesConf) checkLocationInsecurity(location string) *bool {
+	trueValue := true
+	for _, rc := range rsc.registriesMap {
+		if matchRegistry(location, rc.Location) && rc.Insecure != nil && *rc.Insecure != false {
+			return &trueValue
+		}
+	}
+	return nil
+}
+
 type registryConf struct {
 	Location string    `toml:"location"`
 	Prefix   string    `toml:"prefix"`
@@ -184,6 +204,65 @@ func mirrorFor(location string, pullType PullType, insecure *bool) *Mirror {
 		PullFromMirror: pullType,
 		Insecure:       insecure,
 	}
+}
+
+// matchRegistry checks if a registry matches a mirror pattern
+func matchRegistry(mirror, registry string) bool {
+	mirrorHostPort, mirrorPath := splitHostAndPath(mirror)
+
+	registryHostPort, registryPath := splitHostAndPath(registry)
+
+	// Match the host[:port] part
+	if !matchHostPort(mirrorHostPort, registryHostPort) {
+		return false
+	}
+
+	// Match the path part (namespace, repo, tag/digest)
+	return matchPath(mirrorPath, registryPath)
+}
+
+// splitHostAndPath splits the host:port and path components
+func splitHostAndPath(input string) (string, string) {
+	parts := strings.SplitN(input, "/", 2)
+	hostPort := parts[0]
+	path := ""
+	if len(parts) > 1 {
+		path = parts[1]
+	}
+	return hostPort, path
+}
+
+// matchHostPort checks if a host[:port] string matches another, considering wildcards
+func matchHostPort(mirrorHostPort, registryHostPort string) bool {
+	// Handle wildcards in the host part
+	if strings.Contains(registryHostPort, "*") {
+		matched, _ := path.Match(registryHostPort, mirrorHostPort)
+		return matched
+	}
+
+	return mirrorHostPort == registryHostPort
+}
+
+// matchPath checks if the path components match correctly
+func matchPath(mirrorPath, registryPath string) bool {
+	// Split paths into segments
+	mirrorSegments := strings.SplitAfter(mirrorPath, "/")
+	registrySegments := strings.SplitAfter(registryPath, "/")
+
+	if registrySegments[len(registrySegments)-1] == "" {
+		registrySegments = registrySegments[:len(registrySegments)-1]
+	}
+	// Compare each segment
+	for i := range mirrorSegments {
+		if i > len(registrySegments)-1 {
+			// Mirror is more specific than registry
+			break
+		}
+		if mirrorSegments[i] != registrySegments[i] {
+			return false
+		}
+	}
+	return true
 }
 
 // defaultRegistriesConf returns a default registriesConf object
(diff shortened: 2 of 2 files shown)


pkg/systemconfig/syncer.go
```diff
@@ -108,7 +125,7 @@ func (s *SystemConfigSyncer) UpdateRegistryMirroringConfig(registry string, mirr
 	s.mu.Lock()
 	defer s.unlockAndSync()
 	rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
-	rc.Mirrors = mirrorsFor(mirrors, pullType)
+	rc.Mirrors = mirrorsFor(mirrors, pullType, rc.Insecure)
```
aleskandro: This function is called when a new ImageContentSource, ImageDigestMirrorSet or ImageTagMirrorSet is created, right?. You are setting insecure for all the mirrors to the same value as the mirrored registry.

If I remember and understand correctly, we should look for each mirror's specific conf in the 'first-level registry' (using `getRegistryConf`). If the getRegistryConf returns its conf and that conf has insecure=True, we should set that specific mirror to be insecure too.

aleskandro:
```suggestion
	rc.Mirrors = mirrorsFor(mirrors, pullType)
	trueValue := true
	for _, mirror := range rc.Mirrors {
	  if mrc, ok := s.registriesConfContent.getRegistryConf(mirror); ok && mrc.insecure != nil && *mrc.insecure == true {
	    mirror.Insecure = &trueValue
	  }
	}
	
```

pkg/systemconfig/syncer.go
```diff
@@ -84,9 +88,22 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
...
+
 	for _, registry := range insecureRegistries {
 		rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
 		rc.Insecure = &trueValue
+
+		for _, mirror := range rc.Mirrors {
+			if insecureRegistryMap[mirror.Location] {
+				mirror.insecure = &trueValue
+			}
+		}
```
aleskandro: This shouldn't happen only here, but for all the defined registries. Therefore, after the last for loop instead of nested in the insecure registries' one.

After you have that map filled, we need to loop through all the registries in registriesConfContent.

For each registry (maybe we can't loop through all the registries yet), for each mirror, if that mirror has a registryConf (use getRegistryConf) and it has insecure not nil and true, we set that mirror to insecure true.

aleskandro (changes requested):
Hi @AnnaZivkovic, thanks!

This PR needs some work yet. There are two concurrent types of updates that can happen:

1. updating the insecure registries (let's say keeping constant the ITMS/IDMS/ICSPs): in this case, we need to loop through all the defined registries and set each of their mirrors to insecure if needed. => Start Point: image.config controller onAddOrUpdate https://github.com/openshift/multiarch-tuning-operator/blob/25f624994e9cf6e20bb2515ae93e2a133a1797df/ controllers/podplacement/image_config.go#L122-L124 => The controller uses StoreImageRegistryConf.
2. updating an ITMS/IDMS/ICSP: in this case we should look for its mirrors' conf. If a mirror is part of the insecureRegistries set, its 'insecure' field must be set to true. -> Starting point for ICSP: https://github.com/openshift/multiarch-tuning-operator/blob/25f624994e9cf6e20bb2515ae93e2a133a1797df/controllers/podplacement/image_content_source_policy.go#L98-L99 (other similar starting points for IDMS/ITMS)

We also need to cover this code with some test cases (both unit and integration).

Let's start from some unit cases that execute the changed functions given different initial conditions for the involved objects. For example:
1. No insecureRegistries
2. Insecure registries set. Add a new registry that is NOT insecure, with an insecure Mirror. Expectation: the registry has no insecure field, the mirror has.
3. Insecure registries set. Add a new registry that is insecure with a secure mirror. Expection: the registry has insecure set to true and the mirror has no insecure field.
4. Insecure registries set including registry Y; a registry X set with an insecure mirror Y. Remove the registry Y from the insecure registries and let the update go. We expect the registry Y is not anymore set in the main list, and the mirror's instance of registry Y for registry X has no inseucre field anymore.
5. Insecure registries set. Add a registry that is not in the insecure registries set, with mirrors not in the insecure registries set. Expectation: the registry is added with no insecure field and its mirrors have no insecure field either. The insecure registries are still set as they were initially.

pkg/systemconfig/syncer.go
Prashanth684: it should be set only for allowed registries right? we still need to make sure we don't do it for blocked ones

aleskandro: I think we should do for them too. They are two independent configurations: blockedregistries will land into policy.json

pkg/systemconfig/types.go
```diff
@@ -175,21 +175,21 @@ type registryConf struct {
...
-		// insecure: insecure,
+		Insecure:       insecure,
 	}
 }
 
 func mirrorsFor(locations []string, pullType PullType) []Mirror {
 	var mirrors []Mirror
 	for _, location := range locations {
-		mirrors = append(mirrors, mirrorFor(location, pullType))
+		mirrors = append(mirrors, mirrorFor(location, pullType, nil))
```
AnnaZivkovic: Is this an acceptable default value for mirror.Insecure?

aleskandro: do we ever pass a non nil value to this function?

pkg/systemconfig/syncer.go
```diff
@@ -76,17 +86,49 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
 			signature.NewPRInsecureAcceptAnything(),
 		}
 		rc.Blocked = nil
+
+		for _, mirror := range rc.Mirrors {
+			if allowedRegistriesMap[mirror.Location] {
```
aleskandro, line 91:
```suggestion
			if _, ok := allowedRegistriesMap[mirror.Location]; ok {
```

pkg/systemconfig/syncer.go
```diff
@@ -109,6 +151,12 @@ func (s *SystemConfigSyncer) UpdateRegistryMirroringConfig(registry string, mirr
 	defer s.unlockAndSync()
 	rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
 	rc.Mirrors = mirrorsFor(mirrors, pullType)
+	trueValue := true
+	for _, mirror := range rc.Mirrors {
+		if mrc, ok := s.registriesConfContent.getRegistryConf(mirror.Location); ok && mrc.Insecure != nil && *mrc.Insecure {
+			mirror.Insecure = &trueValue
+		}
+	}
```
aleskandro: Should we maybe move this to `mirrorsFor` (and have `mirrorsFor` as a method of `SystemConfigSyncer` rather than a function)?

Doing it here means that mirrorsFor will not be fully responsible for defining the mirrors but some other logic must be implemented 'always' after that method is called, violating "separation of concerns".

pkg/systemconfig/syncer.go
```diff
@@ -76,17 +86,49 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
...
 		rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
 		rc.Insecure = &trueValue
+
+		for _, mirror := range rc.Mirrors {
+			if insecureRegistryMap[mirror.Location] {
+				mirror.Insecure = &trueValue
+			}
+		}
 	}
 	s.registriesConfContent.cleanupAllRegistryConfIfEmpty()
```
aleskandro: I see some value in having an increased in-memory complexity vs an increased processing complexity, but we are having this snippet of code copied over 4 times.

Also, we already have a lookup map to support this 'faster' lookup strategy.

Can we just do the following after `cleanupAllRegistryConfIfEmpty()` call, once the comment in https://github.com/openshift/multiarch-tuning-operator/pull/177/files#r1705717318 for mirrorsFor is addressed?:

```suggestion
	s.registriesConfContent.cleanupAllRegistryConfIfEmpty()
	// Propagate insecure registries to the mirrors of any previously defined registry set by ICSP/ITMS/IDMS.
	for _, rc := range s.registriesConfContent.Registries {
		if len(rc.Mirrors) > 0 {
			rc.Mirrors = s.mirrorsFor(rc.mirrors, rc.mirrors[0].pullType)
		}
	}
```

pkg/systemconfig/syncer.go
```diff
@@ -76,17 +86,49 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
 			signature.NewPRInsecureAcceptAnything(),
 		}
 		rc.Blocked = nil
+
+		for _, mirror := range rc.Mirrors {
+			if allowedRegistriesMap[mirror.Location] {
+				mirror.Insecure = &trueValue
```
aleskandro: Mmm, even re-reading, we should set mirror.Insecure  to true only if mirror is in the insecureRegistries. This would set Insecure to true once any conf is previously set for that registry. 

I would really go ahead with:
1. https://github.com/openshift/multiarch-tuning-operator/pull/177/files#r1705717318
2. https://github.com/openshift/multiarch-tuning-operator/pull/177/files#r1705735790

pkg/systemconfig/types.go
aleskandro: maybe, once this function becomes a method of SystemConfiSyncer (see https://github.com/openshift/multiarch-tuning-operator/pull/177/files#r1705717318)

```suggestion
		insecure = nil // to define before the for loop as var insecure *bool
		if  s.registriesConfContent.getRegistryConf(mirror.Location); ok && mrc.Insecure != nil {
		 insecure = mrc.Insecure
		}
		mirrors = append(mirrors, mirrorFor(location, pullType, insecure))
```

aleskandro:
@AnnaZivkovic the CI is blocked by https://github.com/openshift/release/pull/55145 (revert PR at https://github.com/openshift/release/pull/55260)

pkg/systemconfig/syncer.go
```diff
@@ -89,6 +93,16 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
 		rc.Insecure = &trueValue
 	}
 	s.registriesConfContent.cleanupAllRegistryConfIfEmpty()
+
+	// Propagate insecure registries to the mirrors of any previously defined registry set by ICSP/ITMS/IDMS.
+	for _, rc := range s.registriesConfContent.Registries {
+		for _, mirror := range rc.Mirrors {
```
aleskandro, line 95:
```suggestion
		for _, mirror := range rc.Mirrors {
			mirror.Insecure = nil // reset it everytime before syncing with the insecureRegistries set
```

aleskandro (commented):
This looks good to me... Can we develop some integration or unit tests before merging?

aleskandro:
Also adding Lin for some premerge testing here:

/cc @lwan-wanglin

pkg/systemconfig/syncer.go
AnnaZivkovic: I was doing it [here](https://github.com/openshift/multiarch-tuning-operator/blob/b06240250b54e9dc9cae5f1c7b8ee36ecf89e722/pkg/systemconfig/syncer.go#L64-L66). That is where we are resetting everything else, but it requires an extra loop though the mirrors

aleskandro: oh, I lost that...

lwan-wanglin:
detailed test discussion is in slack channel: https://redhat-internal.slack.com/archives/C06ECDGH4LE/p1723550648966339

pkg/systemconfig/syncer.go
```diff
@@ -89,6 +90,17 @@ func (s *SystemConfigSyncer) StoreImageRegistryConf(allowedRegistries []string,
...
+			mirror.Insecure = nil
+			if mrc, ok := s.registriesConfContent.getRegistryConf(strings.Split(mirror.Location, "/")[0]); ok && mrc.Insecure != nil {
```
AnnaZivkovic: I would have used [string.TrimSuffix](https://pkg.go.dev/strings#TrimSuffix), but it only works if the _end_ of the string matches the suffix provided. 
`strings.TrimSuffix("virthost.cluster-01.okd.on.massopen.cloud:5000/localimages/local-release-image", "/") ---> virthost.cluster-01.okd.on.massopen.cloud:5000/localimages/local-release-image`

`strings.TrimSuffix("virthost.cluster-01.okd.on.massopen.cloud:5000/localimages/local-release-image", "/local-release-image") ---> virthost.cluster-01.okd.on.massopen.cloud:5000/localimages`

aleskandro:
I'm ok with this. I'd wait @lwan-wanglin's results again since we're not implementing tests here.

pkg/systemconfig/syncer.go
aleskandro: We cannot split and take the first part I believe as users can still set just a repo (or even a single image or tag) of a registry as insecure. I think we'd set insecure=True for a Mirror m if a insecure registry r exists such that strings.HasPrefix(m, r) or (r starts with a wildcard (*) and r without the * prefix is included in the first part of m before a slash (/) like you did with split)

Example:

insecure registry: myregistry.com/myallowedinsecurerepo/

Mirror that shouldn't be set as insecure: myregistry.com/myotherrepo

Mirror that should be set insecure:
myregistry.com/myallowedinsecurrepo/myomage:latest

Mirror that should be set insecure:
myregistry.com/myallowedinsecurrepo

(Note the leading slash in the insecure registry. The last mirror would not match if we don't remove it)


Another insecure registry setting can be 
- *.mydomain.tld

These mirrors should be set insecure too (wildcard condition above):
sub4.sub3.mydomain.tld/myrepo

https://docs.openshift.com/container-platform/4.16/openshift_images/image-configuration.html

https://github.com/containers/image/blob/main/docs/containers-registries.conf.5.md#choosing-a-registry-toml-table

lwan-wanglin: in my today's testing, changing to using pointer fixed the previous found issue, we also need to fix above @aleskandro's comments.

pkg/systemconfig/types.go
```diff
@@ -163,10 +164,30 @@ func (rsc *registriesConf) cleanupAllRegistryConfIfEmpty() {
 	}
 }
 
+func (rsc *registriesConf) newMirrorRegistries(locations []string, pullType PullType, registriesConf []*registryConf) []*Mirror {
```
aleskandro: we don't need anymore `registriesConf` as a parameter, right?

pkg/systemconfig/types.go
```diff
@@ -163,10 +164,30 @@ func (rsc *registriesConf) cleanupAllRegistryConfIfEmpty() {
...
+	var mirrors []*Mirror
+
+	for _, location := range locations {
+		mirrors = append(mirrors, mirrorFor(location, pullType, rsc.isLocationInsecure(location, registriesConf)))
+	}
+	return mirrors
+}
+
+func (rsc *registriesConf) isLocationInsecure(mirror string, registriesConf []*registryConf) *bool {
```
aleskandro: Same as the previous, do we still need to pass `registriesConf` as a parameter?

Also, as you call the method `isLocationInsecure`, I'd suggest changing the parameter name from `mirror` to `location`.

Finally, given we have a dedicated function that returns whether a registry is insecure or not, I would instead return a bool value, not a pointer, and then, in mirrorFor, we can add 

```
var insecurePtr *bool := nil
if insecure {
  insecurePtr = &insecure
}
```

`isLocationInsecure` can be used later for other reasons, while the true vs nil behavior is strictly related to the Toml marshaling. Returning `nil` with the meaning of `false` is very specific to the marshaling issue and doesn't respond/fit with the name of the method for me.

pkg/systemconfig/syncer.go
```diff
@@ -108,7 +116,7 @@ func (s *SystemConfigSyncer) UpdateRegistryMirroringConfig(registry string, mirr
 	s.mu.Lock()
 	defer s.unlockAndSync()
 	rc := s.registriesConfContent.getRegistryConfOrCreate(registry)
-	rc.Mirrors = mirrorsFor(mirrors, pullType)
+	rc.Mirrors = s.registriesConfContent.newMirrorRegistries(mirrors, pullType, s.registriesConfContent.Registries)
```
aleskandro: How does `buildMirrorRegistries` or `makeMirrorRegistries` sound to you compared to `newMirrorRegistries`? the prefix `new` is widely used for initializing pointers and gives that idea: `newMirrorRegistries` does more than initializing the list.

pkg/systemconfig/types.go
```diff
@@ -163,10 +164,30 @@ func (rsc *registriesConf) cleanupAllRegistryConfIfEmpty() {
...
+	trueValue := true
+	falseValue := false
+	for _, rc := range registriesConf {
+		if matchRegistry(mirror, rc.Location) && rc.Insecure != nil && rc.Insecure != &falseValue {
```
aleskandro:
```suggestion
		if matchRegistry(mirror, rc.Location) && rc.Insecure != nil && *rc.Insecure != false {
```

You can't compare the two by address as they will always be different values (memory addresses)

pkg/systemconfig/types.go
```diff
@@ -175,23 +195,74 @@ type registryConf struct {
...
+	if strings.Contains(registryHostPort, "*") {
+		matched, _ := path.Match(registryHostPort, mirrorHostPort)
+		return matched
+	}
+
+	return mirrorHostPort == registryHostPort
+}
+
+// matchPath checks if the path components match correctly
+func matchPath(mirrorPath, registryPath string) bool {
```
Prashanth684, line 247: can we just use regexes for these?

PR #193, opened 2024-07-29
MULTIARCH-4309: Add test for ITMS

lwan-wanglin:
the pr implements below functions:
- adding builder for imagetagmirrorset CR creation
- moving func MCP verification to pkg/testing
- moving func co verification to pkg/testing
- Getting PPO and Webhook logs, and save them to files when test cases failed.
- make func getMirroredImageURI more flexible
- Test with ITMS
- Add label for long duration test cases
- Add operador-sdk cleanup


pkg/e2e/podplacement/e2e_test.go
```diff
@@ -28,15 +28,21 @@ import (
...
-	clientset *kubernetes.Clientset
-	ctx       context.Context
-	dns       = ocpconfigv1.DNS{}
-	suiteLog  = ctrl.Log.WithName("setup")
+	client           runtimeclient.Client
+	clientset        *kubernetes.Clientset
+	ctx              context.Context
+	dns              = ocpconfigv1.DNS{}
+	suiteLog         = ctrl.Log.WithName("setup")
+	fakeITMSRegistry = "my-fake-itms-registry.io"
```
Prashanth684: so currently, this test will fallback to pull from source which is quay.io because of the MirrorSourcePolicy being `AllowContactingSource` ? Are we also planning to test with `NeverContactSource` ?

aleskandro: We might also need some more changes in the business logic to support the two different approaches

Prashanth684: we should support when pulling from source or the mirror itself depending on the policy, yes.

lwan-wanglin: in this test, "my-fake-itms-registry.io" is as source, quay.io is as mirror. 
but I will consider both `AllowContactingSource` and `NeverContactSource` policy when organizing testing, thank you for your suggestion.

pkg/e2e/podplacement/e2e_test.go
```diff
@@ -165,3 +189,12 @@ func deleteCertificatesConfigmap(ctx context.Context, client runtimeclient.Clien
 		Expect(err).NotTo(HaveOccurred())
 	}
 }
+
+func waitForMCPComplete() {
```
aleskandro: any reason for this method to stay here and not in https://github.com/openshift/multiarch-tuning-operator/pull/193/files#diff-7bf84bbc60910d9faea665cadca3d90d34edffd61df8800d3b3b7a3d7d974315R49

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -937,6 +936,43 @@ var _ = Describe("The Pod Placement Operand", func() {
 			verifyPodLabels(ns, "app", "test", e2e.Present, schedulingGateLabel)
 		})
 	})
+	Context("When deploying workloads that image has a multiple image mirrors", Serial, func() {
```
```suggestion
	Context("When deploying workloads running images in registries that have mirrors configuration", Serial, func() {
```

do you mean this?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -937,6 +936,43 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+		It("should set the node affinity when image set mirror source in ImageTagMirrorSet", func() {
```
```suggestion
		It("should set the node affinity for images in repositories with mirrors set in an ImageTagMirrorSet", func() {
```

pkg/e2e/podplacement/e2e_test.go
```diff
@@ -28,15 +28,21 @@ import (
...
-	suiteLog  = ctrl.Log.WithName("setup")
+	client           runtimeclient.Client
+	clientset        *kubernetes.Clientset
+	ctx              context.Context
+	dns              = ocpconfigv1.DNS{}
+	suiteLog         = ctrl.Log.WithName("setup")
+	fakeITMSRegistry = "my-fake-itms-registry.io"
+	quayRegistry     = "quay.io"
+	repository       = "openshifttest"
+	itmsName         = "test-itms"
```
aleskandro: Can these be constant and stay together with other similar settings?

aleskandro (commented):
This is lgtm, just some comments to clarify the wording of the test cases titles.
based on https://github.com/openshift/multiarch-tuning-operator/pull/193/files#r1695522289, can you also clarify what case are we currently testing? 

Can you also ensure enough debugging messages are avaialble as discussed in the past?

pkg/e2e/podplacement/e2e_test.go
lwan-wanglin: sure, will fix

pkg/e2e/podplacement/e2e_test.go
lwan-wanglin: I thought we should have this Eventually verify in e2e code.  got it, I will move it to testing/framework too

aleskandro (commented):
Hey Lin, is it possible for you to make more commits?

1. Implementation of the test
2. Implementation of the logs reporting (with go.mod and go.sumchanges)
3. Vendoring separated from go.mod/go.sum

lwan-wanglin:
> Hey Lin, is it possible for you to make more commits?
> 
> 1. Implementation of the test
> 2. Implementation of the logs reporting (with go.mod and go.sumchanges)
> 3. Vendoring separated from go.mod/go.sum

Separate the code to three commits, please help to review the pr https://github.com/openshift/multiarch-tuning-operator/pull/239, thanks

pkg/e2e/podplacement/e2e_test.go
```diff
@@ -86,6 +118,11 @@ var _ = AfterSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Eventually(framework.ValidateDeletion(client, ctx)).Should(Succeed())
 	deleteCertificatesConfigmap(ctx, client)
+	By("Deleting ImageTagMirrorSet after testing and wait for machineconfig finishing updating")
+	itms := NewImageTagMirrorSet().WithName(e2e.ITMSName).Build()
+	err = client.Delete(ctx, itms)
+	Expect(err).NotTo(HaveOccurred())
+	framework.WaitForMCPComplete(ctx, client)
```
aleskandro: Do you think that if we create a few ITMS, IDMS, ICSP and registries BeforeSuite, we can improve the execution time? (it can be done later)

lwan-wanglin: yes, it can improve the time

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -703,9 +702,9 @@ var _ = Describe("The Pod Placement Operand", func() {
 				image.Spec = imageForRemove.Spec
 				err = client.Update(ctx, &image)
 				Expect(err).NotTo(HaveOccurred())
-				waitForMCPComplete()
+				framework.WaitForMCPComplete(ctx, client)
```
aleskandro: What if we remove this line in the defer nodes at [line 671](https://github.com/openshift/multiarch-tuning-operator/pull/193/commits/26b0d934ae8d85ee22f7379bb859f824a761581c#diff-dc7098c6cc728c7ae257d96dd92795647d0eec77673f82eb7319dad5cb18b7c9R671). We should not actually care about the fact that the MCP are rolled-back.

pkg/testing/framework/machineconfig.go
```diff
@@ -0,0 +1,62 @@
...
+
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+
+	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	ocpmachineconfigurationv1 "github.com/openshift/api/machineconfiguration/v1"
+)
+
+func VerifyMCPsAreUpdating(g Gomega, ctx context.Context, client runtimeclient.Client) {
```
aleskandro, line 17: I think we also have a verifyCO, should we move it too?

pkg/testing/framework/pod.go
```diff
@@ -1,10 +1,15 @@
 package framework
 
 import (
+	"bytes"
 	"context"
+	"fmt"
+	"log"
 
 	. "github.com/onsi/gomega"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
```
aleskandro: let's have the imports of our packages as the last ones.

Most of the code base should already follow this order:

1. Internal golang packages
2. k8s
3. sig.k8s
4. openshift
5. mto

pkg/testing/framework/pod.go
```diff
@@ -26,3 +31,68 @@ func VerifyPodsAreRunning(g Gomega, ctx context.Context, client runtimeclient.Cl
...
+	buf := new(bytes.Buffer)
+	_, err = buf.ReadFrom(podLogs)
+	if err != nil {
+		return "", fmt.Errorf("failed to read logs for pod %s: %w", podName, err)
+	}
+
+	return buf.String(), nil
+}
+
+func GetPodsLogToFile(ctx context.Context, clientset *kubernetes.Clientset, client runtimeclient.Client, namespace, labelKey, labelInValue, fileDir string) error {
```
aleskandro:
```suggestion

func StorePodsLog(ctx context.Context, clientset *kubernetes.Clientset, client runtimeclient.Client, namespace, labelKey, labelInValue, dir string) error {

File: pkg/testing/framework/utils.go [hunk 11]
```diff
@@ -84,3 +88,32 @@ func RegisterScheme(s *runtime.Scheme) error {
 	}
 	return nil
 }
+
+func WriteContentsToFile(fileDir, fileName, contents string) error {
```
aleskandro:
```suggestion
func WriteToFile(dir, fileName, content string) error {
```

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -52,6 +53,12 @@ var _ = Describe("The Pod Placement Operand", func() {
 		By("Operand ready. Executing the case")
 	})
 	AfterEach(func() {
+		if CurrentSpecReport().Failed() {
+			By("The test case failed, get the podplacement and podplacement webhook logs for debug")
+			// ignore err
+			_ = framework.GetPodsLogToFile(ctx, clientset, client, utils.Namespace(), "controller", utils.PodPlacementControllerName, os.Getenv("ARTIFACT_DIR"))
+			_ = framework.GetPodsLogToFile(ctx, clientset, client, utils.Namespace(), "controller", utils.PodPlacementWebhookName, os.Getenv("ARTIFACT_DIR"))
```
aleskandro: can we add the operator logs?

pkg/testing/framework/pod.go
```diff
@@ -96,3 +99,51 @@ func GetPodsLogToFile(ctx context.Context, clientset *kubernetes.Clientset, clie
...
+	}
+	return buf.String(), nil
+}
+
+func GetRegistriesConfFromPPCPodToFile(ctx context.Context, clientset *kubernetes.Clientset, client runtimeclient.Client, podRESTConfig *rest.Config, fileDir string) error {
+	pods, err := GetPodsWithLabel(ctx, client, utils.Namespace(), "controller", utils.PodPlacementControllerName)
+	if err != nil {
+		return err
+	}
+	command := []string{"/bin/cat", "/etc/containers/registries.conf"}
```
aleskandro: Do we really need this file? If so, can we just use the API to copy that?

pkg/e2e/podplacement/e2e_test.go
```diff
@@ -138,3 +139,11 @@ func deleteCertificatesConfigmap(ctx context.Context, client runtimeclient.Clien
 		Expect(err).NotTo(HaveOccurred())
 	}
 }
+
+func getReplacedImageURI(image, replacedRegistry string) string {
```
aleskandro: is this only valid for the pod placement test suite or can it be useful for others (and hence moved to the package framework)?

pkg/e2e/podplacement/e2e_test.go
```diff
@@ -87,6 +127,11 @@ var _ = AfterSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	Eventually(framework.ValidateDeletion(client, ctx)).Should(Succeed())
 	deleteCertificatesConfigmap(ctx, client)
+	By("Deleting ImageTagMirrorSet after testing and wait for machineconfig finishing updating")
+	itms := NewImageTagMirrorSet().WithName(e2e.ITMSName).Build()
+	err = client.Delete(ctx, itms)
+	Expect(err).NotTo(HaveOccurred())
+	framework.WaitForMCPComplete(ctx, client)
```
aleskandro: Maybe we can skip this wait too.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -952,6 +951,145 @@ var _ = Describe("The Pod Placement Operand", func() {
 			verifyPodNodeAffinity(ns, "app", "test", expectedNSTs)
 		})
 	})
+	Context("When deploying workloads running images in registries that have mirrors configuration", func() {
```
aleskandro, line 890: @lwan-wanglin with the recent changes, we can also relax the tests requirements for this feature. We will not reconcile ITMSes, IDMSes or ICSPs. We still need to make sure they are considered, but remember that we are now using a hostpath volume for it, especially if these tests require time.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -952,6 +951,145 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+		AfterEach(func() {
+			if CurrentSpecReport().Failed() {
+				By("The registry mirrors related test case failed, get registries.conf from podplacement controller pods for debugging")
+				_ = framework.GetRegistriesConfFromPPCPodToFile(ctx, clientset, client, cfg, os.Getenv("ARTIFACT_DIR"))
+			}
+		})
+		It("Should set node affinity when source registry is unavailable, mirrors working and AllowContactingSource enabled in a ImageTagMirrorSet", func() {
```
aleskandro: Can we tag the specs title with:
- [slow-5] if it potentially requires more than 5 minutes
- [slow-20] if it potentially require more than 20 minutes

Including the current ones for insecure, blocked and allowed registries.

aleskandro:
can you also apply this patch in a separate commit of this PR?

```
diff --git a/hack/ci-test.sh b/hack/ci-test.sh
index ffbc957f..23ce7692 100755
--- a/hack/ci-test.sh
+++ b/hack/ci-test.sh
@@ -37,7 +37,7 @@ if [ $HOME == "/" ]; then
 fi
 
 if [ "$OPENSHIFT_CI" == "true" ] && [ -n "$ARTIFACT_DIR" ] && [ -d "$ARTIFACT_DIR" ]; then # detect ci environment there
-  GINKGO_ARGS="${GINKGO_ARGS} --junit-report=junit_multiarch_tuning_operator.xml --output-dir=${ARTIFACT_DIR}"
+  GINKGO_ARGS="${GINKGO_ARGS} --junit-report=junit_multiarch_tuning_operator${JUNIT_SUFFIX:-}.xml --output-dir=${ARTIFACT_DIR}"
 fi
 
 if [ "$SKIP_COVERAGE" != "true" ] && [ -n "$ARTIFACT_DIR" ] && [ -d "$ARTIFACT_DIR" ]; then
diff --git a/hack/deploy-and-e2e.sh b/hack/deploy-and-e2e.sh
index 11ab6bfb..c3ce1943 100755
--- a/hack/deploy-and-e2e.sh
+++ b/hack/deploy-and-e2e.sh
@@ -37,6 +37,7 @@ if [ "${USE_OLM:-}" == "true" ]; then
   oc registry login || echo "[WARN] Unable to login the registry, this could be expected in non-Prow envs"
 
   export KUBECONFIG="${OLD_KUBECONFIG}"
+  export JUNIT_SUFFIX="-olm"
   operator-sdk run bundle "${OO_BUNDLE}" -n "${NAMESPACE}" --security-context-config restricted
 else
   make deploy IMG="${OPERATOR_IMAGE}"
@@ -50,3 +51,13 @@ oc wait pods -n ${NAMESPACE} \
   --for=condition=Ready=True
 
 make e2e
+
+if [ "${CLEANUP:-false}" == "false" ]; then
+  exit 0
+fi
+
+set +e
+operator-sdk cleanup multiarch-tuning-operator -n ${NAMESPACE}
+make undeploy
+oc delete --ignore-not-found --force namespace ${NAMESPACE}
+exit 0
```

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: that would be no harmful, fixing...

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: sure, will add it

pkg/testing/framework/pod.go
lwan-wanglin: We already use the nodes config file now, so I think we don't really need it, will remove it.

pkg/e2e/podplacement/e2e_test.go
lwan-wanglin: can we keep "the one" in the AfterSuite, we need to keep the cluster as normal becasue we might do origin or openshift-tests-private testing after mto e2e

pkg/e2e/podplacement/pod_placement_test.go
lwan-wanglin: the ITMS already move to beforeSuite, so the test won't take long time

pkg/e2e/podplacement/pod_placement_test.go

aleskandro:
@lwan-wanglin I can merge this but this suite now seems requiring around 10 minutes more. Is there anything we can do to further optimize the time required to test it?

lwan-wanglin:
> @lwan-wanglin I can merge this but this suite now seems requiring around 10 minutes more. Is there anything we can do to further optimize the time required to test it?

I have an idea in my mind, we can move this to beforeEach in the context, and create it if the ITMS don't exist, otherwise skip creation, then we can add slow-15 label on the spec belong to that context, we can skip the test by label, i can fix it in this pr.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1062,14 +1064,168 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+				utils.ArchLabelValue(utils.ArchitectureArm64), "",
+				utils.ArchLabelValue(utils.ArchitectureS390x), "",
+				utils.ArchLabelValue(utils.ArchitecturePpc64le), "",
+			)
+			By("The pod should be running")
+			Eventually(func(g Gomega) {
+				framework.VerifyPodsAreRunning(g, ctx, client, ns, "app", "test")
+			}, e2e.WaitShort).Should(Succeed())
+		})
+		It("Should set node affinity when source registry is unavailable, mirrors are working and NeverContactingSource enabled in a ImageTagMirrorSet [slow-15]", func() {
```
aleskandro: I would choose only one of the two, the one who is not the default behavior maybe

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1062,14 +1064,168 @@ var _ = Describe("The Pod Placement Operand", func() {
 			)
 		})
 	})
+	Context("When deploying workloads running images in registries that have mirrors configuration", func() {
+		BeforeEach(func() {
+			createTestITMSIfNotExist(ctx, client, e2e.ITMSName)
```
aleskandro: @lwan-wanglin can you outline the objects that we currently create in multiple test cases and that trigger the MCO?

I'm thinking that we could create more of them all at once in the BeforeSuite spec so that we can cover most of the cases.

For example, in the BeforeSuite, we can create:
1. static ITMS 
2. static IDMS
3. static ICSP
4. the `cluster` object for images.config.openshift.io with:
  a. BlockedRegistries: a blocked registry
  b. InsecureRegistries: a local and static insecure registry deployed in the BeforeSuite
  c. The config map with the additional Ca for another local registry (not part of the insecureRegistries list)

Then, we may only need to change again the conf above when testing the AllowedRegistries behavior (as it conflicts with BlockedRegistries).

Wdyt?

lwan-wanglin: option 1: (4 times WaitForMCPComplete)
if we create static ITMS/IDMS/ICSP and set image.config except AllowedRegistries in BeforeSuite, we will must trigger mco  twice in Before/AfterSuite  and can't skip it, then we need to run `framework.WaitForMCPComplete(ctx, client)` **four times**(two for AllowedRegistries, two for Before/AfterSuite)

option 2:(5 times WaitForMCPComplete and 1 time WaitForCOComplete(about 3 mins) )
if we won't create them in Suite, static ITMS/IDMS/ICSP can be created once in beforeEach() and trigger only one time WaitForMCPComplete, one time for BlockedRegistries, one time for InsecureRegistries, one time for BlockedRegistries one time, one time for AfterSuilte, and aditional one time WaitForCOComplete for The config map with the additional Ca about 3 mins

the both are ok for me, which one do your prefer?

aleskandro: I'd go for option 1, but:
1. BeforeSuite WaitForMCP ok
2. What if we drop the allowed registries test given that we are now using the hostPath volume?
3. I still see the AfterSuite waitForMCP as optional because any later step should be resilient to that.

lwan-wanglin: I think it's ok to drop the allowed registries, and I will also remove waitForMCP in AfterSuite.

hack/ci-test.sh
```diff
@@ -23,7 +23,7 @@ REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
 OPENSHIFT_CI=${OPENSHIFT_CI:-""}
 ARTIFACT_DIR=${ARTIFACT_DIR:-""}
 GINKGO=${GINKGO:-"go run ${REPO_ROOT}/vendor/github.com/onsi/ginkgo/v2/ginkgo"}
-GINKGO_ARGS=${GINKGO_ARGS:-"-vv --randomize-all --randomize-suites -race -trace --keep-going --timeout=60m "}
+GINKGO_ARGS=${GINKGO_ARGS:-"-vv --randomize-all --randomize-suites -race -trace --keep-going --timeout=30m "}
```
aleskandro, line 26: ❤️

pkg/testing/framework/utils.go
```diff
@@ -84,3 +88,32 @@ func RegisterScheme(s *runtime.Scheme) error {
...
+func WriteToFile(dir, fileName, content string) error {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		// create a temporary folder if the given folder doesn't exist
+		dir = os.TempDir()
+		if _, err = os.Stat(os.TempDir()); os.IsNotExist(err) {
+			err = os.Mkdir(os.TempDir(), 0700)
+			if err != nil {
+				return err
+			}
+		}
```
aleskandro: panic here if `dir` is having length == 0 and fill ARTIFACT_DIR in the ci-test.sh script

lwan-wanglin: fixed

aleskandro (approved):

aleskandro:
/retitle [MULTIARCH-4309](https://issues.redhat.com//browse/MULTIARCH-4309): Add test for ITMS

aleskandro:
great work @lwan-wanglin !

aleskandro:

PR #198, opened 2024-08-06
WIP: MULTIARCH-4812: namespace exclusion based on ownerreference

Prashanth684:
For openshift namespaces, exclude accrodingly
  - check for any namespace having ownerReference of ClusterVersion of Network
  - exclude the namespace if it does. If it is a namespace starting with openshift* and does not have the "-operator" suffix, 
     add the "-operator" suffix and check if the parent operator is managed by CVO. If so ,exclude this operand namespace 
    as well.
  - exclude "openshift-route-controller-manager" and "openshift-ingress-canary"
  - include the "openshift-marketplace" and "openshift-operators" namespaces

### DIFF
--- a/controllers/podplacement/namespaces.go
+++ b/controllers/podplacement/namespaces.go
@@ -69,7 +69,7 @@ func (s *NamespaceCache) onAdd() func(interface{}) {
 	return func(obj interface{}) {
 		ns := obj.(*v1.Namespace)
 		if len(ns.OwnerReferences) != 0 && (ns.Name != "openshift-marketplace" && ns.Name != "openshift-operators") {
-			// Print the kind of the first owner reference if available
+			// Exclude namespaces managed by CVO and networking operators
 			if ns.OwnerReferences[0].Kind == "ClusterVersion" || ns.OwnerReferences[0].Kind == "Network" {
 				s.namespaceMap.Insert(ns.Name)
 				s.log.V(5).Info("Added namespace ", ns.Name, " to map")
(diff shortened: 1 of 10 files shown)


controllers/podplacement/namespaces.go
```diff
@@ -0,0 +1,101 @@
...
+		mgr:          mgr,
+		namespaceMap: sets.New[string](),
+	}
+}
+
+func (s *NamespaceCache) Start(ctx context.Context) (err error) {
+	s.log = log.FromContext(ctx, "handler", "NamespaceCache", "kind", "Namespace")
+	s.log.Info("Starting Namespace informer")
+	// Setup an event handler for the informer
+	s.namespaceMap.Insert("openshift-route-controller-manager", "openshift-ingress-canary")
```
Prashanth684, line 49: need to add vars for these constants and add them to utils

Prashanth684:
next steps is to write e2es for these

controllers/podplacement/namespaces.go
```diff
@@ -0,0 +1,101 @@
...
+	namespaceInformer, err := s.mgr.GetCache().GetInformer(ctx, &v1.Namespace{})
+	if err != nil {
+		s.log.Error(err, "Error getting the informer")
+		return err
+	}
+	_, err = namespaceInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    s.onAdd(),
+		DeleteFunc: s.onDelete(),
```
Prashanth684, line 57: i did not add an update handler..it seems rare that owner references would be updated unless we want to consider other labels

Prashanth684:
closed in favor of #204

PR #204, opened 2024-08-08
MULTIARCH-4812: exclude pods in namespaces that are pinned to the control plane

Prashanth684:
#198 is not a foolproof way to determine and exclude pods which belong to openshift infra.  Instead exclude pods which are scheduled on the control plane as they may be mission critical and part of the core infra. There is no harm in the other pods going through the tuning operator and the gating process.

### DIFF
--- a/controllers/podplacement/scheduling_gate_mutating_webhook.go
+++ b/controllers/podplacement/scheduling_gate_mutating_webhook.go
@@ -76,10 +76,11 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
 	}
 	log := ctrllog.FromContext(ctx).WithValues("namespace", pod.Namespace, "name", pod.Name)
 
-	// ignore the openshift-* namespace as those are infra components, and ignore the namespace where the operand is running too
-	if utils.Namespace() == pod.Namespace || strings.HasPrefix(pod.Namespace, "openshift-") ||
-		strings.HasPrefix(pod.Namespace, "hypershift-") || strings.HasPrefix(pod.Namespace, "kube-") ||
-		pod.Spec.NodeName != "" {
+	// ignore the kube-* and hypershift-* namespace as those are infra components, and ignore the namespace where the operand is running too
+	// Also ignore any pods which are deployed on control plane nodes
+	if utils.Namespace() == pod.Namespace || strings.HasPrefix(pod.Namespace, "hypershift-") ||
+		strings.HasPrefix(pod.Namespace, "kube-") || pod.Spec.NodeName != "" ||
+		pod.Spec.NodeSelector != nil && utils.HasControlPlaneNodeSelector(pod.Spec.NodeSelector) {
 		log.V(5).Info("Ignoring the pod")
 		return a.patchedPodResponse(pod, req)
 	}
--- a/pkg/e2e/operator/pod_placement_config_test.go
+++ b/pkg/e2e/operator/pod_placement_config_test.go
@@ -208,6 +208,52 @@ var _ = Describe("The Multiarch Tuning Operator", func() {
 			verifyPodLabels(ns, "app", "test", e2e.Present, schedulingGateLabel)
 		})
 	})
+	Context("The webhook should not gate pods with node selectors that pin them to the control plane", func() {
+		BeforeEach(func() {
+			err := client.Create(ctx, &v1beta1.ClusterPodPlacementConfig{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "cluster",
+				},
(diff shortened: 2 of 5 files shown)


controllers/podplacement/scheduling_gate_mutating_webhook.go
```diff
@@ -66,13 +66,19 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
 		return admission.Errored(http.StatusBadRequest, err)
 	}
 
-	// ignore the openshift-* namespace as those are infra components, and ignore the namespace where the operand is running too
-	if utils.Namespace() == pod.Namespace || strings.HasPrefix(pod.Namespace, "openshift-") ||
-		strings.HasPrefix(pod.Namespace, "hypershift-") || strings.HasPrefix(pod.Namespace, "kube-") ||
-		pod.Spec.NodeName != "" {
+	// ignore the kube-* and hypershift-* namespace as those are infra components, and ignore the namespace where the operand is running too
+	if utils.Namespace() == pod.Namespace || strings.HasPrefix(pod.Namespace, "hypershift-") ||
+		strings.HasPrefix(pod.Namespace, "kube-") || pod.Spec.NodeName != "" {
```
aleskandro:
```suggestion
		strings.HasPrefix(pod.Namespace, "kube-") || pod.Spec.NodeName != "" ||
		pod.Spec.NodeSelector != nil && utils.HasControlPlaneNodeSelector(pod.Spec.NodeSelector) {
```
what about merging 75-81 here? The Golang short circuit will take care of not executing `utils.HasControlPlaneNodeSelector(pod.Spec.NodeSelector)` when NodeSelector is nil.

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -208,6 +208,74 @@ var _ = Describe("The Multiarch Tuning Operator", func() {
...
+						MatchExpressions: []metav1.LabelSelectorRequirement{
+							{
+								Key:      "multiarch.openshift.io/exclude-pod-placement",
+								Operator: "DoesNotExist",
+							},
+						}}}})
+			Expect(err).NotTo(HaveOccurred())
+			Eventually(deploymentsAreRunning).Should(Succeed())
+		})
+		It("should not gate pods that have the node-role.kubernetes.io/master node selector", func() {
```
aleskandro: IIUC we can loop through DescribeTable in this case

```suggestion
		DescribeTable("should not gate pods to schedule in control plane nodes", func(selector string) {
```

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -208,6 +208,74 @@ var _ = Describe("The Multiarch Tuning Operator", func() {
...
+			var err error
+			ns := framework.NewEphemeralNamespace()
+			err = client.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer client.Delete(ctx, ns)
+			var nodeSelectors = map[string]string{"node-role.kubernetes.io/master": ""}
```
aleskandro:
```suggestion
			var nodeSelectors = map[string]string{selector: ""}
```

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -208,6 +208,74 @@ var _ = Describe("The Multiarch Tuning Operator", func() {
...
+				WithReplicas(utils.NewPtr(int32(1))).
+				WithName("test-deployment").
+				WithNamespace(ns.Name).
+				Build()
+			err = client.Create(ctx, &d)
+			Expect(err).NotTo(HaveOccurred())
+			//should exclude the namespace
+			verifyPodNodeAffinity(ns, "app", "test")
+			verifyPodLabels(ns, "app", "test", e2e.Absent, schedulingGateLabel)
+		})
```
aleskandro:
```suggestion
		}),
		Entry("node-role.kubernetes.io/control-plane", "node-role.kubernetes.io/control-plane"),
		Entry("node-role.kubernetes.io/master", "node-role.kubernetes.io/master")
```

I would also use a constant in `utils/const.go` for storing these labels

aleskandro (commented):
Hey Prashanth, just some minor comments, but it looks good.

controllers/podplacement/scheduling_gate_mutating_webhook.go
Prashanth684: yup can do - just didn't want to make that conditional longer :)

aleskandro:

/lgtm

aleskandro:

PR #205, opened 2024-08-09
MULTIARCH-4933: Do not disable operands deletion

aleskandro:
This will change the UI uninstallation modal to be like the following:

image

We will hopefully have `https://issues.redhat.com/browse/RFE-6144` done and be able to force the deletion of the operands when users uninstall the operator.


aleskandro:

PR #207, opened 2024-08-09
MULTIARCH-4934: ordered removal of the operand and minor fixes 

aleskandro:
When the ClusterPodPlacementConfig is removed, the Pod placement controller should wait for the webhook to be deleted and should ungate all the pods.

This pr will let the operator add a finalizer to the ClusterPodPlacementConfig and the pod placement controller deployment such that when the user deletes the ClusterPodPlacementConfig, the deletion is marked for all the objects, with the webhook immediately collected to avoid new pods from being gated. The pod placement controller will continue running until all the pods are ungated. Then, the operator removes the finalizer in the deployment and the ClusterPodPlacementConfig, allowing the operand to be disabled without leaking any gated pod in the cluster.

This PR also includes changes in the logging system to synchronize the logging level set in the operator with the one requested by the ClusterPodPlacementConfig.
We reconcile the operands when a change to the ClusterPodPlacementConfig is detected. That reconciliation also included the setup of the log-level with the default zap command line flag.
However, the operator didn't change the level of logging.
I'm changing this behavior by leveraging a custom command line flag and the zap.AtomicLevel object that can be changed at runtime.

When a `CusterPodPlacementConfig` is reconciled, the controllers continue to be reconciled as they have always been, but the operator can now change the log level at runtime.
Moreover, the dev mode is now disabled so that the logs output uses the default non-dev JSON encoder, which allows the collection of logs through structured log collectors.

In the future, we may completely deprecate that command line flag and let the ClusterPodPlacementConfig drive the log level and other settings through informers in the operands' components.

### DIFF
--- a/controllers/operator/clusterpodplacementconfig_controller.go
+++ b/controllers/operator/clusterpodplacementconfig_controller.go
@@ -63,7 +63,10 @@ const (
 	serviceAccountName                       = "multiarch-tuning-operator-controller-manager"
 )
 
-const waitingForUngatingPodsError = "waiting for pods with the scheduling gate to be ungated"
+const (
+	waitingForUngatingPodsError         = "waiting for pods with the scheduling gate to be ungated"
+	waitingForWebhookSInterruptionError = "re-queueing to ensure the webhook objects deletion interrupt pods gating before checking the pods gating status"
+)
 
 //+kubebuilder:rbac:groups=multiarch.openshift.io,resources=clusterpodplacementconfigs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=multiarch.openshift.io,resources=clusterpodplacementconfigs/status,verbs=get;update;patch
@@ -114,12 +117,8 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 	log := ctrllog.FromContext(ctx).WithValues("operation", "handleDelete")
 	objsToDelete := []utils.ToDeleteRef{
 		{
-			NamespacedTypedClient: r.ClientSet.AppsV1().Deployments(utils.Namespace()),
-			ObjName:               PodPlacementControllerName,
-		},
-		{
-			NamespacedTypedClient: r.ClientSet.AppsV1().Deployments(utils.Namespace()),
-			ObjName:               PodPlacementWebhookName,
+			NamespacedTypedClient: r.ClientSet.AdmissionregistrationV1().MutatingWebhookConfigurations(),
+			ObjName:               podMutatingWebhookConfigurationName,
 		},
 		{
 			NamespacedTypedClient: r.ClientSet.CoreV1().Services(utils.Namespace()),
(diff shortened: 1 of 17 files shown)


controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -89,19 +97,28 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
...
+		log.V(4).Info("ClusterPodPlacementConfig not found")
+		// If the object is not found, or the finalizer has already been removed successfully, we can return here.
+		return ctrl.Result{}, nil
+	}
+	if !clusterPodPlacementConfig.DeletionTimestamp.IsZero() &&
+		!controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.PodPlacementFinalizerName) {
+		log.V(4).Info("the ClusterPodPlacementConfig object is being deleted, and the finalizer has already been removed successfully.")
+		return ctrl.Result{}, nil
+	}
+	if !clusterPodPlacementConfig.DeletionTimestamp.IsZero() {
```
Prashanth684: nit : consolidate into one for loop

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -128,18 +145,87 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context)
...
+	log.Info("Deleting the PodPlacement operand's resources")
+	err := utils.DeleteResources(ctx, objsToDelete)
+	// NOTE: err aggregates non-nil errors, excluding NotFound errors
+	if err != nil {
+		log.Error(err, "Unable to delete resources")
+		return err
+	}
+	log.Info("Looking for pods with the scheduling gate")
+	// get pending pods as we cannot query for the scheduling gate
+	pods, err := r.ClientSet.CoreV1().Pods("").List(ctx, metav1.ListOptions{
```
Prashanth684, line 161: isn't the phase SchedulingGated ?  that;s what we see through the cli right

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -128,18 +145,87 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context)
...
+					found = true
+				}
+			}
+		}
+		if found {
+			return errors.New(waitingForUngatingPodsError)
+		}
+	}
+
+	// The pods have been ungated and no other errors occurred, so we can remove the finalizer
```
Prashanth684, line 185: before this point i presume the webhook has already stopped running

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: nope phase is Pending, SchedulingGated is in the conditions, but we cannot query them via fieldSelector.

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: do you mean a single if?

I wanted to avoid nesting, but yeah i also wanted something clearer here.. I'll try something.

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: at least the service and the MutatingWebHookConfiguration should be removed (hence no communication will happen anymore with any replicas of the terminating webhook)... But I can re-queue right before line 155 by returning an error and enforcing a second cycle before querying the list of pods.

aleskandro: See https://github.com/openshift/multiarch-tuning-operator/pull/207/commits/788eacbd5fcd187a8b7cb7d10222bfb4ec16354e

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: see https://github.com/openshift/multiarch-tuning-operator/pull/207/commits/7303cc12476bc8049c03ce9624297a40b004ff4b, not sure if this looks better to you. I can consolidate the first two cases into one and back to if, but i also wanted to have the different logging message (for the more verbose logging setup).

Prashanth684: yeah this looks fine.

controllers/operator/clusterpodplacementconfig_controller.go
Prashanth684: yes this is reasonable.

aleskandro:

aleskandro:

PR #208, opened 2024-08-09
MULTIARCH-4936: Send events to the Pods to better inform about the actions of the scheduling gate

aleskandro:
Not all the users can be allowed to look at the logs of the pod placement controller or the web hook. Having information about the pods being scheduling gated and the result of the images inspection through events associated to them (also visible in K8S UIs) will improve the UX for users of this operator.


controllers/podplacement/scheduling_gate_mutating_webhook.go
```diff
@@ -96,5 +109,44 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
...
+		log := ctrllog.FromContext(ctx).WithValues("namespace", pod.Namespace, "name", pod.Name,
+			"function", "delayedSchedulingGatedEvent")
+		// We try to get the pod from the API with exponential backoff until we find it or a timeout is reached
+		err := wait.ExponentialBackoff(wait.Backoff{
+			Duration: 5 * time.Millisecond,
+			Factor:   2,
+			Steps:    100,
+		}, func() (bool, error) {
+			createdPod, err := a.ClientSet.CoreV1().Pods(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
+			if err == nil {
```
Prashanth684, line 141: hmm..this is going to potentially flood the apiserver for each pod...we should try to qualify this to see how it will perform under load.

aleskandro: Yes, each single go-routine will have an ExponentialBackoff-based limiter... But they share the same clientSet, that is limited with a Max number of QPS and Burst requests: https://github.com/openshift/multiarch-tuning-operator/blob/3383958ba5b56647d999ac75da1ed43d392cc9bd/vendor/sigs.k8s.io/controller-runtime/pkg/client/config/config.go#L94-L109

I'll reduce the number of maximum steps for the backoff here in any case.

aleskandro: also see https://github.com/openshift/multiarch-tuning-operator/pull/208/commits/8079b111bb1e5d2a3756552124fa7f5dc1c2e83c

Prashanth684: looks good.

aleskandro:

/approve

PR #214, opened 2024-08-12
MULTIARCH-4937: Implement Status of the ClusterPodPlacementConfig, conditions and improve reliability.

aleskandro:
This commit implements the handling of the following conditions for the ClusterPodPlacementConfig:
   - Degraded: if some components are not available (no replicas) and the object is not deprovisioning
   - Deprovisioning: if the object is being deleted
   - MutatingWebhookConfigurationNotAvailable: if the mutating webhook configuration does not exist
   - PodPlacementControllerNotReady: if the pod placement controller is not available or up-to-date
   - PodPlacementWebhookNotReady: if the pod placement webhook is not available or up-to-date
   - Progressing: if the object is not deprovisioning and some of the components are not up-to-date.
   - Available: if all the components are available to serve the requests and reconcile node affinities (at least one replica).

This allows the users to be informed about the status of the operand.
When the operand is available and all the components are up-to-date, the only condition shown/set is "Available."

When the ClusterPodPlacementConfig is updated, or some components are being rolled out, the "Progressing" condition is set.

When progressing, if at least one replica is available for both the controller and webhook, the operand is considered available.
In this case, we also allow the deployment of the mutating webhook configuration.

Despite the current FailurePolicy of the MutatingWebhookConfiguration is set to None, if some components lack at least one replica's availability, we forcibly delete the MutatingWebhookConfiguration, and the Degraded condition is set. Similarly, we now deploy the MutatingWebhookConfiguration only if at least one replica of the MutatingWebhook and Controller deployments are available (such that we can ensure both the admission patching and the removal of gates through the controller).

When removing the object, the deprovisioning condition is set and the users are informed about the controller pending deletion if there exists at least one pod that is still gated.

This depends on #207

Only consider the last commit for review.

Evolution of the object (creation + pod pending with scheduling gate + deprovisioning): https://gist.github.com/aleskandro/1297d346f6749fa6c0a0fa5db69d764f

Also added columns to print out when using CLI:
```
└ $ oc get clusterpodplacementconfigs.multiarch.openshift.io
NAME      AVAILABLE   PROGRESSING   DEGRADED   SINCE   STATUS
cluster   True        False         False      13h     AllComponentsReady
```

### DIFF
--- a/apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
+++ b/apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
@@ -17,8 +17,14 @@ limitations under the License.
 package v1beta1
 
 import (
-	"github.com/openshift/multiarch-tuning-operator/apis/multiarch/common"
+	"fmt"
+	"strings"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/library-go/pkg/operator/v1helpers"
+
+	"github.com/openshift/multiarch-tuning-operator/apis/multiarch/common"
 )
 
 // ClusterPodPlacementConfigSpec defines the desired state of ClusterPodPlacementConfig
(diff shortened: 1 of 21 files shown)


apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
```diff
@@ -42,6 +48,144 @@ type ClusterPodPlacementConfigSpec struct {
 type ClusterPodPlacementConfigStatus struct {
 	// Conditions represents the latest available observations of a ClusterPodPlacementConfig's current state.
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// The following fields are used to derive the conditions. They are not exposed to the user.
+	available                                bool `json:"-"`
+	progressing                              bool `json:"-"`
+	degraded                                 bool `json:"-"`
+	deprovisioning                           bool `json:"-"`
```
Prashanth684, line 56: does deprovisioning cover the case where we are waiting for pods to be ungated or do we need a new "draining" status

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
```diff
@@ -42,6 +48,144 @@ type ClusterPodPlacementConfigSpec struct {
...
+	// tracks the availability of the pod placement controller and webhook and if they are up to date
+	s.podPlacementControllerNotReady = !podPlacementControllerAvailable || !podPlacementControllerUpToDate
+	s.podPlacementWebhookNotReady = !podPlacementWebhookAvailable || !podPlacementWebhookUpToDate
+	// if all the components exist and have at least one replica ready
+	s.available = mutatingWebhookConfigurationAvailable && podPlacementWebhookAvailable && podPlacementControllerAvailable
+	// if some components are not available (no replicas)
+	s.degraded = !s.available && !s.deprovisioning // degraded will not track deprovisioning
+	// allow the deployment of the mutating webhook configuration if the pod placement controller and webhook are available
+	// (at least one replica)
+	s.canDeployMutatingWebhook = podPlacementWebhookAvailable && podPlacementControllerAvailable && !s.deprovisioning
```
Prashanth684, line 121: do we need `canDeployMutatingWebhook` given that we already report status if the configuration is not available?

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
```diff
@@ -42,6 +48,144 @@ type ClusterPodPlacementConfigSpec struct {
...
+	// tracks existence of the mutating webhook configuration
+	s.mutatingWebhookConfigurationNotAvailable = !mutatingWebhookConfigurationAvailable
+	// tracks the availability of the pod placement controller and webhook and if they are up to date
+	s.podPlacementControllerNotReady = !podPlacementControllerAvailable || !podPlacementControllerUpToDate
+	s.podPlacementWebhookNotReady = !podPlacementWebhookAvailable || !podPlacementWebhookUpToDate
+	// if all the components exist and have at least one replica ready
+	s.available = mutatingWebhookConfigurationAvailable && podPlacementWebhookAvailable && podPlacementControllerAvailable
+	// if some components are not available (no replicas)
+	s.degraded = !s.available && !s.deprovisioning // degraded will not track deprovisioning
+	// allow the deployment of the mutating webhook configuration if the pod placement controller and webhook are available
```
Prashanth684, line 119: `&& !s.progressing` ?

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
aleskandro: yes, it does. It starts to be deprovisioning when the deletionTimestamp is set. Then, the cleanup occur and, once the cleanup is done the finalizers are removed. No other conditions as the last finalizer removed will let the object to disappear. 

The object "dies" as in the last version in the linked dump.

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
aleskandro: this is to inform internally about the fact it is time we can actually deploy the mutatingWebhookConfiguration (letting the APIserver send requests to us for any newly created pod). 

 mutatingWebhookConfiguration can be unavailable and this value can still be false because we miss at least 1 replica in either the controller or the webhook deployments (that would potentially mean patching the pods without having a controller that can remove the gate).

I'm adding this additional gate to gate the creation of the configuration before it's really possible to consider.

We don't get failures due to this now because the failurePolicy is set to None, but I wanted to add it as a good mechanism.

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
aleskandro: it should be "included" based on how progressing is currently computed. Degraded will show up if there's some component with no replicas available at all or the webhookConfiguration is missing.

I'm not sure how to distinguish the specific case where we are actually "progressing" or not.

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
aleskandro: See https://github.com/openshift/multiarch-tuning-operator/pull/214/files/820592071aeb62de6ccd76048f95ef958fe31359#diff-2ccc8a1f8d7e3da0a4eb974bab7660f7a801c392bc16d1114181815f1efe290aR283-R291

Prashanth684: > We don't get failures due to this now because the failurePolicy is set to None

ah..right indeed..i forgot that we have failurepolicy set to none. makes sense

apis/multiarch/v1beta1/clusterpodplacementconfig_types.go
Prashanth684: if it is not deprovisioning, shouldn't it be considered progressing always?

Prashanth684: hmm..but then how do we compute degraded? do we wait for a few times for the individual pod deployments to fail ?

aleskandro: yes, as far as the deployments are not up-to-date, it's progressing.

If some deployment got 0 replicas, it's degraded (and we remove the mutatingWebhookConfiguration).

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -97,8 +98,12 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
 		log.Error(err, "Unable to fetch ClusterPodPlacementConfig")
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
-
 	log.V(3).Info("ClusterPodPlacementConfig fetched...", "name", clusterPodPlacementConfig.Name)
+	err = r.dependentsStatusToClusterPodPlacementConfig(ctx, clusterPodPlacementConfig)
+	if err != nil {
+		log.Error(err, "Unable to gather the status of the dependents")
```
AnnaZivkovic: Would something like this be clearer? 
"Unable to retrieve the status of the PodPlacementConfig dependencies"

aleskandro:
/retest-required

The issue with the e2e tests is due to some race conditions happening now that we have a more complex FSM based on the conditions. 

Some of the fixees may be included in #222 already. Launching some more tests. If #222 solves we may either merge both or incorporate this into it.

PR #222, opened 2024-08-16
OCPBUGS-38601: The operand should not fail when the operator is uninstalled without prior deletion of the operator

aleskandro:
When the operator is uninstalled without prior deletion of the operand, we should not let the operand fail. This is due to the removal of the service account and other RBAC objects. This commit will let the operator create the RBAC objects and additional service accounts detached from the OLM ones.
The first commit also includes an edge-case fix: when a deployment is deleted outside of the operand's lifecycle, the operator must be able to reconcile it, so we need to remove the finalizer.

This PR also adds TopologySpreadConstraints to handle scenarios of nodes maintenance.

~Depends on https://github.com/openshift/multiarch-tuning-operator/pull/214~

### DIFF
--- a/apis/multiarch/v1beta1/groupversion_info.go
+++ b/apis/multiarch/v1beta1/groupversion_info.go
@@ -34,3 +34,5 @@ var (
 	// AddToScheme adds the types in this group-version to the given scheme.
 	AddToScheme = SchemeBuilder.AddToScheme
 )
+
+const ClusterPodPlacementConfigResource = "clusterpodplacementconfigs"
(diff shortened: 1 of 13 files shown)


aleskandro:
/cc @Prashanth684 @AnnaZivkovic 
(about https://github.com/openshift/multiarch-tuning-operator/pull/222/commits/cc8f805fe3d07b506ea9c6d236c9f76d88ca65be)

/cc @lwan-wanglin 
(for testing and in particular about the changes in https://github.com/openshift/multiarch-tuning-operator/pull/222/commits/b93b63ee8d9665492f26e005a2474a71965c10ce)

aleskandro:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_multiarch-tuning-operator/222/pull-ci-openshift-multiarch-tuning-operator-main-e2e-gcp-multi-operator-olm/1826084713811415040 passed. The new commit is just a fix on a nit.

PR #228, opened 2024-08-20
Mount the nodes' /etc/containers/ and /etc/docker/

AnnaZivkovic:
Mount the nodes' /etc/containers/ and /etc/docker/ directories to the PodPlacementConfig. 

This setup enables direct access to the registry configuration files, certificates, and policies, eliminating the need for temporary storage in the /tmp/ directory. This also reduces the computational load associated with maintaining the registries

### DIFF
--- a/controllers/operator/objects.go
+++ b/controllers/operator/objects.go
@@ -103,6 +103,52 @@ func buildService(name string, controllerName string, port int32, targetPort int
 	}
 }
 
+func buildWebhookDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlacementConfig) *appsv1.Deployment {
+	return buildDeployment(clusterPodPlacementConfig, utils.PodPlacementWebhookName, 3, utils.PodPlacementWebhookName, "",
+		"--enable-ppc-webhook",
+	)
+
+}
+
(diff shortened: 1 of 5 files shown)


controllers/operator/objects.go
```diff
@@ -87,6 +87,52 @@ func buildService(name string, controllerName string, port int32, targetPort int
 	}
 }
 
+func buildWebhookDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlacementConfig) *appsv1.Deployment {
+	return buildDeployment(clusterPodPlacementConfig, utils.PodPlacementWebhookName, 3, utils.PodPlacementWebhookName, "",
```
aleskandro, line 107:
```suggestion
	return buildDeployment(clusterPodPlacementConfig, utils.PodPlacementWebhookName, 3, "", utils.PodPlacementWebhookName,
```

lwan-wanglin:
the pre-merge testing passed with below scenrio:

- registries in a mirror list and registries is added as insecure, the mirror will be added as insecure
- registries mirror set with NeverContactSource issue also fixed
- if a registry is added as insecure, the sub-repository will consider as insecure

PR #240, opened 2024-08-26
NO-JIRA: Add securityContext to kube-rbac-proxy containers

aleskandro:
#228 introduced using a hostPath volume to feed the content of the `/etc/containers` folder. The pod security admission controller does not allow this. 
The minimum level for the pod security admission that allows using an host path volume is privileged. See https://kubernetes.io/docs/concepts/security/pod-security-standards/

The OLM test for #228 succeeded because the `operator-sdk run bundle` command changed the namespace labels.

This PR adds the security context for the deployments' kube-rbac-proxy container and a method to reconcile the namespace labels in the cluster pod placement config controller to set the privileged pod admission security policy.

Finally, as we couldn't do that before when the operator is installed via the OperatorHub, I'm exploiting this fix requiring updates permissions to the namespace for adding the 
`"workload.openshift.io/allowed": "management"` label too (see https://github.com/openshift/enhancements/blob/c5b9aea25e/enhancements/workload-partitioning/management-workload-partitioning.md)

```
  Warning  FailedCreate  25s (x6 over 105s)  replicaset-controller  (combined from similar events): Error creating: pods "pod-placement-controller-6b448c5d6-pp6nt" is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false (container "kube-rbac-proxy" must set securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container "kube-rbac-proxy" must set securityContext.capabilities.drop=["ALL"]), restricted volume types (volumes "docker-conf", "containers-conf" use restricted volume type "hostPath"), runAsNonRoot != true (pod or container "kube-rbac-proxy" must set securityContext.runAsNonRoot=true)
```

```
  Warning  FailedCreate  1s (x4 over 19s)  replicaset-controller  (combined from similar events): Error creating: pods "pod-placement-controller-7cb6978477-fsccr" is forbidden: violates PodSecurity "restricted:latest": restricted volume types (volumes "docker-conf", "containers-conf" use restricted volume type "hostPath")
```

It worked on the OLM version as `operator-sdk run bundle` . However, this should not happen when we deploy from the OperatorHub.

namespace patched by operator-sdk run bundle:

```
        {
            "apiVersion": "v1",
            "kind": "Namespace",
            "metadata": {
                "annotations": {
                    "openshift.io/sa.scc.mcs": "s0:c26,c25",
                    "openshift.io/sa.scc.supplemental-groups": "1000700000/10000",
                    "openshift.io/sa.scc.uid-range": "1000700000/10000",
                    "workload.openshift.io/allowed": "management"
                },
                "creationTimestamp": "2024-08-26T18:40:05Z",
                "labels": {
                    "kubernetes.io/metadata.name": "openshift-multiarch-tuning-operator",
                    "pod-security.kubernetes.io/audit": "privileged",
                    "pod-security.kubernetes.io/audit-version": "v1.24",
                    "pod-security.kubernetes.io/enforce": "privileged",
                    "pod-security.kubernetes.io/enforce-version": "v1.24",
                    "pod-security.kubernetes.io/warn": "privileged",
                    "pod-security.kubernetes.io/warn-version": "v1.24",
                    "security.openshift.io/scc.podSecurityLabelSync": "true"
                },
                "name": "openshift-multiarch-tuning-operator",
                "resourceVersion": "32163",
                "uid": "65ac6b40-7515-4c7a-8869-fc22cf34eed3"
            },
            "spec": {
                "finalizers": [
                    "kubernetes"
                ]
            },
            "status": {
                "phase": "Active"
            }
        },
```

### DIFF
--- a/controllers/operator/clusterpodplacementconfig_controller.go
+++ b/controllers/operator/clusterpodplacementconfig_controller.go
@@ -68,6 +68,7 @@ const (
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations,verbs=get;update;patch;create;delete;list;watch
 //+kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations/status,verbs=get
 
+//+kubebuilder:rbac:groups=core,resources=namespaces,verbs=get;update
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;update;patch;create;delete
 //+kubebuilder:rbac:groups=apps,resources=deployments/status,verbs=get
 //+kubebuilder:rbac:groups=apps,resources=deployments/finalizers,verbs=update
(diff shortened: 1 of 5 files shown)


controllers/operator/objects.go
```diff
@@ -310,6 +310,19 @@ func buildDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlacementConfi
...
+									Drop: []corev1.Capability{
+										"ALL",
+									},
+								},
+								Privileged:   utils.NewPtr(false),
+								RunAsNonRoot: utils.NewPtr(true),
+								SeccompProfile: &corev1.SeccompProfile{
+									Type: corev1.SeccompProfileTypeRuntimeDefault,
+								},
+							},
```
AnnaZivkovic, line 325: Why does the webhook also need the SecurityContext?

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -124,6 +125,34 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
...
+	if err != nil {
+		log.Error(err, "Unable to get the namespace")
+		return err
+	}
+	if ns.Labels == nil {
+		ns.Labels = make(map[string]string)
+	}
+	ns.Labels["pod-security.kubernetes.io/audit"] = "privileged"
+	ns.Labels["pod-security.kubernetes.io/audit-version"] = "v1.29"
+	ns.Labels["pod-security.kubernetes.io/enforce"] = "privileged"
```
Prashanth684, line 145: do we need to do a check for each of these labels in case they are deleted individually (manually by the user)

controllers/operator/objects.go
aleskandro: In general, it's better to set it for both the containers. This security context is not increasing any privilege. It's rather limiting the pod. A previous PR introduced it in the first container only.

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: they are reconciled everytime the ClusterPodPlacementConfig progresses, until it's Available and all components are ready. So, if a user deletes them manually and some other events occur for which the pods need to be recreated, we reconcile the labels too.

aleskandro:

PR #246, opened 2024-08-28
OCPBUGS-36244: Cluster wide proxy

AnnaZivkovic:
To ensure the operator works with a cluster-wide proxy, the Pod placement needs the proxy environment variables propagated to it. If a customer adds a custom CA certificate, the Pod placement configuration needs access to it. This is achieved by adding an empty ConfigMap with the label config.openshift.io/inject-trusted-cabundle: "true", which will inject a bundle of CA certificates.
[](https://docs.okd.io/4.16/operators/admin/olm-configuring-proxy-support.html)

### DIFF
--- a/controllers/operator/objects.go
+++ b/controllers/operator/objects.go
@@ -233,6 +233,7 @@ func buildDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlacementConfi
 									Name:  "HTTPS_PROXY",
 									Value: os.Getenv("HTTPS_PROXY"),
 								},
+
 								{
 									Name:  "NO_PROXY",
 									Value: os.Getenv("NO_PROXY"),
@@ -303,6 +304,11 @@ func buildDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlacementConfi
 								{
 									Name:      "ca-projected-volume",
 									MountPath: "/etc/ssl/certs",
+									ReadOnly:  true,
+								},
+								{
+									Name:      "trusted-ca",
+									MountPath: "/etc/pki/ca-trust/extracted/pem",
 									ReadOnly:  true,
 								},
 							},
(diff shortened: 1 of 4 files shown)


controllers/operator/objects.go
```diff
@@ -394,6 +413,28 @@ func buildDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlacementConfi
 								},
 							},
 						},
+						{
+							Name: "trusted-ca",
+							VolumeSource: corev1.VolumeSource{
+								Projected: &corev1.ProjectedVolumeSource{
```
aleskandro: We may be able to get rid of the Projected abstraction in this case as it's only one ConfigMap to consider

aleskandro: Example from https://kubernetes.io/docs/concepts/configuration/configmap/#configmaps-and-pods

```yaml
      volumeMounts:
      - name: config
        mountPath: "/config"
        readOnly: true
  volumes:
  # You set volumes at the Pod level, then mount them into containers inside that Pod
  - name: config
    configMap:
      # Provide the name of the ConfigMap you want to mount.
      name: game-demo
      # An array of keys from the ConfigMap to create as files
      items:
      - key: "game.properties"
        path: "game.properties"
      - key: "user-interface.properties"
        path: "user-interface.properties"
```

controllers/operator/objects.go
```diff
@@ -3,6 +3,8 @@ package operator
 import (
 	"fmt"
 
+	"os"
+
```
aleskandro:
```suggestion
	"fmt"
	"os"

```

in general we are keeping this order:
1. golang packages
2. k8s
3. sig.k8s
4. openshift
5. openshift/multiarch-tuning-operator

aleskandro: or try to :). The linter does only a part of it

PR #255, opened 2024-09-04
MULTIARCH-4877: Advertise the operator as supporting disconnected sce…

AnnaZivkovic:
…narios in the bundle's annotations


aleskandro:

PR #256, opened 2024-09-05
OCPBUGS-38823: skip bundle images when computing the set of compatible architectures for a given pod

aleskandro:
This PR updates the registry inspector to handle bundle images correctly. Bundle images are intended to carry metadata-only information and not be multi-architecture.

It is possible for an operator bundle image to reference a valid operator supported by the cluster's architectures while the bundle image itself remains a single-architecture manifest image. Suppose the cluster (e.g., an arm64+ppc64le cluster) has no nodes supporting the bundle image's architecture (such as an amd64-built bundle). OLM may fail to run the job in that case because the node affinity settings could lead to an ImagePullBackOff error.

To address this, the GetCompatibleArchitectureSet function will now return the set of all the architectures supported by the operator (arm64, amd64, ppc64le, s390x) for OLM bundle images. If there are other container images specified in the pod (as in the OLM case), the node affinity will be set to the intersection of the architectures supported by the non-bundle images. As this fix is strictly related to the OLM bundles, we don't consider the case where a pod has only that bundle as a container image. In fact, the pod cannot run in such a case because no binaries are shipped in the image of a bundle to act as an entry point.

Testing: I added two integration tests for this. Having some e2e might be good (e.g., subscription-based operator installation, operator-sdk run, but I'd leave this choice to @lwan-wanglin and consider it a low-priority task for now as the scenarios we support consider mixarch amd64 + $second_arch today and this edge case would not be faced.

A manual test, though, could be done even before merging.

### DIFF
--- a/pkg/image/inspector.go
+++ b/pkg/image/inspector.go
@@ -31,11 +31,17 @@ import (
 	"github.com/containers/image/v5/manifest"
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/types"
+	"github.com/opencontainers/go-digest"
 
 	"github.com/go-logr/logr"
 	"golang.org/x/sys/unix"
 
 	"github.com/openshift/multiarch-tuning-operator/pkg/systemconfig"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
+)
+
+const (
+	operatorSDKBuilderBundleAnnotation = "operators.operatorframework.io.metrics.builder"
 )
 
 type registryInspector struct {
@@ -44,6 +50,12 @@ type registryInspector struct {
 	mutex sync.RWMutex
 }
 
+// GetCompatibleArchitecturesSet returns the set of compatibles architectures given an imageReference and a list of secrets.
+// It uses the containers/image library to get the manifest of the image and extract the architecture from it.
+// If the image is a manifest list, it will return the set of architectures supported by the manifest list.
+// If the image is a manifest, it will return the architecture set in the manifest's config.
+// If the image is an operator bundle image, it will return an empty set. This is because operator bundle images
+// are not tied to a specific architecture, and we should not set any constraints based on the architecture they report.
 func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, imageReference string, secrets [][]byte) (supportedArchitectures sets.Set[string], err error) {
 	// Create the auth file
 	log := ctrllog.FromContext(ctx, "imageReference", imageReference)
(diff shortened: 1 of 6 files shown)


aleskandro:

aleskandro:

controllers/podplacement/pod_model.go
```diff
@@ -227,6 +227,12 @@ func (pod *Pod) intersectImagesArchitecture(pullSecretDataList [][]byte) (suppor
 			log.V(3).Error(err, "Error inspecting the image", "imageName", imageName)
 			return nil, err
 		}
+		if err == nil && len(currentImageSupportedArchitectures) == 0 {
+			// This case happens when the image is an operator bundle image. We don't set any constraints based on the architecture they report.
+			// See https://issues.redhat.com/browse/OCPBUGS-38823 for more information.
+			log.V(4).Info("No supported architectures found for the image (This image should be a bundle).", "imageName", imageName)
```
prb112: per our discussion, this looks like a reasonable solution. wouldn't it be conversely reasonable to set all architectures? since it should work everywhere?

just a thought.

aleskandro: Hi Paul, I did this because "all architectures" is not really defined outside OCP.

But by doing that, we may remove this `if` statement that i didn't like either. 
I'm going to change it. Thanks for pointing this out :)

aleskandro: Hi Paul, I did this because "all architectures" is not really defined outside OCP.

But by doing that, we may remove this `if` statement that i didn't like either. 
I'm going to change it. Thanks for pointing this out :)

aleskandro:

pkg/image/inspector.go
```diff
@@ -115,33 +123,53 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
+
 	config, err := parsedImage.OCIConfig(ctx)
+
 	if err != nil {
-		// Ignore errors due to invalid images at this stage
 		log.Error(err, "Error parsing the OCI config of the image")
 		return nil, err
 	}
-	return sets.New[string](config.Architecture), nil
+	if _, ok := config.Config.Labels["operators.operatorframework.io.metrics.builder"]; ok {
```
Prashanth684: have this as a const with a name similar to https://github.com/operator-framework/operator-sdk/blob/master/internal/annotations/metrics/metrics.go#L32 ?

aleskandro: Done

aleskandro:

prb112 (commented):
/lgtm

PR #257, opened 2024-09-05
MULTIARCH-4294: Implement prometheus metrics for the pod placement controller

aleskandro:
This PR implements the Prometheus metrics for the Pod placement controller.

As the kube-rbac-proxy is a main component to serve metrics and it is being deprecated, this PR also addresses MULTIARCH-4989.

See https://github.com/kubernetes-sigs/kubebuilder/issues/3871 for further details.

Metrics are documented in the [docs/metrics.md](https://github.com/openshift/multiarch-tuning-operator/pull/257/files#diff-1ad57e035683c4ad698d4b2b6840add43236bfaefbe4b705f0a81cf28959c4b9) file and an example Grafana Dashboard is provided [here](https://github.com/openshift/multiarch-tuning-operator/pull/257/files#diff-f132fa8d8c92404894273333e00bec4bc8359193e202ef70b140b7426aee49f3).

Finally, this PR implements some performance optimizations:
- Concurrent reconciles to improve the processed pods rate
- Caching and reconciling only Pending pods
- Resync period and (concurrent-safe) update of the global pull-secret only when it changes


controllers/podplacement/metrics/common_metrics.go
```diff
@@ -0,0 +1,33 @@
...
+	"sync"
+
+	metrics2 "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+func buckets() []float64 {
+	return []float64{
+		0.001, 0.002, 0.005, 0.010, 0.020, 0.050,
```
Prashanth684: what are the reasoning to have several of these buckets? are we targeting a certain SLO that we can only define buckets for?

docs/metrics.md
```diff
@@ -0,0 +1,75 @@
...
+
+-- Current number of gated pods (with the multiarch tuning operator scheduling gate)
+sum(mto_ppo_pods_gated)
+-- Current number of gated pods (with any scheduling gate)
+sum(scheduler_pending_pods{queue="gated"})
+sum(scheduler_pending_pods) by (queue)
+
+-- Distribution of the time to inspect an image
+sum by(le) (rate(mto_ppo_ctrl_time_to_inspect_pod_images_seconds_bucket[5m]))
+
```
Prashanth684, line 74: would also add a sample query for users to set up alerts based on hitting SLOs. for example:
```
sum(rate(mto_ppo_ctrl_time_to_process_gated_pod_seconds_bucket{le="0.3"}[5m])) by (job)
/sum(rate(mto_ppo_wh_pods_gated_total[5m])) by (job) >=0.90
```
for a 90% SLO for a targeted duration of 300ms.

docs/metrics.md
```diff
@@ -0,0 +1,75 @@
...
+| `mto_ppo_ctrl_processed_pods_total`               | Counter   | pod placement controller | The total number of pods processed by the pod placement controller that had a scheduling gate                   |
+| `mto_ppo_ctrl_failed_image_inspection_total`      | Counter   | pod placement controller | The total number of image inspections that failed.                                                              |
+| `mto_ppo_pods_gated`                              | Gauge     | controller and webhook   | The current number of gated pods (this metric is not considered reliable yet). It should converge to 0.         |
+| `mto_ppo_wh_pods_processed_total`                 | Counter   | mutating webhook         | The total number of pods processed by the webhook.                                                              |
+| `mto_ppo_wh_pods_gated_total`                     | Counter   | mutating webhook         | The total number of pods gated by the webhook.                                                                  |
```
Prashanth684, line 28: do we also need a metric for the pods "bypassed" by the controller after being gated by the webhook ?

docs/metrics.md
aleskandro: Hi @Prashanth684 , I would ignore AlertRule and suggestions of alerts in this PR. I was starting to look at them, but I would collect more data and have the inspection caching implemented first. Otherwise, we risk false positives alerts that pop up for the users easily.

docs/metrics.md
aleskandro: When we have #242 in place, I'd be happier to consider any pod ignored by the controller and not ignored by the webhook (e.g., when another mutating webhook added affinities before us) as `mto_ppo_wh_pods_gated_total - mto_ppo_ctrl_processed_pods_total`

controllers/podplacement/metrics/common_metrics.go
aleskandro: I wanted to build them from a better (mathematical) function, and forget to look back at this.

As of now, the buckets are drawn as in
- 0.001-0.05 (linear): take requests that are below 100ms, fine-grained (requests that HIT the cache should be served in these buckets)
- 0.100-0.900 (linear): I expect requests to remote registries to land in these buckets
- 0.900+ (exponential-like): here, I expect some 'anomalous' requests, slow registries, busy networks, or clusters.

Initially, I was drawing the buckets from an exponential function, but it grows too fast once it reaches 0.100 (e.g., with 2x, 0.200 is the next, and then 0.400, 0.800...)... 

Having a few more buckets allows us to have more clues about the distribution of the requests. I may reduce the ones after 1 to [1.5, 2, 4, 8]. (I missed 0.8 between 0.7 and 0.9).

docs/metrics.md
aleskandro: or maybe we may need one more in that PR. Let me see how the metrics look like in #242 after rebasing when this lands if you're ok with it.

controllers/podplacement/metrics/common_metrics.go
aleskandro: Fixed the second 'tier' of buckets and reduced the third (1.5s+)

aleskandro:

PR #267, opened 2024-09-12
NO-JIRA: Enable cleanup in the CSV

aleskandro:
This is related to MULTIARCH-4301, and based on https://github.com/operator-framework/enhancements/blob/master/enhancements/olm-deletes-operands.md


aleskandro:
that was never implemented

PR #271, opened 2024-09-12
MULTIARCH-4965: Update snapshot multiarch-tuning-operator-nwwnk

AnnaZivkovic:
oc describe snapshots multiarch-tuning-operator-nwwnk
```
Spec:
  Application:  multiarch-tuning-operator
  Artifacts:
  Components:
    Container Image:  quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator/multiarch-tuning-operator@sha256:36705668e673ee0105cb7453416a46430b24432e0a5d2c7e09a6c174098da044
    Name:             multiarch-tuning-operator
    Source:
      Git:
        Context:         ./
        Dockerfile URL:  Dockerfile
        Revision:        d10efe3fddb1745931dfd9b8be8a3492ee12a7f8
        URL:             https://github.com/openshift/multiarch-tuning-operator
    Container Image:     quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator/multiarch-tuning-operator-bundle@sha256:440d0bbcf88a33f83ae9caeb7caad4c9234509e341d7385442b33063d31d8e27
    Name:                multiarch-tuning-operator-bundle
    Source:
      Git:
        Revision:  49ce213316883dd4f54ba5da17d70c78ef3aff67
        URL:       https://github.com/openshift/multiarch-tuning-operator
```

To upgrade the operator via web console follow the instructions [here](https://docs.openshift.com/container-platform/4.16/operators/admin/olm-upgrading-operators.html#olm-changing-update-channel_olm-upgrading-operators)  

To upgrade via `oc` update the channel field in the subscription to change the channel the operator uses for updates.

Find the needed subscription
```oc get subscriptions -n <namespace>```

Run the following command to edit the subscription:

```oc edit subscription <subscription-name> -n <namespace>```

Once in the editor, find the spec.channel field and change it to the desired channel. For example:
```
spec:
  channel: stable
```
Change stable to your desired channel, e.g., tech-preview, alpha, etc. If the update approval method is set to Manual then  the update needs to be manually approved before installation begins otherwise it will begin automatically.

### CODE REVIEW

File: catalog/multiarch-tuning-operator/index.yaml [hunk 1]
```diff
@@ -31,7 +31,7 @@ name: tech-preview
 entries:
   - name: multiarch-tuning-operator.v0.9.0
 ---
-image: registry.redhat.io/multiarch-tuning/multiarch-tuning-operator-bundle@sha256:b7c0e2ac16ea027d5af85e3ee10556e2f4cb5cbda956e432b63030ba688262cb
+image: registry.redhat.io/multiarch-tuning/multiarch-tuning-operator-bundle@sha256:440d0bbcf88a33f83ae9caeb7caad4c9234509e341d7385442b33063d31d8e27
```
aleskandro: - we should not change this reference, but add a new bundle and a new channel, `stable`.
- We should set the default channel to stable (first lines of this file).
- We should configure the upgrade edges based on the OLM doc about cross-channel uprades in this case (we will have the users moving from the `tech-preview` channel to `stable`).

You can use

```shell
podman run -it --entrypoint /bin/opm -v <path/to/code>:/code:Z registry.redhat.io/openshift4/ose-operator-registry:v4.13 render <konflux-bundle-image> --output=yaml > tmp/rendered.yaml
```

to render a new bundle and copy paste the result in the bottom of this file.

catalog/multiarch-tuning-operator/index.yaml
```diff
@@ -31,6 +31,12 @@ name: tech-preview
 entries:
   - name: multiarch-tuning-operator.v0.9.0
 ---
+schema: olm.channel
```
aleskandro, line 34: Do we need to use some keys like `replace: ` to allow the cross-channel upgrade?

PR #272, opened 2024-09-12
MULTIARCH-4296: Add more pre-checking logic in the webhook to skip pods that will not get a patched nodeAffinity

aleskandro:
Changes proposed in this PR:

- Improve the code of the webhook by using the Pod model embedding v1.Pod so that we can share further methods among the controller and the webhook
- Ensure the pod placement controller also runs the shouldIgnore verification: We remove the scheduling gate for pods that no longer meet the criteria for architecture-aware scheduling due to changes between the webhook patch at creation time and the pod's processing by the pod placement controller.
   - A pod may be gated but unprocessed when the operator's configuration changes, making the pod ineligible for architecture-aware scheduling based on the new rules.
   - Alternatively, the pod may have been modified by another webhook in the admission chain (e.g., setting the `nodeAffinity` for the `kubernetes.io/arch` label), rendering it unsuitable for further processing by the operator.
   In both scenarios, we must not set the nodeAffinity, and the gate should be removed.
- Ignore pods that have nodeSelectors/affinity set earlier in the webhook and before even catching pull secrets
- Finally, we stop ignoring hypershift—* namespaces. At this stage, it should mostly be a testing effort to validate that hyper shift works. Also, namespaces with hypershift—* are not a constraint for hypershift management clusters.

### DIFF
--- a/controllers/podplacement/pod_model.go
+++ b/controllers/podplacement/pod_model.go
@@ -346,6 +346,6 @@ func (pod *Pod) isNodeSelectorConfiguredForArchitecture() bool {
 func (pod *Pod) publishIgnorePod() {
 	log := ctrllog.FromContext(pod.ctx)
 	log.V(3).Info("The pod has the nodeSelector or all the nodeAffinityTerms set for the kubernetes.io/arch label. Ignoring the pod...")
-	pod.ensureLabel(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueNotSet)
+	pod.ensureLabel(utils.NodeAffinityLabel, utils.LabelValueNotSet)
 	pod.publishEvent(corev1.EventTypeNormal, ArchitecturePredicatesConflict, ArchitecturePredicatesConflictMsg)
 }
--- a/controllers/podplacement/pod_model.go
+++ b/controllers/podplacement/pod_model.go
@@ -282,3 +282,43 @@ func (pod *Pod) ensureArchitectureLabels(requirement corev1.NodeSelectorRequirem
 		pod.ensureLabel(utils.ArchLabelValue(value), "")
 	}
 }
+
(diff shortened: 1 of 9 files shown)


controllers/podplacement/pod_model.go
```diff
@@ -273,3 +248,95 @@ func (pod *Pod) ensureArchitectureLabels(requirement corev1.NodeSelectorRequirem
 		pod.ensureLabel(utils.ArchLabelValue(value), "")
 	}
 }
+
+// hasControlPlaneNodeSelector returns true if the pod has a node selector that matches the control plane nodes.
+func (pod *Pod) hasControlPlaneNodeSelector() bool {
```
Prashanth684, line 262: this function is already present in utils [here](https://github.com/openshift/multiarch-tuning-operator/blob/03656c55f205ac3e9c56f51695d468a68ea1a584/pkg/utils/utils.go#L9)

Prashanth684: ah..nvm..you moved it here.

controllers/podplacement/scheduling_gate_mutating_webhook.go
```diff
@@ -69,53 +64,37 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
 	if a.decoder == nil {
 		a.decoder = admission.NewDecoder(a.Scheme)
 	}
-	pod := &corev1.Pod{}
-	err := a.decoder.Decode(req, pod)
+	pod := &Pod{
+		ctx:      ctx,
+		recorder: nil, // do we want to publish events if the pod is ignored?
```
Prashanth684, line 73: i don't think we need to record if pods are ignored by the webhook.

PR #273, opened 2024-09-13
NO-JIRA: Do not build a multiarch bundle

aleskandro:
Fix from the initial onboarding in Konflux


aleskandro:

PR #279, opened 2024-09-18
Add LRU cache with expirable keys

AnnaZivkovic:
Reactivated Caching with LRU Cache with expirable Keys; Added Logs and Metrics

- Re-enabled caching for image architectures.
- Switched to an LRU (Least Recently Used) cache with a maximum of 256 entries and expire after 6 hours to ensure freshness.
- Added logging around cache operations, including cache hits, misses.
- Added metrics to track cache usage, time taken to process cache hits/misses with Histograms, and cache size with Gauge.
- Implemented integration tests to validate caching functionality.


Prashanth684:
nit: cashe -> cache

pkg/testing/image/fake/registry/seed.go
```diff
@@ -45,6 +45,15 @@ func SeedMockRegistry(ctx context.Context) (err error) {
...
 
+func FindMockImage(imageName string) MockImage {
+	for _, image := range GetMockImages() {
+		if image.Name == imageName {
+			return image
+		}
+	}
+	return MockImage{}
+}
+
```
aleskandro: Any call to this method?

pkg/image/inspector.go
```diff
@@ -176,12 +175,28 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
+	fd, err := writeMemFile("constant-name-for-now", authJson)
+	if err != nil {
+		return nil, err
+	}
+	// filepath to our newly created in-memory file descriptor
+	fp := fmt.Sprintf("/proc/self/fd/%d", fd)
+	return os.NewFile(uintptr(fd), fp), nil
+}
+
+func (i *registryInspector) marshaledImagePullSecrets(secrets [][]byte) ([]byte, error) {
```
aleskandro: Should this be a function not part of the registryInspector stuct?

pkg/testing/image/fake/registry/seed.go
```diff
@@ -95,6 +104,12 @@ func GetMockImages() []MockImage {
 	return mockImages
 }
 
+func PushMockImage(ctx context.Context, newImage *MockImage) error {
+	authFilePath := makeAuthFile()
+	_, _, err := newImage.pushImage(ctx, authFilePath)
```
aleskandro, line 100: We should add the image to the mockImages slice.
If one with the same name URL exists in that slice, we should remove it.

pkg/image/cache.go
```diff
@@ -18,39 +18,60 @@ package image
...
 	}
 	architectures, err := c.registryInspector.GetCompatibleArchitecturesSet(ctx, imageReference, secrets)
 	if err != nil {
 		return nil, err
 	}
+
 	c.mutex.Lock()
 	defer c.mutex.Unlock()
-	c.imageRefsArchitectureMap[imageReference] = architectures
+	c.imageRefsArchitectureMap[computeFNV128Hash(imageReference, authJson)] = architectures
```
Prashanth684: +1 here for using fnv!

pkg/image/cache.go
```diff
@@ -18,39 +18,60 @@ package image
...
-		return c.imageRefsArchitectureMap[imageReference], nil
+	authJson, err := c.registryInspector.marshaledImagePullSecrets(secrets)
+	if err != nil {
+		return nil, err
+	}
+
+	if architectures, ok := c.imageRefsArchitectureMap[computeFNV128Hash(imageReference, authJson)]; ok {
+		return architectures, nil
 	}
 	architectures, err := c.registryInspector.GetCompatibleArchitecturesSet(ctx, imageReference, secrets)
```
Prashanth684, line 55: minor optimization: in this case `marshaledImagePullSecrets` will be called twice. while the unmarshaling and marshalling might not cost much - would be good to avoid this step if we can by passing in the authjson.

pkg/image/cache.go
```diff
@@ -18,39 +18,60 @@ package image
...
 }
 
-//nolint:unused
-func newCacheProxy() ICache {
+func (c *cacheProxy) GetRegistryInspector() IRegistryInspector {
+	return c.registryInspector
+}
+
+// Problems
+//  1. misses a cache eviction policy
```
Prashanth684: a basic LRU should be fine for this.

pkg/image/cache.go
aleskandro: doing that would mean changing the signature for both, though, as cacheProxy and RegistryInspector implement the same interface (ICache) to abstract the specific implementation in the callers.

Prashanth684: hm..ok..not a big deal then

pkg/image/inspector.go
AnnaZivkovic: I think you have a point. It does not operate on the data held within the struc

pkg/image/cache.go
```diff
@@ -18,39 +18,56 @@ package image
...
 func (c *cacheProxy) GetCompatibleArchitecturesSet(ctx context.Context, imageReference string, secrets [][]byte) (sets.Set[string], error) {
-	if c.imageRefsArchitectureMap[imageReference] != nil {
-		return c.imageRefsArchitectureMap[imageReference], nil
+	authJson, err := marshaledImagePullSecrets(secrets)
+	if err != nil {
+		return nil, err
+	}
+
+	if architectures, ok := c.imageRefsArchitectureMap[computeFNV128Hash(imageReference, authJson)]; ok {
+		return architectures, nil
```
aleskandro, line 53: can we log here that we had a cache hit for `imageReference` with V(3)?

pkg/image/cache.go
```diff
@@ -18,39 +18,56 @@ package image
...
 	}
 	architectures, err := c.registryInspector.GetCompatibleArchitecturesSet(ctx, imageReference, secrets)
 	if err != nil {
 		return nil, err
 	}
+
 	c.mutex.Lock()
```
aleskandro: before the mutex is locked, can we log that we had a cache miss for `imageReference`, V(3) too.

pkg/image/facade.go
```diff
@@ -44,10 +44,10 @@ func (i *Facade) StoreGlobalPullSecret(pullSecret []byte) {
 
 func newImageFacade() *Facade {
 	// TODO: restore to use the cacheProxy once an eviction policy is implemented
```
aleskandro:
```suggestion
```

pkg/image/cache.go
Prashanth684: will counting number of cache hits be a useful metric? it could give us insights into the percentage of time images are reused

aleskandro: A gauge with the size of the cache is set as 'nice to have' in the subsequent story to implement eviction.

We could also store both the HIT and MISS as histograms so that we can get both the count and the time for a HIT or a MISS separately.

aleskandro:
/lgtm
/approve

Thanks @AnnaZivkovic

pkg/image/metrics/common_metrics.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func buckets() []float64 {
+	return []float64{
+		// Values are in seconds
+		0.001, 0.002, 0.005, 0.010, 0.020, 0.050, // exponential-like buckets for values < 0.1
+		0.100, 0.200, 0.300, 0.400, 0.500, 0.600, // linear buckets for the 0.1 <= values < 1
+		0.700, 0.800, 0.900, 1.000, 2.000, 4.000, // exponential buckets for values >= 1
+	}
+}
```
aleskandro: /hold

can we move this to the utils package so that we have only one definition for them?

vendor/github.com/hashicorp/golang-lru/v2/expirable/expirable_lru.go
```diff
@@ -0,0 +1,335 @@
+package expirable
```
aleskandro, line 1: if still possible, can you split this commit into two? (1) feature, (2) vendoring.

pkg/image/cache.go
```diff
@@ -18,39 +18,61 @@ package image
 
 import (
 	"context"
+	"encoding/hex"
+	"hash/fnv"
+	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 	"sync"
 
 	"k8s.io/apimachinery/pkg/util/sets"
```
aleskandro:
```suggestion
	"sync"

	"k8s.io/apimachinery/pkg/util/sets"

    ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
```

aleskandro:
@lwan-wanglin note that we missed an optional: true in the 4.17 test config and it is not set as required.

pkg/image/metrics/common_metrics.go
```diff
@@ -0,0 +1,50 @@
...
+
+func InitCommonMetrics() {
+	onceCommon.Do(func() {
+		InspectionGauge = prometheus.NewGauge(
+			prometheus.GaugeOpts{
+				Name: "mto_inspection_cache_size",
+				Help: "Current size of the MTO inspection cache",
+			},
+		)
+		TimeToInspectImageGivenHit = prometheus.NewHistogram(
```
Prashanth684, line 28: our cache is pretty simplistic..will this metric be of any relevance?

aleskandro: the time should always be in the lowest buckets, but you wanted to count how many hits we had. This will allow calculating that rate too (and also maintain the time required to access the cache).

vendor/github.com/hashicorp/golang-lru/v2/expirable/expirable_lru.go
AnnaZivkovic: @aleskandro I can split it in terms of adding gauge and adding histogram otherwise we break having "working" commits?

lwan-wanglin:
> @lwan-wanglin note that we missed an optional: true in the 4.17 test config and it is not set as required.
 fixed in pr https://github.com/openshift/release/pull/57201, PTAL, thanks!

pkg/image/metrics/common_metrics.go
```diff
@@ -0,0 +1,42 @@
+package metrics
+
+import (
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
+	"sync"
+
+	"github.com/prometheus/client_golang/prometheus"
+	metrics2 "sigs.k8s.io/controller-runtime/pkg/metrics"
```
aleskandro:
```suggestion
	"sync"

	"github.com/prometheus/client_golang/prometheus"
	metrics2 "sigs.k8s.io/controller-runtime/pkg/metrics"
	
	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
```

go.mod
```diff
@@ -85,7 +85,7 @@ require (
 	github.com/gorilla/mux v1.8.1 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.21.0 // indirect
 	github.com/hashicorp/golang-lru/arc/v2 v2.0.5 // indirect
-	github.com/hashicorp/golang-lru/v2 v2.0.5 // indirect
+	github.com/hashicorp/golang-lru/v2 v2.0.5
```
aleskandro, line 88: @AnnaZivkovic this should stay on the other commit (with go.sum changes too).

only the changes in the vendor/ folder should stay on a separate commit.

This way we ensure the code is functional, and keep vendored modules outside features commits

aleskandro:
We need to open a new story for the polling goroutine to expose the size of the cache as it is expirable and the size can change (shrink in particular) regardless of the cache being used by the GetCompatibleArchitectureSet method

aleskandro:
/override "Red Hat Konflux/Red Hat Konflux/multiarch-tuning-operator-enterprise-contract/multiarch-tuning-operator"

The failure is due to another issue in the base image. We only need to wait for this to eventually pass.
Overriding to make this land

aleskandro:
/override "Red Hat Konflux/Red Hat Konflux/multiarch-tuning-operator-enterprise-contract/multiarch-tuning-operator"

aleskandro:

PR #252, opened 2024-09-04
Revert 8928d81e5d: Extract the single-arch manifest from the bundle to allow extraction on arm64 nodes

aleskandro:
Now that ci-operator doesn't trigger a manifest-list build for bundles, we can revert the workaround in 8928d81e5d.

Related to openshift/ci-tools#4307 
Related to [DPTP-4143](https://issues.redhat.com//browse/DPTP-4143)

PR #286, opened 2024-09-23
fix path env variable in deploy-and-e2e.sh

lwan-wanglin:
the pr fix below issues
- when deploy-and-e2e.sh is running twice, the symbolic link has been created to `/tmp/bin/kubectl` but the path is reset without `/tmp/bin`,  wich kubectl and` ln -s` command will fail
- add WaitForMCPComplete in AfterSuite, otherwise the step `*-gather-audit-logs` might fail

### DIFF
--- a/hack/deploy-and-e2e.sh
+++ b/hack/deploy-and-e2e.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 set -euxo pipefail
 
+mkdir -p /tmp/bin
+export PATH=/tmp/bin:${PATH}
 if ! which kubectl >/dev/null; then
-  mkdir -p /tmp/bin
-  export PATH=/tmp/bin:${PATH}
   ln -s "$(which oc)" "/tmp/bin/kubectl"
 fi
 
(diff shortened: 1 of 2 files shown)


hack/deploy-and-e2e.sh
```diff
@@ -38,7 +38,7 @@ if [ "${USE_OLM:-}" == "true" ]; then
 
   export KUBECONFIG="${OLD_KUBECONFIG}"
   export JUNIT_SUFFIX="-olm"
-  operator-sdk run bundle "${OO_BUNDLE}" -n "${NAMESPACE}" --security-context-config restricted --timeout=10m
+  operator-sdk run bundle "${OO_BUNDLE}" -n "${NAMESPACE}" --security-context-config restricted --timeout=15m
```
aleskandro: Why this increased timeout?

lwan-wanglin: I see several times the command hit timeout when wait for the condition
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_multiarch-tuning-operator/280/pull-ci-openshift-multiarch-tuning-operator-main-e2e-gcp-multi-operator-olm/1837239546450808832

aleskandro: Yeah, I think it's not because there's a need for more time but it depends on something not properly done in the builds. I never reproduced that issue locally. So this may also be related to #252

lwan-wanglin: got it

PR #293, opened 2024-09-25
Debug info when the deployment of the operator fails and minor fixes

aleskandro:


lwan-wanglin:
Can you help to expand the gingo timeout in [ci-test.sh](https://github.com/openshift/multiarch-tuning-operator/blob/main/hack/ci-test.sh) in your pr? I added WaitForMCPComplete in AfterSuite, and forget to expand the gingo timeout, the run may cost about 30 mins, sometime will over 30 mins

PR #298, opened 2024-09-30
MULTIARCH-5029: Explicitly set AutomountServiceAccountToken in pod specs and disable it by default in the service account specs

aleskandro:


aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-operator-bundle" "Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-operator"

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -356,6 +356,7 @@ spec:
                         operator: In
                         values:
                         - linux
+              automountServiceAccountToken: true
```
Prashanth684, line 359: even if you don't set this - won't kubelet mount the serviceaccount's API creds? the question is - do we need to set it to true?

aleskandro: yes, we do as we talk to the API.

The default is true, but we've been asked to check if we can make it false.

I'm making it false for the serviceAccounts so that pods need to explicitly ask for the creds to be mounted by setting it to true.

aleskandro: (In our case, all the workloads need the creds)

aleskandro: I'm seeing that just as a good practice... I'm not sure it can actually be considered a 'weakness' given we just need them to be mounted.

Prashanth684: yeah better to not mount it in that case.

PR #307, opened 2024-10-03
MULTIARCH-4965: snapshots multiarch-tuning-operator-1-0-jwfr7

AnnaZivkovic:
oc describe snapshots multiarch-tuning-operator-nwwnk
```
Spec:
  Application:  multiarch-tuning-operator-1-0
  Artifacts:
  Components:
    Container Image:  quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator-1-0/multiarch-tuning-operator-1-0@sha256:531048914f493b44f8b1c6f23adaf506c2a67bee23fb92eabc3d7938f5e5d56d
    Name:             multiarch-tuning-operator-1-0
    Source:
      Git:
        Context:         ./
        Dockerfile URL:  Dockerfile
        Revision:        b31311f3e642b2ba4ac71b32a675395962d5dd38
        URL:             https://github.com/openshift/multiarch-tuning-operator
    Container Image:     quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator-1-0/multiarch-tuning-operator-bundle-1-0@sha256:1a3cb62081d6e6c6ce874b861bf94906c6f18fae39e2ad85b7c4dba1481dddd5
    Name:                multiarch-tuning-operator-bundle-1-0
    Source:
      Git:
        Revision:  634b61eaa9362e6708c5ec01ee7df214603ea482
        URL:       https://github.com/openshift/multiarch-tuning-operator
```

To upgrade the operator via web console follow the instructions [here](https://docs.openshift.com/container-platform/4.16/operators/admin/olm-upgrading-operators.html#olm-changing-update-channel_olm-upgrading-operators)  

To upgrade via `oc` update the channel field in the subscription to change the channel the operator uses for updates.

Find the needed subscription
```oc get subscriptions -n <namespace>```

Run the following command to edit the subscription:

```oc edit subscription <subscription-name> -n <namespace>```

Once in the editor, find the spec.channel field and change it to the desired channel. For example:
```
spec:
  channel: stable
```
Change stable to your desired channel, e.g., tech-preview, alpha, etc. If the update approval method is set to Manual then  the update needs to be manually approved before installation begins otherwise it will begin automatically.

### CODE REVIEW

File: catalog/multiarch-tuning-operator/index.yaml [hunk 1]
```diff
@@ -74,4 +81,56 @@ relatedImages:
...
+  - type: olm.bundle.object
+    value:
+      data: eyJhcGlWZXJzaW9uIjoidjEiLCJraW5kIjoiU2VydmljZSIsIm1ldGFkYXRhIjp7ImNyZWF0aW9uVGltZXN0YW1wIjpudWxsLCJuYW1lIjoibXVsdGlhcmNoLXR1bmluZy1vcGVyYXRvci1jb250cm9sbGVyLW1hbmFnZXItc2VydmljZSJ9LCJzcGVjIjp7InBvcnRzIjpbeyJuYW1lIjoid2ViaG9vayIsInBvcnQiOjQ0MywidGFyZ2V0UG9ydCI6OTQ0M30seyJuYW1lIjoibWV0cmljcyIsInBvcnQiOjg0NDMsInRhcmdldFBvcnQiOjg0NDN9XSwic2VsZWN0b3IiOnsiY29udHJvbC1wbGFuZSI6ImNvbnRyb2xsZXItbWFuYWdlciJ9fSwic3RhdHVzIjp7ImxvYWRCYWxhbmNlciI6e319fQ==
+  - type: olm.bundle.object
+    value:
```
aleskandro: 0.9 should keep to the same bundle as already in prod (we won't release a new 0.9 bundle)

AnnaZivkovic: Should I close this pr then? I am not sure what update the is to do if we are not adding a new bundle.

aleskandro: No, we should add the 1.0 bundle as an additional node/package and an edge from 0.9 to 1.0. What we shouldn't do is replacing the 0.9 bundle digest: we'd keep the current one, that we released

aleskandro:
/lgtm
/approve
Thanks!

PR #318, opened 2024-10-08
MULTIARCH-4307: Add fix for e2e running in cluster-wide proxy based cluster

lwan-wanglin:
Add proxy info for e2e test cases which need connection to network in the container for running e2e on cluster-wide proxy based cluster.


lwan-wanglin:
@aleskandro I see the multiarch jobs already moved back to build10 in https://github.com/openshift/release/pull/57614, I don't know why the job in this pr still choose build11?

PR #330, opened 2024-10-14
OCPBUGS-42165: Do not cache images from containers with image pull policy 'Always'

aleskandro:
When a container in a pod uses the ImagePullPolicy `Always`, we should not use the cache to retrieve its metadata.

While changing a behavior in the pod_builder to accommodate a test for this PR, I noticed an issue with the integration test of the operators bundle cases. This PR includes the fix as the new behavior of the `WithContainerImages` function would let the test to fail.

### DIFF
--- a/controllers/podplacement/pod_model.go
+++ b/controllers/podplacement/pod_model.go
@@ -37,6 +37,11 @@ var (
 	imageInspectionCache image.ICache = image.FacadeSingleton()
 )
 
+type containerImage struct {
+	imageName string
+	skipCache bool
+}
+
 type Pod struct {
 	corev1.Pod
 	ctx      context.Context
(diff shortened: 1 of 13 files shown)


lwan-wanglin:
- test on image  `brew.registry.redhat.io/rh-osbs/iib:835019 ` without the pr merged failed the below steps.

- test on image `quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator/multiarch-tuning-operator:on-pr-7f49bf6824cc7bd9c298915daee243093839528e` with the pr merged passed the below steps.

test steps:
1. create deployment with image quay.io/openshifttest/tmpvalid:latest which support [amd64 ppc64le], check nodeAffinity set by PPC, the value should be [amd64 ppc64le]
2. rebuild image with the same tag to support [arm64 s390x] , recreate pod, check nodeAffinity set by PPC, the value should be [arm64 s390x]

PR #334, opened 2024-10-16
WIP: OCPBUGS-41554: remove service-ca.crt

aleskandro:
The pod placement controller does not need the cluster service-ca.crt to function. In fact, it  should not use cluster service-ca.crt CA to inspect image arch if the user did not add it to trusted anchors explicitly.

### DIFF
--- a/bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
@@ -27,7 +27,7 @@ metadata:
     categories: OpenShift Optional, Other
     console.openshift.io/disable-operand-delete: "false"
     containerImage: registry.ci.openshift.org/origin/multiarch-tuning-operator:main
-    createdAt: "2024-09-30T12:29:39Z"
+    createdAt: "2024-10-16T20:07:51Z"
     features.operators.openshift.io/cnf: "false"
     features.operators.openshift.io/cni: "false"
     features.operators.openshift.io/csi: "false"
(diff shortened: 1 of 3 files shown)


bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -432,12 +432,6 @@ spec:
               - name: ca-projected-volume
                 projected:
                   sources:
-                  - configMap:
-                      items:
-                      - key: service-ca.crt
```
Prashanth684, line 437: isn't this needed to communicate with the internal registry?

aleskandro: it looks like the internal registry's CA is set in the /etc/docker/certs.d folder in the nodes, and only these hostnames are validate with it:
```
ls /etc/docker/certs.d/
image-registry.openshift-image-registry.svc.cluster.local:5000  image-registry.openshift-image-registry.svc:5000
```
Therefore, any other registry created internally and exposed with a route that uses the openshift service-ca needs to be added in the registry-cas configmap to be considered valid by the nodes and kubelet to pull the images. We are essentially going to do the same: do not consider the service-ca as a global trusted one and rely on the registry config CR and related configmap for the per-registry custom CAs that users may want to use.

Prashanth684: > it looks like the internal registry's CA is set in the /etc/docker/certs.d folder in the nodes,

Do we mount these certs then, so we can access the internal registry if needed?

Prashanth684: it looks like the image-registry-certificates configmap has certs for the internal registry as well, so we should be good.

PR #339, opened 2024-10-18
MULTIARCH-4973: change all the builders to build pointers

lwan-wanglin:
the pr fix the below issue:
- change all the builders to build pointers
- rename filename without `_builder` suffix


aleskandro:
Reverting for now as a few cases failed in the post-submits and this is not a priority for the release.

PR #346, opened 2024-10-23
MULTIARCH-4973: Add more log printing

lwan-wanglin:
the pr fix the below issue:
- move basic function from test suite to testing/framework
- add more By statements
- fix a pontential bug in registry.go

### DIFF
--- a/pkg/e2e/operator/pod_placement_config_test.go
+++ b/pkg/e2e/operator/pod_placement_config_test.go
@@ -8,7 +8,6 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


AnnaZivkovic:
@lwan-wanglin  Why we’re moving basic functions from the test suite to the testing/framework? Just curious about the benefits

lwan-wanglin:
> @lwan-wanglin Why we’re moving basic functions from the test suite to the testing/framework? Just curious about the benefits

moving functions that are not specific to a given test suite to testing/framework so that they can be consumed by more suites instead of duplicating code, like function `verifyPodNodeAffinity` and `verifyPodLabels` are existing in both file `pkg/e2e/operator/pod_placement_config_test.go` and `pkg/e2e/podplacement/pod_placement_test.go`

PR #362, opened 2024-11-11
MULTIARCH-5053: Make e2e pass on hypershift managed clusters

lwan-wanglin:
I have tested all e2e cases locally in a hosted cluster(iscp, pull-secret,image.config need to be configurated at mgmt hostedcluster CR side). then we will skip those test cases which need to config at mgmt side from e2e.  there is no different between ocp and hosted cluster, so it doesn't need to make e2e complex.
This pr fix below issue
- skip registry and image config tests on hypershift managed offerings
- fix Build tests, the supported arch is related to the payload of the cluster, so it's enough to just verify label `multiarch.openshift.io/node-affinity:set` and `multiarch.openshift.io/scheduling-gate: removed`


aleskandro:

PR #364, opened 2024-11-19
OCPBUGS-44703: Invalidate registry cache before inspection of pods

Prashanth684:
Registry config is cached by the container image library. Invalidate the cache each time we inspect images for a pod in the case of any updates.

A follow up refinement to this should be to only invalidate cache when the MCP has been updated for optimization.


aleskandro (commented):
Thanks @Prashanth684 . good catch!

/lgtm
/approve
/cherrypick v1.0

aleskandro:

PR #365, opened 2024-11-19
Verify signature only for the single manifest image, not for the top level manifest

Prashanth684:
According to the containers team, IsRunningImageAllowed is as per practice, expected to pass for the per-platform manifest, not for the top-level one. The correct location to verify signatures is after instanceDigest is chosen.

xref: https://redhat-internal.slack.com/archives/CBBJY9GSX/p1731953705698749?thread_ts=1731952085.942329&cid=CBBJY9GSX

### DIFF
--- a/pkg/image/inspector.go
+++ b/pkg/image/inspector.go
@@ -111,20 +111,6 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 		log.Error(err, "Error creating the PolicyContext")
 		return nil, err
 	}
-	if allowed, err := policyCtx.IsRunningImageAllowed(ctx, image.UnparsedInstance(src, nil)); !allowed {
-		// IsRunningImageAllowed returns true iff the policy allows running the image.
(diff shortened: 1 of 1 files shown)


pkg/image/inspector.go
```diff
@@ -146,7 +131,24 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
+			// false and valid error
+			log.V(3).Info("The signature policy JSON file configuration does not allow inspecting this image",
+				"validationError", e)
+			return nil, e
+		}
+		log.Error(err, "Unable to perform the signature validation")
+		return nil, err
+	}
+
+	supportedArchitectures = sets.New[string]()
```
aleskandro: can we keep this above before https://github.com/openshift/multiarch-tuning-operator/pull/365/files#diff-9642b2d113dcd6f29959b6098aa6c393e5f037047e3a501502eca81f8a70e2a1R116 ?

Prashanth684: oops..diidn't notice it was used there...will move

PR #377, opened 2024-11-25
MULTIARCH-5181: pod-level circuit-breaking for image registry inspection errors

aleskandro:


controllers/podplacement/pod_reconciler.go
```diff
@@ -77,52 +78,65 @@ func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
-		return ctrl.Result{}, err
+	// Prepare the requirement for the node affinity.
+	psdl, err := r.pullSecretDataList(ctx, pod)
+	pod.handleError(err, "Unable to retrieve the image pull secret data for the pod.")
+	// If no error occurred when retrieving the image pull secret data, set the node affinity.
+	if err == nil {
+		_, err = pod.SetNodeAffinityArchRequirement(psdl)
+		pod.handleError(err, "Unable to set the node affinity for the pod.")
+	}
+	if pod.maxRetries() && err != nil {
```
Prashanth684, line 126: one worry i have is that we might get rate limited by the registry if it is slow..maybe a backoff of sorts would be good to implement.. i.e, if the same pod is again queued, we might want to give it some time

aleskandro: Hi @Prashanth684 the (exponential) backoff is already implemented through the controller-runtime facilities.

PR #384, opened 2024-12-03
MULTIARCH-5056: cache the ClusterPodPlacementConfig using informers

AnnaZivkovic:
Add an informer to watch the ClusterPodPlacementConfig

### DIFF
--- a/controllers/podplacement/clusterpodplacementconfg_informer.go
+++ b/controllers/podplacement/clusterpodplacementconfg_informer.go
@@ -1,0 +1,121 @@
+package podplacement
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/go-logr/logr"
+	multiarchv1beta1 "github.com/openshift/multiarch-tuning-operator/apis/multiarch/v1beta1"
+	"github.com/openshift/multiarch-tuning-operator/pkg/informers"
+	"k8s.io/client-go/tools/cache"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// CPPCSyncer syncs ClusterPodPlacementConfig resources using an informer.
+type CPPCSyncer struct {
+	mgr manager.Manager
+	log logr.Logger
+}
+
+// NewCPPCSyncer creates a new CPPCSyncer.
+func NewCPPCSyncer(mgr manager.Manager) *CPPCSyncer {
+	return &CPPCSyncer{
+		mgr: mgr,
+	}
+}
+
+// Start initializes the CPPC informer and starts syncing.
+func (s *CPPCSyncer) Start(ctx context.Context) error {
+	s.log = log.FromContext(ctx, "handler", "CPPCSyncer")
+	s.log.Info("Starting CPPC Syncer")
+	mgr := s.mgr
+
+	ic := informers.CacheSingleton()
+	// Get informer for ClusterPodPlacementConfig
+	CPPCInformer, err := mgr.GetCache().GetInformerForKind(ctx, multiarchv1beta1.GroupVersion.WithKind("ClusterPodPlacementConfig"))
+	if err != nil {
+		s.log.Error(err, "Error getting informer for ClusterPodPlacementConfig")
+		return err
+	}
+
+	// Register event handlers
+	_, err = CPPCInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    s.onAdd(ic),
+		UpdateFunc: s.onUpdate(ic),
+		DeleteFunc: s.onDelete(ic),
+	})
+	if err != nil {
+		s.log.Error(err, "Error registering handler for ClusterPodPlacementConfig")
+		return err
+	}
+
+	return nil
+}
+
+// onAdd handles the addition of a ClusterPodPlacementConfig.
+func (s *CPPCSyncer) onAdd(ic informers.ICache) func(obj interface{}) {
+	return func(obj interface{}) {
+		CPPC, ok := obj.(*multiarchv1beta1.ClusterPodPlacementConfig)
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -265,6 +279,7 @@ func bindFlags() {
 	flag.BoolVar(&enableClusterPodPlacementConfigOperandWebHook, "enable-ppc-webhook", false, "Enable the pod placement config operand webhook")
 	flag.BoolVar(&enableClusterPodPlacementConfigOperandControllers, "enable-ppc-controllers", false, "Enable the pod placement config operand controllers")
 	flag.BoolVar(&enableOperator, "enable-operator", false, "Enable the operator")
+	flag.BoolVar(&enableCPPCInformer, "enable-ppc-informer", false, "Enable informer for ClusterPodPlacementConfig")
```
Prashanth684, line 282: why? in what case would we not want this? why not make it default?

AnnaZivkovic: This is one section of work of this [epic](https://issues.redhat.com/browse/MULTIARCH-4970). The thought is to disable it until [MULTIARCH-4974](https://issues.redhat.com/browse/MULTIARCH-4974) is ready. Then we can make it default.

controllers/podplacement/clusterpodplacementconfg_informer.go
```diff
@@ -64,29 +71,29 @@ func (s *CPPCSyncer) onAdd(ic informers.ICache) func(obj interface{}) {
...
 		if err != nil {
 			s.log.Error(err, "Error marshalling ClusterPodPlacementConfig")
 		}
 
-		jsonCPPCWebhook, err := json.Marshal(CPPCWebhook)
-		if err != nil {
-			s.log.Error(err, "Error marshalling WebhookConfiguration")
+		jsonWebhook := json.RawMessage{}
+		// Get the MutatingWebhookConfiguration by name
+		mutatingWebhookConfig, err := s.clientSet.AdmissionregistrationV1().MutatingWebhookConfigurations().Get(context.TODO(), utils.PodMutatingWebhookConfigurationName, metav1.GetOptions{})
```
AnnaZivkovic: @aleskandro @Prashanth684 I ended up using an API call to retrieve the mutating webhook instead of adding another informer, with the thought that it would reduce overhead. Thoughts?

controllers/podplacement/clusterpodplacementconfg_informer.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Start initializes the CPPC informer and starts syncing.
+func (s *CPPCSyncer) Start(ctx context.Context) error {
+	s.log = log.FromContext(ctx, "handler", "CPPCSyncer")
+	s.log.Info("Starting CPPC Syncer")
+	mgr := s.mgr
+
+	ic := informers.CacheSingleton()
+	// Get informer for ClusterPodPlacementConfig
+	CPPCInformer, err := mgr.GetCache().GetInformerForKind(ctx, multiarchv1beta1.GroupVersion.WithKind("ClusterPodPlacementConfig"))
```
aleskandro: We should have the string `ClusterPodPlacementConfig` as a constant somewhere already.

main.go
aleskandro: Also, when implementing the namespace-scoped PPC, the controller will not have to run the CPPC informer

controllers/podplacement/clusterpodplacementconfg_informer.go
aleskandro: Why are we storing the MutatingWebhookConfiguration?

pkg/informers/cache.go
```diff
@@ -0,0 +1,49 @@
...
+	"sync"
+)
+
+var (
+	singletonSystemConfigInstance ICache
+	once                          sync.Once
+)
+
+type ClusterPodPlacementConfigSyncer struct {
+	config        json.RawMessage // For raw JSON object
```
aleskandro: Can we just store it as ClusterPodPlacementConfig?

aleskandro (commented):
Hi @AnnaZivkovic thanks for the updates!

Could you implement one simple integration test for this feature?

We will use this feature in components other than the operator, but I would feel good with having a test case implemented in the operator integration test suite as it's where we can easily create, update and delete the CPPC object.

For example, the test case can consist of:
1. (Starting point) No CPPC
2. Get the CacheSingleton and verify there is no CPPC stored
3. Create a CPPC and wait for/verify reconciliation as we already do
4. get the informers and verify there is a CPPC
5. Update the CPPC
6. Get the updated CPPC's resourceVersion
7. Get the cached CPPC and verify the resourceVersion at point 6 is the same as the cached one
8. Delete the cached CPPC
9. Verify there is no cached CPPC

controllers/podplacement/clusterpodplacementconfg_informer.go
```diff
@@ -0,0 +1,121 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// onAdd handles the addition of a ClusterPodPlacementConfig.
+func (s *CPPCSyncer) onAdd(ic informers.ICache) func(obj interface{}) {
+	return func(obj interface{}) {
+		CPPC, ok := obj.(multiarchv1beta1.ClusterPodPlacementConfig)
```
lwan-wanglin: I am new for this, and I have a question for why we don't use pointer here? I see we use the pointer for func `onUpdate` and `onDelete`.

AnnaZivkovic: I wanted to avoid caching a pointer. The cached value of ClusterPodPlacementConfig should be isolated from changes. However, passing non-pointer objects is cumbersome. 

I switched to using pointers simplified object handling. To maintain cache integrity, I used DeepCopy when caching. Hopefully this approach balances the flexibility of pointers with the need for a consistent cache.

aleskandro (commented):
/lgtm

PR #393, opened 2024-12-10
MULTIARCH-5157: Update go version to 1.23

AnnaZivkovic:
updated go 1.22 -> 1.23.4
go get -u
make vendor

Updated  events.NewKubeRecorder to have the correct number of arguments


aleskandro:
Given ocp4.18 is not going to be based on go1.23, should we just update the k8s api?

AnnaZivkovic:
Closing to delay work until 4.19

PR #424, opened 2025-01-15
changing the test suite run cases concurrently

lwan-wanglin:
The pr implements the below things:
- changing the test suite run cases concurrently, keep operator testcases running` Serial`
- Fix funcs which exists data race

### DIFF
--- a/controllers/podplacement/suite_test.go
+++ b/controllers/podplacement/suite_test.go
@@ -38,9 +38,12 @@ import (
 	v1 "k8s.io/api/admissionregistration/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -52,6 +55,7 @@ import (
 
 	"github.com/panjf2000/ants/v2"
 
+	"github.com/openshift/multiarch-tuning-operator/pkg/testing/framework"
 	testingutils "github.com/openshift/multiarch-tuning-operator/pkg/testing/framework"
 	"github.com/openshift/multiarch-tuning-operator/pkg/testing/image/fake/registry"
 	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
@@ -73,6 +77,13 @@ var (
 	suiteLog = ctrl.Log.WithName("setup")
 )
 
+type sharedData struct {
+	Kubeconfig       api.Config `json:"kubeconfig"`
+	RegistryAddress  string     `json:"registryAddress"`
+	RegistryCert     []byte     `json:"registryCert"`
+	RegistryCertPath string     `json:"registryCertpath"`
+}
+
 func TestAPIs(t *testing.T) {
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "Controllers Integration Suite", Label("integration"))
@@ -80,12 +91,13 @@ func TestAPIs(t *testing.T) {
 
 var _ = BeforeAll
 
-var _ = BeforeSuite(func() {
+var _ = SynchronizedBeforeSuite(func() []byte {
 	suiteLog = zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true), zap.Level(zapcore.Level(-5)))
 
 	logf.SetLogger(suiteLog)
 	SetDefaultEventuallyPollingInterval(5 * time.Millisecond)
 	SetDefaultEventuallyTimeout(5 * time.Second)
+	By("Set up a shared test environment and local registry for testing on process 1 node")
 	wg := &sync.WaitGroup{}
 	testingutils.DecorateWithWaitGroup(wg, startTestEnv)
 	testingutils.DecorateWithWaitGroup(wg, startRegistry)
@@ -93,7 +105,6 @@ var _ = BeforeSuite(func() {
 	testingutils.DecorateWithWaitGroup(wg, seedRegistry)
 	testingutils.DecorateWithWaitGroup(wg, seedK8S)
 	wg.Wait()
-
 	var err error
 	dir, err = os.MkdirTemp("", "multiarch-tuning-operator")
 	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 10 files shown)


controllers/operator/clusterpodplacementconfig_controller_test.go
```diff
@@ -38,7 +38,7 @@ import (
 	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
 )
 
-var _ = Describe("Controllers/ClusterPodPlacementConfig/ClusterPodPlacementConfigReconciler", Ordered, func() {
+var _ = Describe("Controllers/ClusterPodPlacementConfig/ClusterPodPlacementConfigReconciler", Serial, Ordered, func() {
```
aleskandro: Doesn't `Ordered` imply `Serial`?

pkg/systemconfig/types.go
```diff
@@ -49,30 +50,39 @@ var (
 	registriesCertsDir,
 	registriesConfPath,
 	policyConfPath string
+	mu sync.Mutex
```
aleskandro, line 53: we can use a RWMutex here. See example below.

pkg/testing/image/fake/registry/registry.go
```diff
@@ -135,6 +139,8 @@ func buildRegistryTLSConfig() (*registryTLSConfig, error) {
 	}
 
 	// create a folder for the certs
+	mu.Lock()
+	defer mu.Unlock()
```
aleskandro: Do we need this?

Does this method now create N different folders as the number of processes and that var will have a different value for each 'runner'?

pkg/testing/image/fake/registry/registry.go
```diff
@@ -34,7 +35,10 @@ type registryTLSConfig struct {
 	caCertBytes     []byte
 }
 
-var perRegistryCertDirPath string
+var (
+	perRegistryCertDirPath string
```
aleskandro, line 39: what if we include this field in the RegistryConfig struct?

pkg/systemconfig/types.go
```diff
@@ -49,30 +50,39 @@ var (
...
 )
 
 func DockerCertsDir() string {
+	mu.Lock()
+	defer mu.Unlock()
 	if dockerCertsDir == "" {
 		dockerCertsDir = lookupEnvOr("DOCKER_CERTS_DIR", "/etc/docker/certs.d")
 	}
 	return dockerCertsDir
 }
```
aleskandro:
```suggestion
func DockerCertsDir() string {
	if dockerCertsDir == "" {
	    once.Do(func() {
  		  dockerCertsDir = lookupEnvOr("DOCKER_CERTS_DIR", "/etc/docker/certs.d")
  		}
	}

	return dockerCertsDir
}
```

What about this?

aleskandro: My concern here is that using a Mutex might slightly impact the performance of the registryInspector that uses the result of this function for every image it inspects.

Alternatively, we may use an RWMutex.

controllers/podplacement/suite_test.go
```diff
@@ -109,9 +112,54 @@ var _ = BeforeSuite(func() {
...
+	registryCert, err = base64.StdEncoding.DecodeString(registryC)
+	Expect(err).NotTo(HaveOccurred())
+	// Sync test cluster environment
+	cfgPath := decodedData["cfgPath"].(string)
+	cfg, err := clientcmd.BuildConfigFromFlags("", cfgPath)
+	Expect(err).NotTo(HaveOccurred(), "Error loading kubeconfig")
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
 })
```
aleskandro, line 165: Should we remove https://github.com/openshift/multiarch-tuning-operator/pull/424/files#diff-48315070288e341898824f044c70596b21f8be94b403d389fb931dd66277236fR191

controllers/podplacement/suite_test.go
```diff
@@ -109,9 +112,54 @@ var _ = BeforeSuite(func() {
...
+	By("Prepare shared data and Pass to all processes")
+	// Save envtest.Environment to a kubeconfig file and share with all processes
+	kc := fromEnvTestConfig(cfg)
+	kubeconfigFileName := "envtest.kubeconfig"
+	err = testingutils.WriteToFile(dir, kubeconfigFileName, string(kc))
+	Expect(err).NotTo(HaveOccurred(), "failed to save kubeconfig")
+	// The registry.perRegistryCertDirPath is used by registry.PushMockImage,
+	// need to pass to all processes
+	registryCertPath := registry.GetCertPath()
+	data := map[string]interface{}{
```
aleskandro: Can we define a dedicated struct?

controllers/podplacement/suite_test.go
```diff
@@ -334,3 +382,32 @@ func getMutatingWebHook() *v1.MutatingWebhookConfiguration {
 		},
 	}
 }
+
+func fromEnvTestConfig(cfg *rest.Config) []byte {
```
aleskandro:
```suggestion
func fromEnvTestConfig(cfg *rest.Config) api.Config {
```

controllers/podplacement/suite_test.go
```diff
@@ -334,3 +382,32 @@ func getMutatingWebHook() *v1.MutatingWebhookConfiguration {
...
+			cfg.Username: {
+				ClientKeyData:         cfg.KeyData,
+				ClientCertificateData: cfg.CertData,
+			},
+		},
+		CurrentContext: contextName,
+	}
+	data, err := clientcmd.Write(c)
+	Expect(err).NotTo(HaveOccurred(), "failed to write kubeconfig")
+	return data
```
aleskandro:
```suggestion
	return c
```

controllers/podplacement/suite_test.go
```diff
@@ -109,9 +112,54 @@ var _ = BeforeSuite(func() {
...
+	// Save envtest.Environment to a kubeconfig file and share with all processes
+	kc := fromEnvTestConfig(cfg)
+	kubeconfigFileName := "envtest.kubeconfig"
+	err = testingutils.WriteToFile(dir, kubeconfigFileName, string(kc))
+	Expect(err).NotTo(HaveOccurred(), "failed to save kubeconfig")
+	// The registry.perRegistryCertDirPath is used by registry.PushMockImage,
+	// need to pass to all processes
+	registryCertPath := registry.GetCertPath()
+	data := map[string]interface{}{
+		"cfgPath":          filepath.Join(dir, kubeconfigFileName),
```
aleskandro:
```suggestion
	// The registry.perRegistryCertDirPath is used by registry.PushMockImage,
	// need to pass to all processes
	registryCertPath := registry.GetCertPath()
	data := map[string]interface{}{
		"kubeConfig":          kc,
```

controllers/podplacement/suite_test.go
```diff
@@ -109,9 +112,54 @@ var _ = BeforeSuite(func() {
...
+	// Sync registryAddress and registryCert
+	registryAddress = decodedData["registryAddress"].(string)
+	registryC := decodedData["registryCert"].(string)
+	registryCert, err = base64.StdEncoding.DecodeString(registryC)
+	Expect(err).NotTo(HaveOccurred())
+	// Sync test cluster environment
+	cfgPath := decodedData["cfgPath"].(string)
+	cfg, err := clientcmd.BuildConfigFromFlags("", cfgPath)
+	Expect(err).NotTo(HaveOccurred(), "Error loading kubeconfig")
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
```
aleskandro:
```suggestion
	// Sync test cluster environment
	kc := decodedData["kubeConfig"] // <<<---- This will be api.Config once you define the struct
	ocg := clientcmd.NewDefaultClientConfig(kc, &clientcmd.ConfigOverrides{})
	cfg, err = ocg.ClientConfig()
	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
```

controllers/podplacement/suite_test.go
lwan-wanglin: we might can not, the seedK8S() needs the k8sclient.

controllers/operator/clusterpodplacementconfig_controller_test.go
lwan-wanglin: All Ordered cases will run on an ordered container,  but they can be run parallel wrt other specs,  I didn't share test.env to other processes for this suite,  if enable serial, all case will run on process 1.
I can also remove Serial, and share test.env to other processes, which one do you prefer?

controllers/podplacement/suite_test.go
lwan-wanglin: thanks, this makes code more clear

pkg/testing/image/fake/registry/registry.go
lwan-wanglin: removed lock, only process 1 will create this tmp folder

controllers/operator/clusterpodplacementconfig_controller_test.go
lwan-wanglin: becase we only have one Describe block  and already set all Describe as Ordered for operator test, I share the cluster and remove Serial decorator

controllers/podplacement/suite_test.go
lwan-wanglin: thanks,  will fix as suggestion

pkg/systemconfig/types.go
lwan-wanglin: Once.Do doesn't work, changed to use RWMutex, thanks

pkg/testing/image/fake/registry/registry.go
lwan-wanglin: We directly use the shared variable perRegistryCertDirPath in the file [mock_image.go ](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/testing/image/fake/registry/mock_image.go#L63) , rather than a passing parameter, so need change several functions if we adding it to RegistryConfig struct, what do you think?

pkg/systemconfig/types.go
```diff
@@ -49,30 +50,39 @@ var (
...
 	}
 	return registriesConfPath
 }
 
 func PolicyConfPath() string {
+	rwMutex.RLock()
+	defer rwMutex.RUnlock()
 	if policyConfPath == "" {
 		policyConfPath = lookupEnvOr("POLICY_CONF_PATH", "/etc/containers/policy.json")
 	}
```
aleskandro:
```suggestion
	rwMutex.RLock()
	if policyConfPath != "" {
		defer rwMutex.RUnlock()
	    return policyConfPath
	}
	rwMutex.RUnlock()
	rwMutex.Lock()
	defer rwMutex.Unlock()
	if policyConfPath == "" { // avoid race condition in-between rwMutex.RUnlock and rwMutex.Lock
	  policyConfPath = lookupEnvOr("POLICY_CONF_PATH", "/etc/containers/policy.json")
	}
```

We should write lock when writing.

aleskandro (commented):
/lgtm
/approve

PR #427, opened 2025-01-17
OCPBUGS-45885: ignore pods owned by DaemonSet

lwan-wanglin:
ignore pods owned by a DaemonSet.

### DIFF
--- a/controllers/podplacement/pod_model.go
+++ b/controllers/podplacement/pod_model.go
@@ -332,9 +332,11 @@ func (pod *Pod) hasControlPlaneNodeSelector() bool {
 // - the pod is in the kube-* namespace
 // - the pod has a node name set
 // - the pod has a node selector that matches the control plane nodes
+// - the pod is owned by a daemonset
 func (pod *Pod) shouldIgnorePod() bool {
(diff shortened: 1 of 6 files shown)


controllers/podplacement/pod_model.go
```diff
@@ -397,6 +399,17 @@ func (pod *Pod) isNodeSelectorConfiguredForArchitecture() bool {
 	return true
 }
 
+// isPodFromDaemonSet returns true if the pod is from a daemonSet.
+func (pod *Pod) isFromDaemonSet() bool {
+	// Check all ownerRef
+	for _, ownerRef := range pod.OwnerReferences {
+		if ownerRef.Kind == "DaemonSet" && ownerRef.Controller != nil && *ownerRef.Controller {
```
aleskandro, line 406:
```suggestion
		if ownerRef.Kind == "DaemonSet" && ownerRef.Controller != nil && *ownerRef.Controller {
```

is there a constant we can use in the apimachinery or similar k8s libs?

aleskandro (commented):
Hi @lwan-wanglin, just one comment about the "DaemonSet" string. Other than that, LGTM. Thanks!

controllers/podplacement/pod_model.go
aleskandro: not from a quick look.

PR #433, opened 2025-01-21
fix data race for scheme

lwan-wanglin:
`scheme.Scheme` is a global variable, only configure it once in case occuring dara race.


aleskandro (commented):
/lgtm

aleskandro:
Should we close this?

PR #434, opened 2025-01-21
Revert "changing the test suite run cases concurrently"

lwan-wanglin:
This reverts commit 6c7d21b84be4797fee9efc1186266a40b9a12c6f.
above pr involve a data race when deal with scheme.Scheme, revert the code to unblock the testing, will create new pr after the issue fix


aleskandro:
/override "Red Hat Konflux / multiarch-tuning-operator-on-pull-request"

PR #450, opened 2025-02-04
Update tekton by moving tasks to separate file

AnnaZivkovic:
Update tekton files to match [olm-operator-konflux-sample](https://github.com/konflux-ci/olm-operator-konflux-sample/tree/main/.tekton) 

https://github.com/konflux-ci/olm-operator-konflux-sample/blob/main/.tekton/multi-arch-build-pipeline.yaml#L79-L126


.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
```diff
@@ -29,6 +29,12 @@ spec:
       value: .
     - name: revision
       value: '{{revision}}'
+    - name: build-platforms
+      value:
+        - localhost
```
aleskandro, line 34: this sounds so unsafe to me
cc @jeffdyoung it would be nice if konflux makes it always explicit which architecture we are requesting instead of relying on the assumption that 'localhost' (konflux cluster) is amd64.

.tekton/multiarch-tuning-operator-bundle-push.yaml
```diff
@@ -26,6 +26,12 @@ spec:
       value: .
     - name: revision
       value: '{{revision}}'
+    - name: build-platforms
```
aleskandro, line 29: the bundle should be built as single-arch

.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
aleskandro: (unrelated to the PR)

this sounds unsafe for a future where a Konflux cluster might run on another arch than the (correctly, but also just currently) assumed amd64 or even in the multi-arch compute possibility.

cc @jeffdyoung @arewm  it would be nice if konflux makes it always explicit which architecture we are requesting instead of relying on the assumption that 'localhost' (konflux cluster) is amd64.

Maybe, the script that processes this array might inspect the cluster where the builds are going to run and check whether the platform is available in the cluster (local build0 or not (delegate the build to the multi-platform manager):

```
node_arches = $(oc get nodes -o jsonpath="{range .items[*]}{.status.nodeInfo.architecture}{'\n'}{end}" | sort | uniq)
for platform in platforms:
  if platform in node_arches:
    build_local(...)
  else:
    build_remote_platform_manager(...)
```

.tekton/multiarch-tuning-oprator-build-pipeline.yaml
```diff
@@ -0,0 +1,433 @@
+apiVersion: tekton.dev/v1
```
aleskandro, line 1: Hi @arewm, what happens when we will have this pipeline in multiple branches handled by the same namespace in the Konflux cluster?

aleskandro: Should we rather keep this Pipeline definition in the internal gitops repository that sets up other objects in our namespace? Or could this pipeline be accessed from a shared namespace owned by Konflux so that we stop maintaining it?

.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
arewm: Yes, that is a limitation to the current implemenation. If you would prefer a targeted localhost, you can use `linux/x86_64`. https://github.com/redhat-appstudio/infra-deployments/blob/main/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml#L10

My desire _was_ to change `linux/amd64` to also be a local host, requiring users to explicitly specify whether they want a VM with something like `linux-mlarge/amd64`. 

That would eventually mean that once we have multiple node pools, we might change `linux/arm64` to be in-cluster, requiring users to explicitly request VMs again.

https://issues.redhat.com/browse/KONFLUX-4073 was the feature that I opened for this which has since been closed in favor of https://issues.redhat.com/browse/KONFLUX-6091.

.tekton/multiarch-tuning-oprator-build-pipeline.yaml
arewm: PAC will use the pipeline in the same branch. If you want to use some common pipeline, a resolver referencing a pipeline by bundle or git reference would likely be better.

Here is an initial documentation topic on it: https://konflux-ci.dev/docs/how-tos/workflows/keep-remote-pipelines-up-to-date/

.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
aleskandro: However, in a more k8s/ocp/nonVM-native way, that desire can be solved by allowing to set the resource requests in the Build Tasks and by configuring the cluster autoscaler with machineset autoscalers that can fit different scenarios. Once a 'large' build is created, the cluster would create a Machine with the proper size to host the build (if needed).

aleskandro: > Yes, that is a limitation to the current implemenation. If you would prefer a targeted localhost, you can use linux/x86_64. https://github.com/redhat-appstudio/infra-deployments/blob/main/components/multi-platform-controller/production/stone-prd-rh01/host-config.yaml#L10

This works and would allow us to delegate you about fixing the equivalence between localhost and amd64 :)

.tekton/multiarch-tuning-oprator-build-pipeline.yaml
aleskandro: Thanks @arewm

.tekton/multiarch-tuning-operator-bundle-pull-request.yaml
Prashanth684: > However, in a more k8s/ocp/nonVM-native way, that desire can be solved by allowing to set the resource requests in the Build Tasks and by configuring the cluster autoscaler with machineset autoscalers

This would still not account for the case where the architecture is not supported by the platform, right?

aleskandro: It doesn't for now... But starting from the available arches is already a good step forward imho

aleskandro:
```suggestion
        - linux/amd64
```

PR #369, opened 2024-11-20
MULTIARCH-5058: Implement the node affinity scoring specs in the ClusterPodPlacementConfig

pkenchap:
Task to add the nodeAffinityScoring for ClusterPodPlacemetnConfig API specs can be updated as in the following to host a skeleton of plugins, currently implementing only the nodeAffinityScoring.

Example: 
```
spec:
  plugins:
    nodeAffinityScoring:
        enabled: true
        platforms:
            - architecture: amd64
              # os: linux # TODO: this structure can support the os field in the future (e.g., for win containers)
              weight: 100
```

PR #452, opened 2025-02-05
MULTIARCH-4974: Cluster wide architecture weighted affinity

AnnaZivkovic:
Implementing cluster wide weights using  `PreferredDuringSchedulingIgnoredDuringExecution ` which stores its items as a list.

In the case where we have predefined PreferredSchedulingTerms. We must preserve the weights and avoid unbalancing them with the new arch weights.  To do so we can  normalize the existing weights using arch weights.

`new_weight = 100 * old_weight/ sum(arch weights)`

For example 
User defined arch weights
- arm64, 50
- amd64, 75

The pod yaml would look like the following if there was a pre existing rule
```
      preferredDuringSchedulingIgnoredDuringExecution:
        - weight: 100 * (15 + 50)/ (50 + 75) =  52
          preference:
            matchExpressions:
              - key: label-1
                operator: In
                values:
                  - key-1
              - key: lable/arch
                operator: In
                values:
                  - arm64
          - weight: 100 * (15 + 75)/ (50 + 75) =  72
          preference:
            matchExpressions:
              - key: label-1
                operator: In
                values:
                  - key-1
              - key: lable/arch
                operator: In
                values:
                  - amd64
        - weight: 100 * 50 / (50 + 75) = 40
          preference:
            matchExpressions:
              - key: label/arch
                operator: In
                values:
                  - arm64
          -weight: 100 * 75 / (50 + 75) = 60
          preference:
            matchExpressions:
              - key: label/arch
                operator: In
                values:
                  -amd64
             
```

This pr is dependent on https://github.com/openshift/multiarch-tuning-operator/pull/369


aleskandro:
@Prashanth684 and I were also discussing about providing two _merging_ strategies with an additional field for the nodeAffinityScoring plugin, `mergingStrategy: append|normalize`.

In `append` mode, we concatenate the list of preferred affinity terms in the CPPC to the ones in the Pod we are reconciling. We let the user be free to set the weights in their pods and the CPPC according to their preference, considering no normalization will take place.

### `normalize` merging strategy

In `normalize` mode, we can (a) combine the two sets with the cartesian product, and (b) append the CPPC preferences as they are too.

Combining the two sets with the cartesian product allows us to scale the preference for a given term to N preferences weighted by both the architecture preference and the one already defined. For example, considering the architecture preferences to be ${(25, arm64), (75, amd64), a pod expressing the preference for a node labeled `disktype=ssd` with weight 1 (with no other terms, a weight of 1 is the same as 100%, which may potentially loose in the case the weights of the architectures are set differently, e.g., by letting their sum be 100) should be scaled to prefer, in order:
1. disktype==ssd + amd64
2. disktype==ssd + arm64
3. amd64 (includes the case no disktype label is present in the node)
4. arm64 (includes the case no disktype label is present in the node)

Changing the previously set preferred affinity term, $$(1, disktype=ssd)$$, can be considered safe as there is no node where the architecture labels are not available (that could let the scheduler skip scoring a node labeled `disktype==ssd`).

Finally, also appending the preferred affinity terms as they are in the CPPC for the architectures will cover the scoring of nodes that have none of the other labels in the pod preferred affinity terms, but the architecture ones. In this way, we don't risk to skip the architecture-aware scoring of nodes that do not match any of the matchExpressions already set in the pod specs' preferred affinity terms.

> In the case where we have predefined PreferredSchedulingTerms. We must preserve the weights and avoid unbalancing them with the new arch weights. To do so we can normalize the existing weights using arch weights.

> new_weight = 100 * old_weight/ sum(arch weights)


I think the normalization should use the sum of all the weights (arch weights and already existing weights) in the denominator.

Each preferred term is a tuple

```math
T = (weight \in \mathbb{N}^+, labelSelectorPredicates)
```

The resulting set of preferred affinity terms should consist of the union of the cartesian product of the pod ones with the CPPC ones and the CPPC ones:

```math
\text{PodPreferredAffinityTerms}_\text{New} = \text{PodPreferredAffinityTerms}_\text{Old} \bigtimes \text{CPPCPreferredAffinityTerms} \bigcup \text{CPPCPreferredAffinityTerms}

\left|\text{PodPreferredAffinityTerms}_\text{New}\right| = \left|\text{PodPreferredAffinityTerms}_\text{Old} \bigtimes \text{CPPCPreferredAffinityTerms} \bigcup \text{CPPCPreferredAffinityTerms}\right| \le \left|\text{PodPreferredAffinityTerms}_\text{Old}\right| \cdot 4 + 4

```

For each preferred affinity term $$T_{old} = (weight_{old}, labelSelectorPredicates_{old})$$ in the pod specifications, we will have N terms $$T_i$$, where N is the number of (architecture) preferred affinity terms in the CPPC (at most 4).

In the new set, $$T_{i} \forall i \in \text{CPPCPreferredAffinityTerms} \subseteq  \\{ amd64, arm64, ppc64le, s390x \\} $$ should consist of:

```math
labelsSelectorPredicates_{new, i} = labelSelectorPredicates_{old} \cup \{ \text{"kubernetes.io/arch"}\in \{ i \} \}
```
```math
weight_{new, i} = 100 \cdot \Bigg\lceil \frac{weight_{old} + weight_{i}}{\sum_{j=1}^{M}{weight_{old,j}} + \sum_{i=1}^{N}{weight_{i}}} \Bigg\rceil
```

where M is the number of preferred affinity terms in the Pod specs: $$M = \left|\text{PodPreferredAffinityTerms}\right|$$

#### Example
```math
\text{CPPCPreferredAffinityTerms} = \{ (25, \{amd64\}), (75, \{arm64\}) \}
```

```math
\text{PodPreferredAffinityTerms} = \{ (10, \{ssd\}) \}
```

```math
\text{PodPreferredAffinityTerms}_{new} = \{(31, \{amd64, ssd\}), (77, \{arm64, ssd\}), (22, \{amd64\}), (68, \{arm64\})\}
```

For that pod, the `normalize` merging strategy will prioritize:
1. ARM64 Node with disktype=ssd
2. ARM64 Node without disktype=ssd
3. AMD64 Node with disktype=ssd
4. AMD64 Node without disktype=ssd

The `concatenate` merging strategy would produce
```math
\text{PodPreferredAffinityTerms}_{new} = \{(25, \{amd64\}), (75, \{arm64\}), (10, \{ssd\}) \}
```
In this case, the scheduler will prioritize:
1. ARM64 Node with disktype=ssd
3. ARM64 Node without disktype=ssd
4. AMD64 Node with disktype=ssd
5. AMD64 Node without disktype=ssd

**Which is the same, though. So I'm not sure yet if we should skip the two strategies (implementing `concatenation` only) or find a better normalization formula.**



@AnnaZivkovic, having this information in a small EP (and doing some more examples to find a logic idea of how we will affect the scheduling of pods with pre-existing affinity terms) in the docs/ folder might be nice.

Prashanth684:
> @AnnaZivkovic, having this information in a small EP (and doing some more examples to find a logic idea of how we will affect the scheduling of pods with pre-existing affinity terms) in the docs/ folder might be nice.

+1 Nice write up @aleskandro . We should definitely get this recorded in an EP and define the two strategies clearly. In future there might also be room to add more strategies like maybe a strategy where an input variable or even a normalization function is provided by user to influence the scheduling.

Prashanth684:
```
Which is the same, though. So I'm not sure yet if we should skip the two strategies (implementing concatenation only) or find a better normalization formula.
```
wait I'm confused..the whole reason we thought about the merging strategy was because the nodes with ssd should be considered over nodes without ssd? or is it just purely based on the higher number?

AnnaZivkovic:
> ```
> Which is the same, though. So I'm not sure yet if we should skip the two strategies (implementing concatenation only) or find a better normalization formula.
> ```
> 
> wait I'm confused..the whole reason we thought about the merging strategy was because the nodes with ssd should be considered over nodes without ssd? or is it just purely based on the higher number?

> When the scheduler finds nodes that meet all the other scheduling requirements of the Pod, the scheduler iterates through every preferred rule that the node satisfies and adds the value of the weight for that expression to a sum. The final sum is added to the score of other priority functions for the node. Nodes with the highest total score are prioritized when the scheduler makes a scheduling decision for the Pod.

https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

So it will prefer the one with the highest score.  We could just append to this list, but we risk  unbalancing  any predefined user rules

main.go
```diff
@@ -113,10 +114,25 @@ func main() {
...
+				// Override to remove any field selector for this resource
+				Field: fields.Everything(),
+			},
+			&multiarchv1alpha1.ClusterPodPlacementConfig{}: {
+				// Override to remove any field selector for this resource
+				Field: fields.Everything(),
+			},
+		}
 	}
+
```
aleskandro:
```suggestion

		// The controllers should watch the pods with status.phase==Pending to update its nodeAffinity.
		// We can discard other pods that are not Pending because they are already scheduled.
		cacheOpts.ByObject = map[client.Object]cache.ByObject{
			&corev1.Pod{}: {
				Field: fields.OneTermEqualSelector("status.phase", "Pending"),
			},
		}
	}

```

What about this?

pkg/testing/framework/pod.go
```diff
@@ -130,6 +130,40 @@ func VerifyPodNodeAffinity(ctx context.Context, client runtimeclient.Client, ns
...
+			}
+
+			expectedNodeAffinity := &v1.NodeAffinity{
+				RequiredDuringSchedulingIgnoredDuringExecution: &v1.NodeSelector{
+					NodeSelectorTerms: requiredTerms,
+				},
+				PreferredDuringSchedulingIgnoredDuringExecution: preferredSchedulingTerms,
+			}
+
+			g.Expect(pod).To(HaveEquivalentNodeAffinity(expectedNodeAffinity),
```
lwan-wanglin: @aleskandro @AnnaZivkovic does it make sense to split [HaveEquivalentNodeAffinity](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/testing/framework/node_affinity_equivalence_matcher.go)  into two, one for requiredDuringSchedulingIgnoredDuringExecution, and another for preferredDuringSchedulingIgnoredDuringExecution,  now if we want to check one of them, we will need to give both.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1128,4 +1131,59 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+						{Architecture: utils.ArchitectureAmd64, Weight: 50},
+					},
+				},
+			}
+
+			err = client.Update(ctx, ppc)
+			Expect(err).NotTo(HaveOccurred(), "failed to update ClusterPodPlacementConfig", err)
+			Expect(ppc.Spec.Plugins).NotTo(BeNil())
+		})
+		It("appends ClusterPodPlacementConfig node affinity to nil preferred affinities", func() {
```
lwan-wanglin: this test is better to put into `pkg/e2e/operator` folder

pkg/testing/framework/pod.go
aleskandro: Yes I think so

controllers/podplacement/pod_reconciler_test.go
```diff
@@ -384,4 +389,226 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 			)
 		})
 	})
+	When("The node affinity scoring plugin is enabled", func() {
```
lwan-wanglin, line 387: hi @AnnaZivkovic is it possible to move  this `"When"` spec to `controllers/operator/clusterpodplacementconfig_controller_test.go` ?  the test cases in `controllers/podplacement/... ` suite are running in parallel,  the CPPC creation in below `BeforeEach(){}` might be run at the same time in different process.

controllers/podplacement/pod_model.go
```diff
@@ -122,13 +122,18 @@ func (pod *Pod) SetNodeAffinityArchRequirement(pullSecretDataList [][]byte) (boo
 		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
 	}
 
-	pod.setArchNodeAffinity(requirement)
+	if pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution == nil {
+		pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = []corev1.PreferredSchedulingTerm{}
+	}
```
aleskandro: Move to line 185 (to perform only if the plugin is enabled).
```suggestion
```

controllers/podplacement/pod_model.go
```diff
@@ -166,6 +171,51 @@ func (pod *Pod) setArchNodeAffinity(requirement corev1.NodeSelectorRequirement)
...
+func (pod *Pod) setPreferredArchNodeAffinity(cache ICache) {
+	cppc := cache.GetClusterPodPlacementConfig()
+	if cppc == nil {
+		return
+	}
+
+	if !cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() && len(cppc.Spec.Plugins.NodeAffinityScoring.Platforms) > 0 {
+		return
+	}
+
```
aleskandro:
```suggestion

	if pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution == nil {
		pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = []corev1.PreferredSchedulingTerm{}
	}
```

xref https://github.com/openshift/multiarch-tuning-operator/pull/452#discussion_r1961154591

controllers/podplacement/pod_model.go
```diff
@@ -166,6 +171,51 @@ func (pod *Pod) setArchNodeAffinity(requirement corev1.NodeSelectorRequirement)
...
+			},
+		}
+		pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(
+			pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, preferredSchedulingTerm)
+	}
+
+	// if the nodeSelectorTerms were patched at least once, we set the nodeAffinity label to the set value, to keep
+	// track of the fact that the nodeAffinity was patched by the operator.
+	pod.ensureLabel(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet)
+	pod.publishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
```
aleskandro: These are the same label and event set for the required node affinity. Can we have dedicated ones for the preferred affinity?

controllers/podplacement/clusterpodplacementconfg_informer.go
```diff
@@ -119,3 +124,49 @@ func (s *CPPCSyncer) onUpdate(ic informers.ICache) func(oldObj, newObj interface
...
+}
+
+func newCache() ICache {
+	c := &clusterPodPlacementConfig{
+		config: nil,
+	}
+	return c
+}
+
+func (c *clusterPodPlacementConfig) StoreClusterPodPlacementConfig(config *multiarchv1beta1.ClusterPodPlacementConfig) error {
```
aleskandro: can we remove the return value if it's always nil?

controllers/podplacement/clusterpodplacementconfg_informer.go
```diff
@@ -119,3 +124,49 @@ func (s *CPPCSyncer) onUpdate(ic informers.ICache) func(oldObj, newObj interface
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.config = config.DeepCopy()
+	return nil
+}
+
+func (c *clusterPodPlacementConfig) DeleteClusterPodPlacementConfig() error {
```
aleskandro: can we remove the return value if it's always nil?

controllers/operator/suite_test.go
```diff
@@ -219,7 +218,7 @@ func runManager() {
 	By("Setting up Cluster Podplacement Config informer")
 	err = mgr.Add(podplacement.NewCPPCSyncer(mgr))
 	Expect(err).NotTo(HaveOccurred())
-	ic := informers.CacheSingleton()
+	ic := podplacement.CacheSingleton()
```
aleskandro: Is the operator actually using this cache?

controllers/podplacement/clusterpodplacementconfg_informer.go
```diff
@@ -119,3 +124,49 @@ func (s *CPPCSyncer) onUpdate(ic informers.ICache) func(oldObj, newObj interface
 		s.onAdd(ic)(newobj)
 	}
 }
+
+type ICache interface {
```
aleskandro: Now that we've moved this code here, I wonder if we could improve its organization further.

- The `ICache` interface and the `clusterPodPlacementConfig` struct seem unnecessary as-is. The `(Store|Get|Delete)ClusterPodPlacementConfig` functions don’t need to be struct methods. Instead of wrapping `ClusterPodPlacementConfig` in a struct with a mutex, we could use private package-level variables. Only `GetClusterPodPlacementConfig` should remain exposed.  
- The `CPPCSyncer` might eventually be useful beyond the pod placement controller. Moving it, along with related logic, to `pkg/informers/clusterpodplacementconfig` would make it more reusable.  

I've not digged into it yet but, a possible proposal can be:

#### `pkg/informers/clusterpodplacementconfig`  
- `GetClusterPodPlacementConfig()` → Calls `internal.GetClusterPodPlacementConfig()`
- Exposed `CPPCSyncer` and its methods

#### `pkg/informers/clusterpodplacementconfig/internal`  
- Private `clusterPodPlacementConfig *multiarchv1beta1.ClusterPodPlacementConfig`, accessible only within this package  
- `(Store|Delete)ClusterPodPlacementConfig()`: Ensures only the `clusterpodplacementconfig` package modifies the cache  
- `GetClusterPodPlacementConfig()` → Returns `clusterPodPlacementConfig`  
- Private `mu *sync.RWMutex`


We would earn in terms of:
- Encapsulation: The cache only exposes `GetClusterPodPlacementConfig()` to other entities than the CPPCSyncer. The CPPCsyncer struct is the only other exposed entity, for use by the main.go or test suites.
- Reusability: By moving this out of `controllers/podplacement`, any controller can use it.
- Simplification: Eliminates unnecessary abstractions (`ICache`), making the implementation cleaner.  

Also see [Go Modules Layout](https://go.dev/doc/modules/layout#package-or-command-with-supporting-packages)

controllers/podplacement/pod_reconciler_test.go
aleskandro: I think the test belongs here, Tori has now removed the creation of the CPPC and set the plugin enabled by default in this suite's creation request for it.

controllers/operator/suite_test.go
AnnaZivkovic: Yes we have integration tests checking that the cache is working

AnnaZivkovic: @aleskandro Does this look better?

controllers/podplacement/clusterpodplacementconfg_informer.go
AnnaZivkovic: @aleskandro  Does this look better?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1128,4 +1134,103 @@ var _ = Describe("The Pod Placement Operand", func() {
 			Eventually(framework.VerifyPodNodeAffinity(ctx, client, ns, "app", "test"), e2e.WaitShort).Should(Succeed())
 		})
 	})
+	Context("with different types of preferred affinities", func() {
+		It("does not append any ClusterPodPlacementConfig node affinity", func() {
```
lwan-wanglin: you have already created CPPC with NodeAffinityScoring plugin enabled in the [pkg/e2e/podplacement/e2e_test.go](https://github.com/openshift/multiarch-tuning-operator/pull/452/files#diff-208446009960023d12f0d8faab7579cf4eab1626d23905936dd4891bac781a29R67-R78), so this test case will always failed

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1128,4 +1134,103 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+
+			archLabelNSR := NewNodeSelectorRequirement().
+				WithKeyAndValues(utils.ArchLabel, corev1.NodeSelectorOpIn, utils.ArchitectureAmd64).Build()
+			expectedNSTs := NewNodeSelectorTerm().WithMatchExpressions(archLabelNSR).Build()
+			preferredSchedulingTerm := NewPreferredSchedulingTerm().WithKeyAndValues(50, *expectedNSTs).Build()
+
+			By("The pod should have been processed by the webhook and the scheduling gate label should be added")
+			Eventually(framework.VerifyPodLabels(ctx, client, ns, "app", "test", e2e.Present, schedulingGateLabel), e2e.WaitShort).Should(Succeed())
+
+			By("The pod should have been set node affinity of arch info.")
```
lwan-wanglin:
```suggestion
			By("The pod should have been set preferred node affinity of arch info.")
```

pkg/testing/builder/pod.go
```diff
@@ -105,9 +105,30 @@ func (p *PodBuilder) WithRequiredDuringSchedulingIgnoredDuringExecution() *PodBu
...
+		term := v1.PreferredSchedulingTerm{}
+		opt(&term)
+		p.pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(
+			p.pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution,
+			term,
+		)
+	}
+
+	return p
+}
```
lwan-wanglin:
```suggestion
func (p *PodBuilder) WithPreferredDuringSchedulingIgnoredDuringExecution(values ...*v1.PreferredSchedulingTerm) *PodBuilder {
	p.WithNodeAffinity()

	if p.pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution == nil {
		p.pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = []v1.PreferredSchedulingTerm{}
	}

	for i := range values {
		if values[i] == nil {
			panic("nil value passed to WithPreferredDuringSchedulingIgnoredDuringExecution")
		}
		p.pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = append(
			p.pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution, *values[i])
	}
	return p
}
```
can this func just pass `*v1.PreferredSchedulingTerm` directly instead of `func(*v1.PreferredSchedulingTerm)`?

controllers/operator/suite_test.go
aleskandro: Those tests should stay in the podplacement as the operator doesn't seem to use this informer at all... Testing should happen in the podplacement test suite to ensure we have everything in place for that informer to run in the podplacement controller pods (eg. the RBAC rules to access ClusterPodPlacementConfigs).

aleskandro: There are a few verifications there: before creation, after creation, after update, after delete. We can spread that case in the Before and After suite nodes of the podplacement test suite.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1128,4 +1134,111 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+			}), ppc)
+			Expect(err).NotTo(HaveOccurred(), "failed to get the v1beta1 ClusterPodPlacementConfig", err)
+			Eventually(framework.ValidateCreation(client, ctx)).Should(Succeed())
+
+			// set the plugins to nil
+			if ppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() {
+				ppc.Spec.Plugins.NodeAffinityScoring.Enabled = false
+			}
+
+			err = client.Update(ctx, ppc)
```
aleskandro: this cannot run in parallel with other cases. As I see there are still problems with others, can we comment this entire case so that we work with one for now?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1128,4 +1134,111 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+					Plugins: &plugins.Plugins{
+						NodeAffinityScoring: &plugins.NodeAffinityScoring{
+							BasePlugin: plugins.BasePlugin{
+								Enabled: true,
+							},
+							Platforms: []plugins.NodeAffinityScoringPlatformTerm{
+								{Architecture: utils.ArchitectureAmd64, Weight: 50},
+							},
+						},
+					}}}), ppc)
```
aleskandro: No need for all the Spec stanza to retrieve the CPPC.

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -95,6 +95,12 @@ var _ = Describe("The Pod Placement Operand", func() {
 			), e2e.WaitShort).Should(Succeed())
 			By("The pod should have been set node affinity of arch info.")
 			Eventually(framework.VerifyPodNodeAffinity(ctx, client, ns, "app", "test", *expectedNSTs), e2e.WaitShort).Should(Succeed())
+			archLabelNSR = NewNodeSelectorRequirement().
```
```suggestion
			By("The pod should have the expected preferred affinities")
			archLabelNSR = NewNodeSelectorRequirement().
```

aleskandro: And so on where you will verify that

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -244,6 +260,8 @@ var _ = Describe("The Pod Placement Operand", func() {
 			Eventually(framework.VerifyPodLabels(ctx, client, ns, "app", "test", e2e.Present, schedulingGateNotSetLabel), e2e.WaitShort).Should(Succeed())
```
aleskandro, line 262: I think the pod is ignored by the webhook when the required node affinity for architecture is already set. Therefore, the scheduling gate is not set and the preferred node affinity neither.

As we are now having this plugin that sets the preferred node affinity, we should change the behavior in the webhook that let it skip setting the scheduling gate in the pod:
https://github.com/openshift/multiarch-tuning-operator/blob/f9402c7673e2192762ea5acca31b6dbd8fff4c8a/controllers/podplacement/pod_model.go#L329-L341

aleskandro: I think pod.isNodeSelectorConfiguredForArchitecture() should change to

```
pod.isNodeSelectorConfiguredForArchitecture() && !cppc.plugins.nodeAffinity.enabled ||
pod.isNodeSelectorConfiguredForArchitecture() && cppc.Plugins.nodeAffinity.enabled && pod.isPreferredAffinityConfiguredForArchitecture()
```

where `isPreferredAffinityConfiguredForArchitecture()` has to be implemented and `cppc` comes from the informer cache. The webhook runs in the pod placement controller, so it should already be available.

aleskandro: What I would do is to Change this test case such that the deployment has also the preferred affinity set.

After the regression cases are set and succeeded, we will need two other testcases in which the user provides either the required node affinity (only required node affinity is set) or the preferred node affinity (only preferred node affinity is set).

controllers/podplacement/pod_model.go
```diff
@@ -429,3 +492,26 @@ func (pod *Pod) handleError(err error, s string) {
...
+		return false
+	}
+
+	if pod.Spec.Affinity.NodeAffinity == nil {
+		return false
+	}
+
+	if pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution == nil {
+		return false
+	}
```
aleskandro: Use [short-circuit](https://kuree.gitbooks.io/the-go-programming-language-report/content/22/text.html):
```suggestion
	if pod.Spec.Affinity == nil || pod.Spec.Affinity.NodeAffinity == nil || 
		pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution == nil {
		return false
	}
```

controllers/podplacement/pod_model.go
```diff
@@ -327,16 +367,39 @@ func (pod *Pod) hasControlPlaneNodeSelector() bool {
...
+
+	// If NodeAffinityScoring is enabled, check preferred affinity and node selector
+	if !pod.isPreferredAffinityConfiguredForArchitecture() && isNodeSelectorConfigured {
+		return false
+	}
+
+	// Otherwise, do not ignore the pod
+	return true
 }
```
aleskandro: Ref https://github.com/openshift/multiarch-tuning-operator/pull/452#discussion_r1971487463

```suggestion
// The operator should ignore the pod in the following cases:
// - The pod is in the same namespace as the operator.
// - The pod is in a kube-* namespace.
// - The pod has a node name set (indicating it's already assigned).
// - The pod has a node selector that matches control plane nodes
// - The pod is owned by a DaemonSet.
// - Both the nodeSelector and the preferredAffinity are set for the kubernetes.io/arch label.
// - Only the nodeSelector is set for the kubernetes.io/arch label and the NodeAffinityScoring plugin is disabled.
func (pod *Pod) shouldIgnorePod(cppc *v1beta1.ClusterPodPlacementConfig) bool {
	return utils.Namespace() == pod.Namespace || strings.HasPrefix(pod.Namespace, "kube-") ||
		pod.Spec.NodeName != "" || pod.hasControlPlaneNodeSelector() || pod.isFromDaemonSet() ||
		pod.isNodeSelectorConfiguredForArchitecture() && (pod.isPreferredAffinityConfiguredForArchitecture() ||
			cppc.Spec.Plugins == nil || !cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled())
}
```

main.go
```diff
@@ -230,6 +230,15 @@ func RunClusterPodPlacementConfigOperandControllers(mgr ctrl.Manager) {
...
 
+	if enableCPPCInformer {
+		cppcSyncer := clusterpodplacementconfig.NewCPPCSyncer(mgr)
+		if err := mgr.Add(cppcSyncer); err != nil {
+			setupLog.Error(err, "unable to add CPPCSyncer")
+			os.Exit(1)
+		}
+		setupLog.Info("CPPCSyncer is enabled")
+	}
+
```
aleskandro: Can we move this before line 178 and avoid duplicating?

https://github.com/openshift/multiarch-tuning-operator/pull/452/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261L176-L178

AnnaZivkovic: When i tried this change the CPPC was not enabled

controllers/podplacement/pod_model.go
```diff
@@ -100,6 +101,9 @@ func (pod *Pod) RemoveSchedulingGate() {
 // Then, it computes the intersection of the architectures supported by the images used by the pod via pod.getArchitecturePredicate.
 // Finally, it initializes the nodeAffinity for the pod and set it to the computed requirement via the pod.setArchNodeAffinity method.
 func (pod *Pod) SetNodeAffinityArchRequirement(pullSecretDataList [][]byte) (bool, error) {
+	if pod.isNodeSelectorConfiguredForArchitecture() {
```
lwan-wanglin, line 104: if we can check `pod.isRequiredAffinityConfiguredForArchitecture()` here?

controllers/podplacement/pod_reconciler_test.go
```diff
@@ -384,4 +384,183 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
...
+		Context("with different types of preferred affinities", func() {
+			It("appends ClusterPodPlacementConfig node affinity to nil preferred affinities", func() {
```
lwan-wanglin, line 389:
```suggestion
			It("appends ClusterPodPlacementConfig node affinity to the predefined preferred affinities", func() {
```
isn't this test case for appending the predefined preferred affinities?

controllers/podplacement/pod_model.go
aleskandro: restored the previous method... Ok to refactor, but this PR has already many changes.

controllers/podplacement/pod_reconciler_test.go
AnnaZivkovic: That should be the case where PreferredDuringSchedulingIgnoredDuringExecution = [] and we just append to it. On line 436 is where we should have the predefined PreferredDuringSchedulingIgnoredDuringExecution

controllers/podplacement/pod_reconciler.go
```diff
@@ -115,6 +115,11 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 		pod.publishEvent(corev1.EventTypeWarning, ArchitectureAwareGatedPodIgnored, ArchitectureAwareGatedPodIgnoredMsg)
 		return
 	}
+
+	if cppc != nil && cppc.Spec.Plugins != nil && cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() {
+		pod.SetPreferredArchNodeAffinity(cppc)
```
Prashanth684, line 123: if cppc = nil, the controller wouldn't even be created right? is this just a safety check?

controllers/podplacement/pod_model.go
```diff
@@ -166,6 +166,52 @@ func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequi
...
+		pod.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution = []corev1.PreferredSchedulingTerm{}
+	}
+
+	for _, nodeAffinityScoringPlatformTerm := range cppc.Spec.Plugins.NodeAffinityScoring.Platforms {
+		preferredSchedulingTerm := corev1.PreferredSchedulingTerm{
+			Weight: nodeAffinityScoringPlatformTerm.Weight,
+			Preference: corev1.NodeSelectorTerm{
+				MatchExpressions: []corev1.NodeSelectorRequirement{
+					{
+						Key:      utils.ArchLabel,
```
Prashanth684, line 203: do we do sanity checking on duplicate entries?it's probably not needed because the scheduler can handle it - (it'll probably aggregate the duplicate weights so that architecture will be massively favored), but I was wondering if we are a smart scheduling influencer platform whether we should start doing some sanity checks

docs/cluter_wide_weights.md
```diff
@@ -0,0 +1,149 @@
+# Considerations for Cluster-wide Weights
+# This needs more work
```
Prashanth684: this doc appears to be something that needs to be in an enhancement, not necessarily user facing docs.

controllers/podplacement/pod_reconciler.go
aleskandro: yes, but i think we should keep this here (and i also think the linter tests should fail if we don't)

controllers/podplacement/pod_model.go
aleskandro: yes @pkenchap implemented this already.

docs/cluter_wide_weights.md
aleskandro: yep. I dropped this commit for now, we already discussed the past Tuesday about having it in a separate PR when Tori's back.

PR #453, opened 2025-02-05
MULTIARCH-4252: namespace-scoped Pod Placement Config

aleskandro:
This PR provides the design proposal for the namespace-scoped Pod Placement Configuration and minor fixes to the (to be updated) initial EP.

### DIFF
--- a/docs/enhancements/MTO-0002-local-pod-placement.md
+++ b/docs/enhancements/MTO-0002-local-pod-placement.md
@@ -1,0 +1,348 @@
+---
+title: Introducing the namespace-scoped `PodPlacementConfig`
+authors:
+  - "@aleskandro"
+reviewers:
+  - "@AnnaZivkovic"
+  - "@jeffdyoung"
+  - "@lwan-wanglin"
+  - "@Prashanth684"
+  - "@prb112"
+approvers:
+  - "@Prashanth684"
+creation-date: 2025-02-04
+last-updated: 2025-02-04
+tracking-link:
+  - https://issues.redhat.com/browse/MULTIARCH-4252
+see-also: []
+---
+
+# Introducing the namespace-scoped `PodPlacementConfig`
+
+## Summary
+Currently, the Multiarch Tuning Operator exposes a cluster-scoped Custom Resource Definition (CRD) that allows users to
+(a) enable the pod placement controller, (b) define a namespace selector to filter the namespaces whose pods should be
+patched with the scheduling gate at admission time, and (c) have all the matching pods in the selected namespaces patched
+with the node selector term for the compatible architectures to influence the pod scheduling's strong predicates run by the 
+node affinity filtering plugin.
+In [MULTIARCH-4970](https://issues.redhat.com/browse/MULTIARCH-4970), we introduced new fields in the `PodPlacementConfig` that
+allow users to define cluster-wide preferences to influence the node affinity scoring plugin with respect to the pod's
+preferred architectures. However, users may want to have more granular control over the pod placement controller's behavior
+by defining different preferences for different namespaces and sub-set of pods within a namespace.
+This enhancement proposes to introduce a namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement
+controller's behavior at a namespace level. The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined 
+in the current version of the cluster-scoped `PodPlacementConfig` CRD, and label selectors to refine the scope of the configuration
+to subset of pods within a namespace matching the label selector.
+
+
+## Motivation
+- Users want to have more granular control over the pod placement controller's behavior by defining different preferences 
+for different namespaces and a subset of pods within a namespace, given they are aware some pods can perform better on certain 
+architectures or support cost-reduction strategies by running pods on cheaper nodes such as arm64 ones.
+
+### User Stories
+- As a cluster administrator, I want to define different preferences for different namespaces and a subset of pods within a namespace
+so that I can influence the node affinity scoring plugin with respect to the pod's preferred architectures at a namespace level.
+
+### Goals
+- Provide a namespace-scoped resource to define the pod placement controller's behavior at a namespace level for the node affinity 
+  scoring plugin
+- Provide a namespace-scoped resource to define the pod placement controller's behavior for the node affinity 
+  scoring plugin on a subset of pods within a namespace
+
+### Non-Goals
+- Allow users to define multiple pod placement configurations that tune the placement of pods across multiple namespaces
+- Allow users to define which namespaces should get the node affinity filtering plugin's behavior managed by the cluster-scoped 
+  `ClusterPodPlacementConfig` CRD: the cluster-scoped `PodPlacementConfig` CRD will continue to be used for this purpose and will be
+  a dependency for the namespace-scoped `PodPlacementConfig` CRD
+
+## Proposal
+
+We introduce a new namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement controller's 
+behavior at a namespace level.
+
+Its development is expected to ship in the following phases:
+
+- Phase 1 - initial introduction of the namespace-scoped `PodPlacementConfig` CRD:
+	- The multiarch tuning operator will refuse the creation of the namespace-scoped `PodPlacementConfig` CRD if the cluster-scoped
+	  `ClusterPodPlacementConfig` CRD is not present
+	- The pod placement controller deployed when the `ClusterPodPlacementConfig` CRD is present will be able to also apply 
+      the configuration to the pods in the selected namespaces, if any namespace-scoped `PodPlacementConfig` CRD is present.
+    - The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined in the current version of the 
+	  cluster-scoped `ClusterPodPlacementConfig` CRD, and label selectors to refine the scope of the configuration to subset of pods within 
+	  a namespace matching the label selector.
+    - The namespace-scoped `PodPlacementConfig` CRD will take precedence over the cluster-scoped `ClusterPodPlacementConfig` CRD when 
+	  both configure the same plugins for the same pods.
+- Phase 2 - Sharding the pod placement controller into per-namespace deployments:
+	- When at least one namespace-scoped `PodPlacementConfig` CRD is present in a given namespace, the multiarch tuning operator
+      deploys a local pod placement controller in the namespace that will make use of an additional scheduling gate to patch
+      the pods specifications with the node affinity terms defined by the enabled plugins in the namespace-scoped `PodPlacementConfig` CRD.
+
+### Phase 1
+
+#### Namespace-scoped `PodPlacementConfig` CRD
+
+The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined in the current version of the
+cluster-scoped `ClusterPodPlacementConfig` CRD, and label selectors to refine the scope of the configuration to subset of pods within
+the namespace.
+
+The following `yaml` snippet shows the structure of the namespace-scoped `PodPlacementConfig` CRD:
+
+```yaml
+apiVersion: multiarch.openshift.io/v1beta1
+kind: `PodPlacementConfig`
+metadata:
+  name: my-namespace-config
+  namespace: my-namespace
+spec:
+  labelSelector:
+      matchExpressions:
+        - key: app
+          operator: In
+          values:
+            - my-label-for-apps-performing-better-on-arm64
+  priority: 100
+  plugins:
+    nodeAffinityScoring:
+      enabled: true
+      platforms:
+        - architecture: amd64
+          weight: 25
+        - architecture: arm64
+          weight: 75
+```
+
+The `plugins.nodeAffinityScoring` fields are defined as in the current version of the cluster-scoped `ClusterPodPlacementConfig` CRD,
+implemented at https://github.com/openshift/multiarch-tuning-operator/pull/369.
+
+The `priority` field is an integer that defines the priority of the namespace-scoped `PodPlacementConfig` CRD. The higher the value, the higher the priority.
+If multiple namespace-scoped `PodPlacementConfig` objects are present in a namespace, the one with the highest priority will take precedence over the others when
+a pod matches the label selector of multiple `PodPlacementConfig` objects.
+
+#### Changes to the Multiarch Tuning Operator
+
+A new validating webhook will be added to the multiarch tuning operator to validate the namespace-scoped `PodPlacementConfig`.
+In particular, the webhook will deny the creation of the namespace-scoped `PodPlacementConfig` if the cluster-scoped 
+`ClusterPodPlacementConfig` CRD is not present.
+This validating webhook will also be the base for the multi-version support of the `PodPlacementConfig` CRD.
+The `ClusterPodPlacementConfig` validating webhook also has to deny the deletion of the `ClusterPodPlacementConfig` object if there are any `PodPlacementConfig` objects in the cluster.
+The webhook is deployed as part of the multiarch tuning operator deployment, similarly to how the current version of the operator 
+deploys the validating webhook for the cluster-scoped `ClusterPodPlacementConfig`.
+
+#### Changes to the Pod Placement Controller
+
+In order to apply the configurations defined by the namespace-scoped `PodPlacementConfig` CRD, the pod placement controller will be changed such that it applies any
+matching namespace-scoped `PodPlacementConfig` CRD to the pods in the selected namespaces, sorted by descending priority, before applying the cluster-scoped
+`ClusterPodPlacementConfig` configuration.
+
+Given a pod hitting the pod reconciliation loop:
+1. Verify the pod has the scheduling gate (return if not)
+2. Get all the pod placement configurations in the pod's namespace, sorted by descending priority
+3. For each pod placement configuration:
+	- Check if the pod matches the label selector
+	- Apply the configuration to the pod.
+    - If a proposed configuration overlaps with the current pod specification, it is not applied and an event
+      is recorded in the pod's event log. This ensures that the operator does not overwrite the user's configuration and 
+      that a `PodPlacementConfig` CRD with a higher priority gets precedence over the ones with lower priority.
+4. Apply the cluster-scoped `ClusterPodPlacementConfig` configuration to the pod. As the namespace-scoped configuration 
+   has already been applied, the cluster-scoped configuration will not overwrite any overlapping namespace-scoped configuration.
+   We can also garantee that the cluster-scoped `ClusterPodPlacementConfig`'s strong predicates will be applied to the pods as they
+   are not handled by the namespace-scoped resource, which is focused on the scheduling scoring plugins.
+
+#### Changes to the Mutating Webhook
+
+No changes are required to the mutating webhook, as the namespace-scoped `PodPlacementConfig` CRD depends on the cluster-scoped `ClusterPodPlacementConfig` CRD and its business logic is implemented by the same controller responsible for reconciling through its scheduling gate.
+
+#### RBAC
+
+We will introduce rules to allow the
+multiarch tuning operator and pod placement controller to, respectively, reconcile (CRUD) and read 
+the namespace-scoped `PodPlacementConfig` CRD.
+
+#### Considerations about uninstallation
+
+No changes in the current process are required. The users are still recommended to uninstall the Multiarch Tuning Operator resources before proceeding with the uninstallation: in the case such resources are not deleted before the operator is uninstalled, the controllers will continue to be able to reconcile the pods, but changes (creation and deletion in particular) will not guarantee the expected behavior.
+
+### Phase 2: Sharding the pod placement controller into per-namespace instances
+
+In phase 2, we will partition the pod placement controller into per-namespace instances and a cluster-wide one that is mainly responsible for the required affinity. When at least one namespace-scoped `PodPlacementConfig` CRD is present in a given namespace, the multiarch tuning operator will deploy a local pod placement controller and mutating webhook in the namespace that will make use of an additional scheduling gate (`multiarch.openshift.io/local-scheduling-gate`) to patch the pods specifications with the node affinity terms defined by the enabled plugins in the namespace-scoped `PodPlacementConfig` CRD.
+
+#### Changes to the Multiarch Tuning Operator
+
+A reconciler for the namespace-scoped `PodPlacementConfig` CRD will be added to the multiarch tuning operator.
+
+When the first `PodPlacementConfig` is created in a namespace, the multiarch tuning operator will deploy in that namespace:
+- a pod placement controller
+- a mutating webhook
+- a mutating webhook configuration
+- service accounts and RBAC rules to allow the pod placement controller to reconcile the pods and the mutating webhook to mutate the pods
+
+When the last `PodPlacementConfig` is deleted in a namespace, the multiarch tuning operator will delete the resources deployed in that namespace.
+
+The Log Verbosity level of the pod placement controller will be set according to the cluster-scoped `ClusterPodPlacementConfig` CRD.
+
+#### Changes to the Pod Placement Controller
+
+The current pod reconciler will have some dedicated flags to allow the pod placement controller to run in a per-namespace mode. The pod placement controller will be able to run in a per-namespace mode by setting the `--namespace` flag to the namespace where the controller is running. The controller will then only reconcile the pods in the namespace where it is running.
+
+The cluster-wide pod placement controller will not process a pod and remove the related scheduling gate (`multiarch.openshift.io/scheduling-gate`) until the local `multiarch.openshift.io/local-scheduling-gate` scheduling gate is removed.
+
+The behavior of the pod placement controller will be the same as in the first phase, but split into two different instances: one for the namespace-scoped configuration, applied first and considering the available `PodPlacementConfig` sorted by descending priority, and the other for the cluster-scoped configuration, applied after the namespace-scoped configuration.
+
+#### Changes to the Mutating Webhook
+
+The Mutating Webhook will be deployed in the namespace where the pod placement controller is running. The Mutating Webhook will be configured to mutate the pods in the namespace where it is running and will add a different scheduling gate to the pods (`multiarch.openshift.io/local-scheduling-gate`).
+
+The Mutating Webhook Configuration will be responsible for instructing the kube-apiserver to call the Mutating Webhook in the namespace that has a `PodPlacementConfig` for the pods created in that namespace.
+Each `PodPlacementConfig` will have a Mutating Webhook Configuration rule with the object selector to match the pods in the namespace where the `PodPlacementConfig` is created.
+
+#### Consideration about uninstallation
+
+When the last `PodPlacementConfig` is deleted in a namespace, the multiarch tuning operator will delete the resources deployed in that namespace.
+The users are still recommended to uninstall the Multiarch Tuning Operator resources before proceeding with the uninstallation: in the case such resources are not deleted before the operator is uninstalled, the controllers will continue to be able to reconcile the pods, but changes (creation and deletion in particular) are not guaranteed to take effect as expected.
+
+#### RBAC
+
+All the RBAC rules for the local pod placement controller and the local mutating webhook will be created by the multiarch tuning operator when the first `PodPlacementConfig` is created in a namespace. Such rules will include `Role`, `RoleBinding` and `ServiceAccount` resources. No cluster-scoped RBAC rules should be required for the local pod placement controller and the local mutating webhook.
+
+#### Summary of the rules
+##### Pod Placement Controller
+
+---------------
+
+| Resource                                         | Methods                         | comments                                                                                                            |
+|--------------------------------------------------|---------------------------------|---------------------------------------------------------------------------------------------------------------------|
+| pods                                             | get, list, watch, patch, update |                                                                                                                     | 
+| pods/status                                      | update                          |                                                                                                                     |
+| security.openshift.io/securitycontextconstraints | use                             | it's used in case the pods are created by a user/SA that needs some privileges                                      |
+| events                                           | create, patch                   |                                                                                                                     |
+| podplacementconfigs.multiarch.openshift.io       | get, list, watch                |                                                                                                                     |
+| authorization.openshift.io/subjectaccessreviews  | create                          | automatically created by kubebuilder (required for leader election)                                                 |
+| authentication.k8s.io/tokenreviews               | create                          | automatically created by kubebuilder (required for leader election)                                                 |
+| configmaps, coordination.k8s.io/leases           | create, get, update, delete     | automatically created by kubebuilder (required for leader election)                                                 |
+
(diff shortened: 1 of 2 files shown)


docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+patched with the scheduling gate at admission time, and (c) have all the matching pods in the selected namespaces patched
+with the node selector term for the compatible architectures to influence the pod scheduling's strong predicates run by the 
+node affinity filtering plugin.
+In [MULTIARCH-4970](https://issues.redhat.com/browse/MULTIARCH-4970), we introduced new fields in the `PodPlacementConfig` that
+allow users to define cluster-wide preferences to influence the node affinity scoring plugin with respect to the pod's
+preferred architectures. However, users may want to have more granular control over the pod placement controller's behavior
+by defining different preferences for different namespaces and sub-set of pods within a namespace.
+This enhancement proposes to introduce a namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement
+controller's behavior at a namespace level. The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined 
+in the current version of the cluster-scoped `PodPlacementConfig` CRD, and label selectors to refine the scope of the configuration
```
Prashanth684, line 34: nit: might want to call these out as concise bullet points to list out what the refinements are.

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+so that I can influence the node affinity scoring plugin with respect to the pod's preferred architectures at a namespace level.
+
+### Goals
+- Provide a namespace-scoped resource to define the pod placement controller's behavior at a namespace level for the node affinity 
+  scoring plugin
+- Provide a namespace-scoped resource to define the pod placement controller's behavior for the node affinity 
+  scoring plugin on a subset of pods within a namespace
+
+### Non-Goals
+- Allow users to define the pod placement controller's behavior at a pod level, selecting pods across multiple namespaces
```
Prashanth684: instead of pod level - should it be "app" level across namespaces ?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+- Allow users to define which namespaces should get the node affinity filtering plugin's behavior managed by the cluster-scoped 
+  `ClusterPodPlacementConfig` CRD: the cluster-scoped `PodPlacementConfig` CRD will continue to be used for this purpose and will be
+  a dependency for the namespace-scoped `PodPlacementConfig` CRD
```
Prashanth684, line 57: why should it be a dependency? why can't they operate independently?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+
+## Proposal
+
+We introduce a new namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement controller's 
+behavior at a namespace level.
+
+Its development is expected to ship in the following phases:
+
+- Phase 1 - initial introduction of the namespace-scoped `PodPlacementConfig` CRD:
+	- The multiarch tuning operator will refuse the creation of the namespace-scoped `PodPlacementConfig` CRD if the cluster-scoped
```
Prashanth684, line 67: Why can't the namespace podplacementconfig encompass what the clusterpodplacementconfig does?we can evaluate both if present and make sure the namespace configs get preference over the cluster wide config

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+        - architecture: amd64
+          weight: 25
+        - architecture: arm64
+          weight: 75
+```
+
+The `plugins.nodeAffinityScoring` fields are defined as in the current version of the cluster-scoped `ClusterPodPlacementConfig` CRD,
+implemented at https://github.com/openshift/multiarch-tuning-operator/pull/369.
+
+The `priority` field is an integer that defines the priority of the namespace-scoped `PodPlacementConfig` CRD. The higher the value, the higher the priority.
```
Prashanth684, line 118: does priority have a range?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+	  `ClusterPodPlacementConfig` CRD is not present
+	- The pod placement controller deployed when the `ClusterPodPlacementConfig` CRD is present will be able to also apply 
+      the configuration to the pods in the selected namespaces, if any namespace-scoped `PodPlacementConfig` CRD is present.
+    - The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined in the current version of the 
+	  cluster-scoped `ClusterPodPlacementConfig` CRD, and label selectors to refine the scope of the configuration to subset of pods within 
+	  a namespace matching the label selector.
+    - The namespace-scoped `PodPlacementConfig` CRD will take precedence over the cluster-scoped `ClusterPodPlacementConfig` CRD when 
+	  both configure the same plugins for the same pods.
+- Phase 2 - Sharding the pod placement controller into per-namespace deployments:
+	- When at least one namespace-scoped `PodPlacementConfig` CRD is present in a given namespace, the multiarch tuning operator
```
Prashanth684, line 77: we will also make the pod unecessarily go through 2 scheduling gates if we mandate that a cluster wide podplacementconfig is mandatory

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,296 @@
...
+
+We introduce a new namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement controller's 
+behavior at a namespace level.
+
+Its development is expected to ship in the following phases:
+
+- Phase 1 - initial introduction of the namespace-scoped `PodPlacementConfig` CRD:
+	- The multiarch tuning operator will refuse the creation of the namespace-scoped `PodPlacementConfig` CRD if the cluster-scoped
+	  `ClusterPodPlacementConfig` CRD is not present
+	- The pod placement controller deployed when the `ClusterPodPlacementConfig` CRD is present will be able to also apply
```
Prashanth684, line 69: so are we saying that in phase 1 - the same controller(used for cluster wide ppc) is used for namespace scoped ppc as well? is that why we have the dependence?

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: We want the cluster pod placement object to handle the required affinity. Users can use namespace selectors to select or restrict the namespaces the required affinity is reconciled on, but that work requires secrets that the service account in the user namespace should be allowed to access and that's something we would rather want to be widely enabled in clusters as more critical and a prelude to touning the scoring plugins.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: In the first phase, having it as a dependency will allow us to reduce the development effort in the operator as the operator will not have to reconcile the same objects with changes to both the cluster-scoped object and namespace-scoped objects.

In the second phase, this traces back to the comment at https://github.com/openshift/multiarch-tuning-operator/pull/453#discussion_r1944944721. I would avoid users that enable the preference affinity (that tunes how pods are scheduled among differnet architectures) but no required affinity (that instead is a bare minimum to guarantee the scheduling correctness when dealing with multiple architecture nodes and pods of which the supported is unknown in the original specs).

An idea could be to let the operator create a default CPPC at installation time, but I would keep this possible work outside the scope of the feature discussed here.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: Yes, I'm essentially proposing a simple distributed transaction following the micro-services patterns.

- The scheduling gate of the local PodPlacementConfig will be used for tuning the preferences (and possibly other fields handled by future plugins)
- The scheduling gate of the ClusterPodPlacementConfig will mainly be used for handling the required affinity (or the global preferences if no PodPlacementConfig handles it already).

aleskandro: I'll add this question and respons at the end of the document.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: > so are we saying that in phase 1 - the same controller(used for cluster wide ppc) is used for namespace scoped ppc as well?

yes

> is that why we have the dependence?

Yes, but also see https://github.com/openshift/multiarch-tuning-operator/pull/453#discussion_r1944949949

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: Added this discussion in the open questions

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: Added some more details in https://github.com/openshift/multiarch-tuning-operator/pull/453/commits/ba58c764501df40a5d957a8f0ec7aeb2f81812a9, thanks for pointing this out.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: What about https://github.com/openshift/multiarch-tuning-operator/pull/453/commits/7f13afc3d8687509b437a3943cfc887c79038c53 ?

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: I reserved a list later in the doc. The idea here was to give a shorter summary/abstract about the feature. Wdyt?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+   are not handled by the namespace-scoped resource, which is focused on the scheduling scoring plugins.
+
+#### Changes to the Mutating Webhook
+
+No changes are required to the mutating webhook, as the namespace-scoped `PodPlacementConfig` CRD depends on the cluster-scoped `ClusterPodPlacementConfig` CRD and its business logic is implemented by the same controller responsible for reconciling through its scheduling gate.
+
+#### RBAC
+
+We will introduce rules to allow the
+multiarch tuning operator and pod placement controller to, respectively, reconcile (CRUD) and read
```
Prashanth684, line 159: what roles would the namespace ppc controller need? i hope it doesn't inherit all the roles from the operator

Prashanth684: ok - the below rbac section answers this question - would be nice to consolidate those

docs/enhancements/MTO-0002-local-pod-placement.md
Prashanth684: it makes sense. when there are two controllers working async - we need two scheduling gates. maybe not for phase 1 if the same controller is doing both cluster wide and namespace scoped (so only one mutating webhook in phase 1?), we can get away with one gate

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+implemented at https://github.com/openshift/multiarch-tuning-operator/pull/369.
+
+The `priority` field is an integer that defines the priority of the namespace-scoped `PodPlacementConfig` CRD. The higher the value, the higher the priority.
+If multiple namespace-scoped `PodPlacementConfig` objects are present in a namespace, the one with the highest priority will take precedence over the others when
+a pod matches the label selector of multiple `PodPlacementConfig` objects.
+
+#### Changes to the Multiarch Tuning Operator
+
+A new validating webhook will be added to the multiarch tuning operator to validate the namespace-scoped `PodPlacementConfig`.
+In particular, the webhook will deny the creation of the namespace-scoped `PodPlacementConfig` if the cluster-scoped
```
Prashanth684, line 125: can this validation be done as part of the CRD reconciliation itself? do we need a validating webhook for this? i.e, when the namespace ppc is created, check for existence of cluster wide ppc and in the cluster wide ppc reconciliation make sure it cannot be deleted

docs/enhancements/MTO-0002-local-pod-placement.md
Prashanth684: agreed with this idea as per dicussion. we should also consider creating the CPPC by default when deploying through OLM - this means the feature is "turned" on by default - but since we want that bare minimum guarantee this makes sense.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: A validation webhook should automatically in place once we build the multi version feature for the api as it is needed for future improvement. Therefore we can use it for both defaulting the priority to the right value and to check whether to deny the creation of the ppc

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+with the node selector term for the compatible architectures to influence the pod scheduling's strong predicates run by the 
+node affinity filtering plugin.
+In [MULTIARCH-4970](https://issues.redhat.com/browse/MULTIARCH-4970), we introduced new fields in the `PodPlacementConfig` that
+allow users to define cluster-wide preferences to influence the node affinity scoring plugin with respect to the pod's
+preferred architectures. However, users may want to have more granular control over the pod placement controller's behavior
+by defining different preferences for different namespaces and sub-set of pods within a namespace.
+This enhancement proposes to introduce a namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement
+controller's behavior at a namespace level. The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined 
+in the current version of the cluster-scoped `PodPlacementConfig` CRD, and label selectors to refine the scope of the configuration
+to subset of pods within a namespace matching the label selector.
```
prb112, line 35: Nice!

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+
+
+## Motivation
+- Users want to have more granular control over the pod placement controller's behavior by defining different preferences 
+for different namespaces and sub-set of pods within a namespace, given they are aware some pods can perform better on certain
```
prb112:
```suggestion
for different namespaces and a subset of pods within a namespace, given they are aware some pods can perform better on certain 
```

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+This enhancement proposes to introduce a namespace-scoped `PodPlacementConfig` CRD that allows users to define the pod placement
+controller's behavior at a namespace level. The namespace-scoped `PodPlacementConfig` CRD will include fields similar to the ones defined 
+in the current version of the cluster-scoped `PodPlacementConfig` CRD, and label selectors to refine the scope of the configuration
+to subset of pods within a namespace matching the label selector.
+
+
+## Motivation
+- Users want to have more granular control over the pod placement controller's behavior by defining different preferences 
+for different namespaces and sub-set of pods within a namespace, given they are aware some pods can perform better on certain 
+architectures or support cost-reduction strategies by running pods on cheaper nodes such as arm64 ones.
```
prb112, line 41: Also, for the situation when multiple operators share the same namespace are independently supporting different architectures. For instance, the openshift-operators example.

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+
+### User Stories
+- As a cluster administrator, I want to define different preferences for different namespaces and sub-set of pods within a namespace
```
prb112:
```suggestion
- As a cluster administrator, I want to define different preferences for different namespaces and a subset of pods within a namespace
```

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+so that I can influence the node affinity scoring plugin with respect to the pod's preferred architectures at a namespace level.
+
+### Goals
+- Provide a namespace-scoped resource to define the pod placement controller's behavior at a namespace level for the node affinity 
+  scoring plugin
+- Provide a namespace-scoped resource to define the pod placement controller's behavior for the node affinity 
+  scoring plugin on a subset of pods within a namespace
+
+### Non-Goals
+- Allow users to define multiple pod placement configurations that tune the placement of pods across multiple namespaces
```
prb112, line 54: Confirming my understanding... Does that imply a single PodPlacementConfig would be used for multiple different placements?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+- Allow users to define which namespaces should get the node affinity filtering plugin's behavior managed by the cluster-scoped
```
prb112, line 55: I mentioned the openshift-operators example namespace. Currently, openshift-* is excluded. 

Would there be a need for a relaxed approach now?

prb112: Or special treatment for that namespace?

docs/enhancements/MTO-0002-local-pod-placement.md
prb112: I think this makes sense.  ~The suggestion is for a union of the two?~ ~Does namespace override Cluster?~

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+In order to apply the configurations defined by the namespace-scoped `PodPlacementConfig` CRD, the pod placement controller will be changed such that it applies any
+matching namespace-scoped `PodPlacementConfig` CRD to the pods in the selected namespaces, sorted by ascending priority, before applying the cluster-scoped
+`ClusterPodPlacementConfig` configuration.
+
+Given a pod hitting the pod reconciliation loop:
+1. Verify the pod has the scheduling gate (return if not)
+2. Get all the pod placement configurations in the pod's namespace, sorted by descending priority
+3. For each pod placement configuration:
+	- Check if the pod matches the label selector
+	- Apply the configuration to the pod.
```
prb112, line 143: Something strange about the preview.... bad character/formatting?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,309 @@
...
+
+
+In phase2, is it necessary to let the pods go through two scheduling gates? Does it duplicate the work done by the controllers?
+
+- Using two scheduling gates enable a basic distributed transaction following the micro-services patterns, mediated by the scheduling gates set in the pod specs (distributed shared state):
+	 - The scheduling gate of the namespace-scoped `PodPlacementConfig` will be used for tuning the preferences (and possibly other fields handled by future plugins) first
+	 - Then, the scheduling gate of the ClusterPodPlacementConfig will mainly be used by the cluster-wide controller to handle the required affinity, and the global preferences if no PodPlacementConfig handles it already.
+     - The controllers of the namespace-scoped `PodPlacementConfig` will not handle the strong predicates, as (a) we want the users willing to tune their architecture-aware pod placement preferences only after the strong predicates are guarateed, and (b) we would need to allow the service accounts in the user namespace to access secrets like the global pull secret, which we might rather prefer to allow only in the core namespace running the cluster-wide controller and operator.
```
prb112, line 348: Overall awesome... I'm excited to see this in place. Thank you for the write up and consideration

prb112 (approved):
/lgtm 
minor comment

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: Yes, the namespace-scoped PPCs are applied in ascending order of priority (the highest in priority takes precedence over any others that would change the same scheduling fields). You can consider the CPPC as the lowest in priority (if no PPC match a given pod, the CPPC preferredAffinity configuration is applied).

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: MTO 1.0 already removed that restriction and changed it to ignore any pods that should land in the control plane nodes. The operator pods in the openshift-operators namespace that are not bound to the master nodes should already be considered.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: A single namespace-scoped PodPlacementConfig will only work on pods of its namespace.

If you have two pods in two different namespaces that you would patch with the same preferred affinities you should either use the CPPC or two PPCs (one per namespace)

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: See https://github.com/openshift/multiarch-tuning-operator/pull/453#discussion_r1946426754

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: It should be a tab, used to build a nested list.

aleskandro: Confirmed

prb112 (approved):
This is awesome... minor change since last update
/lgtm

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: I'd create a story for this.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: Added two tables for teh phase 2

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,348 @@
...
+| pods/status                                      | update                          |                                                                                                                     |
+| security.openshift.io/securitycontextconstraints | use                             | it's used in case the pods are created by a user/SA that needs some privileges                                      |
+| events                                           | create, patch                   |                                                                                                                     |
+| podplacementconfigs.multiarch.openshift.io       | get, list, watch                |                                                                                                                     |
+| configmaps, secrets							    | get, list, watch                | (future) (not needed by the current plugins) if other plugins will require them  to communicaet with  the registries |
```
Prashanth684: will secrets be a cluster role since we need the global pull secret?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,348 @@
...
+- Integration Test Suite: Run integration tests against a simulated control plane using the operator SDK's envtest
+  facilities. We will add the necessary test cases to ensure the reconciliation loop of the new `PodPlacementConfig` is working as expected and that pods are reconciled according to both the cluster-scoped and namespace-scoped configurations in the correct order.
+
+#### Functional Test Suite
+
+##### Phase 1
+
+- The operator should refuse the creation of a namespace-scoped `PodPlacementConfig` CRD if the cluster-scoped `ClusterPodPlacementConfig` CRD is not present.
+- The operator should refuse the deletion of the cluster-scoped `ClusterPodPlacementConfig` CRD if there are any namespace-scoped `PodPlacementConfig` CRDs in the cluster.
+- The pod placement controller should apply the configuration defined by the namespace-scoped `PodPlacementConfig` CRD to the pods in the selected namespaces, and the cluster-scoped `ClusterPodPlacementConfig` configuration for preferred affinities should be ignored
```
lwan-wanglin, line 267: is this mean that we can configure  `plugins.nodeAffinityScoring` in both cluster-scoped `ClusterPodPlacementConfig`  and namespace-scoped `PodPlacementConfig`? , and When both are configured simultaneously, namespace-scoped `PodPlacementConfig` has high priority?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,348 @@
...
+- The pod placement controller should apply the configuration defined by the highest-priority namespace-scoped `PodPlacementConfig` CRD to pods matching multiple `PodPlacementConfig` CRDs in the same namespace
+- No PPC with the same priority should be allowed in the same namespace
+- When no priority is set in the `PodPlacementConfig`, the operator should default to the lowest priority among the existing Pod Placement Configs in the same namespace decreased by 1
```
lwan-wanglin: what will occur when new ppc with no priority is created but the current lowest priority of ppc in the namespace is already 0?  will the creation be rejected?

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,348 @@
...
+- The pod placement controller should not process a pod with the configuration of the namespace-scoped `PodPlacementConfig` CRD if the pod does not match its label selector
+
+##### Phase 2
+- The operator should deploy a local pod placement controller and mutating webhook in the namespace where the first namespace-scoped `PodPlacementConfig` CRD is created
+- The operator should delete the resources deployed in the namespace when the last namespace-scoped `PodPlacementConfig` CRD is deleted
+- The local pod placement controller should reconcile the pods in the namespace where it is running
+- The cluster pod placement controller should not process a pod and remove the related scheduling gate until the local scheduling gate is removed
```
lwan-wanglin, line 277: is this sentence mean `1. local ppc process the pod and removing local scheduling gate 2. CPPC process the pod 3. CPPC remove cluster scheduling gate`?

docs/enhancements/MTO-0002-local-pod-placement.md
lwan-wanglin: it is, I review the doc more carefully, and got it https://github.com/openshift/multiarch-tuning-operator/pull/453/files#diff-27f8af5ef3f4486c4eb3358b6a8eab0f5a2b3e87af390e523c5bb6a1355b73d1R73-R74

docs/enhancements/MTO-0002-local-pod-placement.md
```diff
@@ -0,0 +1,348 @@
...
+In order to apply the configurations defined by the namespace-scoped `PodPlacementConfig` CRD, the pod placement controller will be changed such that it applies any
+matching namespace-scoped `PodPlacementConfig` CRD to the pods in the selected namespaces, sorted by ascending priority, before applying the cluster-scoped
+`ClusterPodPlacementConfig` configuration.
+
+Given a pod hitting the pod reconciliation loop:
+1. Verify the pod has the scheduling gate (return if not)
+2. Get all the pod placement configurations in the pod's namespace, sorted by descending priority
+3. For each pod placement configuration:
+	- Check if the pod matches the label selector
+	- Apply the configuration to the pod.
```
lwan-wanglin, line 143: Does the following situation exist:
the image supported architecture doesn't exist in preference Arch in plugins.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: yes.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: Hi @lwan-wanglin, can you clarify your point here?

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: That's a good point @lwan-wanglin. I updated the proposal by defaulting to 0 and increasing the value for each new one PPC that gets created.

In this way:
- 'newly' created PPCs (if the priority field is omitted), gets a higher priority (instead of a lower one)
- The issue arises when the highest priority is 255, in which case, we'd reject the creation (Adding a test case for this too).

docs/enhancements/MTO-0002-local-pod-placement.md
lwan-wanglin: what I mean is that if a pod container using single arch image like `ppc64le`, but the namespace-scoped ppc set a plugins configure like below,  finially we set nodeAffinity with supported arch `ppc64le`, right?
```
plugins:
    nodeAffinityScoring:
      enabled: true
      platforms:
        - architecture: amd64
          weight: 25
        - architecture: arm64
          weight: 75
```

Prashanth684: the required during scheduling affinity will always be added by cppc. in this case the preferred during scheduling affinity will also be added with weights - it just won't do anything

docs/enhancements/MTO-0002-local-pod-placement.md
Prashanth684: i'm wondering if we are being too rigid with the priority. if no priority is defined, we should just set it to 0. If there are multiple ppc's with no priority it's a random pick. If the user really cares, they will specify the priority. This way is less deterministic when ppcs get evaluated, but assuming that a PPC which is created first has a higher priority might not be the best assumption to go with.

On the other hand, if a value is specified, we should take care that no other ppc has the same value - I'm fine with that.

aleskandro: Given what you say, is your idea to:
1. Allow multiple PPCs with priority 0 (when users omit it, we set it to 0 automatically)
2. Do not allow multiple PPCs with same priority != 0
?

Prashanth684: that was the idea. If we want we could allow multiple ppc's with the same non-zero priority as well - but then we have to think about a strategy of the order of evaluating ppc's within each priority bucket (FIFO, lifo, etc...) . Not sure that level of complexity is required at first so I thought if priority is specified we make sure it doesn't collide

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: You're right, removed for now.

docs/enhancements/MTO-0002-local-pod-placement.md
aleskandro: I'm removing the constraint to have the priority field as autoincremented but still ensuring the priorities are unique in the same namespace.

I wouldn't go into defining a sorting system among 'buckets' of PPC with the same priority either.

When users do not set the priority, it defaults to 0.

Finally, if a PPC with priority 0 exists already, the creation fails, requesting the user to set the priority explicitly.

aleskandro: See https://github.com/openshift/multiarch-tuning-operator/pull/453/commits/6f4e4ca2aa83dbe7f0eac749574d200de6a9d435, now squashed

Prashanth684: > Finally, if a PPC with priority 0 exists already, the creation fails, requesting the user to set the priority explicitly.

I'm ok with this model for now - just wanted more flexibility from the perspective of the user in case they don't care about setting priority and the order in which it is processed - because now we are forcing them to think about it.

PR #454, opened 2025-02-06
MULTIARCH-5320: Only create the monitoring stack objects when needed

aleskandro:
This PR changes the behavior of the operator to create the monitoring objects only when needed (when the namespace is labeled to enable monitoring) so that we can avoid the prometheus operator prints warnings because finding a ServiceMonitor in a namespace that should not be monitored.


Prashanth684:
i say that we continue keeping the logic and place and just document that the label needs to be added when using the prometheus operator.

PR #458, opened 2025-02-13
on-cel-expression update to build when only the Pipeline objects are …

aleskandro:
…modified


aleskandro:
/override ci/prow/ci-index-multiarch-tuning-operator-bundle ci/prow/images

PR #523, opened 2025-03-17
WIP: Multiarch 5354 update tests

AnnaZivkovic:
Add more validation for preferred affinity


AnnaZivkovic:
Leaving in this [TODO](https://github.com/openshift/multiarch-tuning-operator/blob/a80bb579446e4029d944c5015de8d06219ea0618/pkg/e2e/operator/pod_placement_config_test.go#L305) as the test will break due to a bug where the label is not properly set when both the required affinity and preferred affinity are set by the user. The fix will be in https://github.com/openshift/multiarch-tuning-operator/pull/525

PR #525, opened 2025-03-18
MULTIARCH-5363: add logic to check if the preferred node affinity has already been set as to not overwrite it

AnnaZivkovic:
If there is an error setting the required node affinity the preferred node affinity will be set independently. However, when retrying to set the node affinity it will overwrite the preferred node affinity liable 

Adding in more e2e testing to validate results are as desired. This means incorporating MULTIARCH-5354.

### DIFF
--- a/controllers/podplacement/pod_model.go
+++ b/controllers/podplacement/pod_model.go
@@ -174,11 +174,8 @@ func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequi
 // SetPreferredArchNodeAffinity sets the node affinity for the pod to the preferences given in the ClusterPodPlacementConfig.
 // TODO[Tori]: Missing unit tests for this method.
 func (pod *Pod) SetPreferredArchNodeAffinity(cppc *v1beta1.ClusterPodPlacementConfig) {
-	// Prevent overriding of user-provided kubernetes.io/arch preferred affinities
+	// Prevent overriding of user-provided kubernetes.io/arch preferred affinities or overwriting previously set preferred affinity
 	if pod.isPreferredAffinityConfiguredForArchitecture() {
-		pod.ensureLabel(utils.PreferredNodeAffinityLabel, utils.LabelValueNotSet)
-		pod.publishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
-			ArchitecturePreferredPredicateSkippedMsg)
 		return
 	}
 
--- a/controllers/podplacement/pod_reconciler.go
+++ b/controllers/podplacement/pod_reconciler.go
@@ -142,7 +142,7 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 	}
 	// If the pod has been processed successfully or the max retries have been reached, remove the scheduling gate.
 	if err == nil || pod.maxRetries() {
-		if status := pod.Labels[utils.PreferredNodeAffinityLabel]; status == utils.LabelValueNotSet {
+		if pod.Labels[utils.PreferredNodeAffinityLabel] == utils.LabelValueNotSet {
 			pod.publishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
 				ArchitecturePreferredPredicateSkippedMsg)
 		}
--- a/controllers/podplacement/pod_reconciler.go
+++ b/controllers/podplacement/pod_reconciler.go
@@ -142,6 +142,11 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 	}
 	// If the pod has been processed successfully or the max retries have been reached, remove the scheduling gate.
 	if err == nil || pod.maxRetries() {
+		if status := pod.Labels[utils.PreferredNodeAffinityLabel]; status == utils.LabelValueNotSet {
+			pod.publishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
+				ArchitecturePreferredPredicateSkippedMsg)
+		}
+
 		log.V(1).Info("Removing the scheduling gate from pod.")
 		pod.RemoveSchedulingGate()
 	}
--- a/controllers/podplacement/pod_reconciler_test.go
+++ b/controllers/podplacement/pod_reconciler_test.go
@@ -50,6 +50,10 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 					}), "scheduling gate not removed")
 					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.SchedulingGateLabel, utils.SchedulingGateLabelValueRemoved),
 						"scheduling gate annotation not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"preferred node affinity label not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"node affinity label not found")
 				}).Should(Succeed(), "failed to remove scheduling gate from pod")
 				Eventually(func(g Gomega) {
 					g.Expect(*pod).To(HaveEquivalentNodeAffinity(
@@ -99,6 +103,10 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.SchedulingGateLabel, utils.SchedulingGateLabelValueRemoved),
 						"scheduling gate annotation not found")
 					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.ImageInspectionErrorCountLabel, strconv.Itoa(MaxRetryCount)), "image inspection error count not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"preferred node affinity label not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.NodeAffinityLabel, utils.LabelValueNotSet),
+						"node affinity label not found")
 				}).WithTimeout(e2e.WaitShort).WithPolling(time.Millisecond*250).Should(Succeed(), "failed to remove scheduling gate from pod")
 				// Polling set to 250ms such that the error count is shown in the logs at each update.
 			})
@@ -147,6 +155,10 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 					}), "scheduling gate not removed")
 					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.SchedulingGateLabel, utils.SchedulingGateLabelValueRemoved),
 						"scheduling gate annotation not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"preferred node affinity label not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"node affinity label not found")
 				}).Should(Succeed(), "failed to remove scheduling gate from pod")
 				By("Checking that the pod has the correct node affinity")
 				Expect(*pod).To(HaveEquivalentNodeAffinity(
@@ -195,6 +207,10 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 					}), "scheduling gate not removed")
 					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.SchedulingGateLabel, utils.SchedulingGateLabelValueRemoved),
 						"scheduling gate annotation not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"preferred node affinity label not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"node affinity label not found")
 				}).Should(Succeed(), "failed to remove scheduling gate from pod")
 				By("Checking that the pod has the wrong, cached, node affinity [this proves we are not querying the remote registry]")
 				Expect(*pod).To(HaveEquivalentNodeAffinity(
@@ -246,6 +262,10 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 					}), "scheduling gate not removed")
 					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.SchedulingGateLabel, utils.SchedulingGateLabelValueRemoved),
 						"scheduling gate annotation not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"preferred node affinity label not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"node affinity label not found")
 				}).Should(Succeed(), "failed to remove scheduling gate from pod")
 				By("Checking that the pod has the correct node affinity")
 				Expect(*pod).To(HaveEquivalentNodeAffinity(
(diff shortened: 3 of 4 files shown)


controllers/podplacement/pod_reconciler_test.go
```diff
@@ -564,6 +564,38 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 						"unexpected preferred node affinity")
 				}).Should(Succeed(), "failed to set preferred node affinity in pod")
 			})
+			It("sets the preferred node affinity when and label when there is an error setting the required affinity", func() {
```
Prashanth684:
```suggestion
			It("sets the preferred node affinity and label when there is an error setting the required affinity", func() {
```

Prashanth684:
lgtm

lwan-wanglin:
@AnnaZivkovic  the fix pr works well in my local test, can we also add node affinity label verification like below for e2e test, we missed label verification for e2e test either.
```
                               By("Verify node affinity label are set correct")
                               Eventually(framework.VerifyPodLabels(ctx, client, ns, "app", "test", e2e.Present,
				map[string]string{utils.NodeAffinityLabel: utils.NodeAffinityLabelValueSet,
					utils.PreferredNodeAffinityLabel: utils.NodeAffinityLabelValueSet}),
				e2e.WaitShort).Should(Succeed())
```

you can also use VerifyPodLabelsAreSet after this fix [pr](https://github.com/openshift/multiarch-tuning-operator/pull/526) merged
```
                        By("Verify node affinity label are set correct")
			Eventually(framework.VerifyPodLabelsAreSet(ctx, client, ns, "app", "test",
				utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet,
				utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet,
			), e2e.WaitShort).Should(Succeed())
```

lwan-wanglin:
hi @AnnaZivkovic, I found one more bug, when the user provides both preferred and required arch-related affinities, we are missing set multiarch.openshift.io/preferred-node-affinity label.

AnnaZivkovic:
@lwan-wanglin I added in the label checks, but I am still working out the problems. I also added in more e2e tests. Can you look over the e2es and let me know what you think?

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -368,5 +407,53 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+				WithSelectorAndPodLabels(podLabel).
+				WithPodSpec(ps).
+				WithReplicas(utils.NewPtr(int32(1))).
+				WithName("test-deployment").
+				WithNamespace(ns.Name).
+				Build()
+			err = client.Create(ctx, d)
+			Expect(err).NotTo(HaveOccurred())
+			By("The pod should not have been processed by the webhook and the scheduling gate label should be set as not-set")
+			Eventually(framework.VerifyPodLabels(ctx, client, ns, "app", "test", e2e.Present, schedulingGateNotSetLabel), e2e.WaitShort).Should(Succeed())
```
AnnaZivkovic: @lwan-wanglin Do you have any idea why the e2e are failing here? If I have this check instead it works 
```
Eventually(framework.VerifyPodLabelsAreSet(ctx, client, ns, "app", "test",
                                utils.schedulingGateNotSetLabel, utils.LabelValueNotSet,
				utils.NodeAffinityLabel, utils.LabelValueNotSet,
				utils.PreferredNodeAffinityLabel, utils.LabelValueNotSet,
			), e2e.WaitShort).Should(Succeed())
```

lwan-wanglin: if the fix for the func `VerifyPodLabelsAreSet` haven't merged at the case running?

lwan-wanglin: if `NodeAffinityScoring` disabled, do we still handle PreferredNodeAffinity related things? like set label `multiarch.openshift.io/preferred-node-affinity=not-set`?

lwan-wanglin: I checked it in my local, if NodeAffinityScoring disabled, and required node affinity conflict, the pod will get below label
```
multiarch.openshift.io/node-affinity: not-set
multiarch.openshift.io/scheduling-gate: not-set
```

AnnaZivkovic: This comes down to how we define "disabled."

Should we explicitly set a "disabled" value? If the feature is disabled, the pod would receive `multiarch.openshift.io/node-affinity: disabled`. When enabled, it would be `multiarch.openshift.io/node-affinity: not-set/set`. If the feature is nil (unset), no label would be added.

One downside is that all plugins would need to follow this convention. Additionally, it introduces more states to track and could imply the need for an explicit "enabled" label.

From a user perspective, if a feature is disabled, I wouldn't expect any trace of it. Seeing `multiarch.openshift.io/node-affinity: not-set` on a disabled feature might be confusing. As a user I would wonder why the label exists at all. However, keeping `multiarch.openshift.io/node-affinity: not-set` when disabled could be useful for validation. We could ensure that the feature is explicitly handled rather than missing due to an error.

@lwan-wanglin @aleskandro @Prashanth684 What do you think?

aleskandro: If the nodeAffinityScoring is disabled, we should not add the preferred-node-affinity label at all, IMHO.

controllers/podplacement/pod_reconciler_test.go
```diff
@@ -564,6 +564,38 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
 						"unexpected preferred node affinity")
 				}).Should(Succeed(), "failed to set preferred node affinity in pod")
 			})
+			It("sets the preferred node affinity and label when there is an error setting the required affinity", func() {
```
aleskandro, line 607: Do we need this case, or do the circuit breaker tests already cover this case?

Are all the other cases checking the correct labels already?

controllers/podplacement/pod_model.go
```diff
@@ -172,13 +172,14 @@ func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequi
...
 
 // SetPreferredArchNodeAffinity sets the node affinity for the pod to the preferences given in the ClusterPodPlacementConfig.
-// TODO[Tori]: Missing unit tests for this method.
 func (pod *Pod) SetPreferredArchNodeAffinity(cppc *v1beta1.ClusterPodPlacementConfig) {
 	// Prevent overriding of user-provided kubernetes.io/arch preferred affinities
 	if pod.isPreferredAffinityConfiguredForArchitecture() {
-		pod.ensureLabel(utils.PreferredNodeAffinityLabel, utils.LabelValueNotSet)
-		pod.publishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
-			ArchitecturePreferredPredicateSkippedMsg)
+		if _, exists := pod.Labels[utils.PreferredNodeAffinityLabel]; !exists {
```
aleskandro, line 178:
```suggestion
		if v, exists := pod.Labels[utils.PreferredNodeAffinityLabel]; !exists || v == utils.LabelValueSet {
```

do we maybe need to be more specific?

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -333,7 +332,47 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
 				Name: common.SingletonResourceObjectName,
 			}, v1alpha1obj)
 			Expect(err).NotTo(HaveOccurred(), "failed to get the v1alpha1 version of the ClusterPodPlacementConfig", err)
-			// TODO[tori] Add more validation. After the operator reconciles the operand, we should validate that both the required and preferred node affinities.
+			By("Create an ephemeral namespace")
```
aleskandro: What's the purpose of this additional logic in this case?

controllers/podplacement/pod_reconciler_test.go
AnnaZivkovic: This the only integration test that is checking the affinity labels. With the addition of  the label checking in the podplacement e2e test this case is covered.  I think it would be a good idea to also add the label checks to the integration tests

pkg/e2e/operator/pod_placement_config_test.go
AnnaZivkovic: The behavior we want going forward is that if the feature is disabled, the preferred affinity label will not be set.

controllers/podplacement/pod_reconciler_test.go
lwan-wanglin: the same case with this one https://github.com/openshift/multiarch-tuning-operator/pull/525/files#diff-df2edd807d0091a847537768f5e4c6b078a7c84491628ea26cbee5c69bc210e6R82

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -368,5 +370,56 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			By("Verify node affinity and scheduling gate label are set")
+			Eventually(framework.VerifyPodLabelsAreSet(ctx, client, ns, "app", "test",
+				utils.SchedulingGateLabel, utils.LabelValueNotSet,
+				utils.NodeAffinityLabel, utils.LabelValueNotSet,
+			), e2e.WaitShort).Should(Succeed())
+			By("Verify preferred node affinity label is not present")
+			Eventually(framework.VerifyPodLabelsAreSet(ctx, client, ns, "app", "test",
+				utils.PreferredNodeAffinityLabel, utils.LabelValueNotSet,
+			), e2e.WaitShort).ShouldNot(Succeed())
```
lwan-wanglin: this should be like below. when we use e2e.Absent, the func will check if the key of label exist, when we use e2e.Present, the func will check if the key=value of label exist.
```suggestion
			Eventually(framework.VerifyPodLabels(ctx, client, ns, "app", "test", e2e.Absent, map[string]string{utils.PreferredNodeAffinityLabel: utils.LabelValueNotSet}), e2e.WaitShort).Should(Succeed())
```
for current logic, if there is label `multiarch.openshift.io/preferred-node-affinity=set`, the verification is pass. but  if the label exist the case should fail

controllers/podplacement/pod_model_test.go
```diff
@@ -525,6 +527,175 @@ func TestPod_setArchNodeAffinity(t *testing.T) {
...
+func TestPod_SetPreferredArchNodeAffinityWithCPPC(t *testing.T) {
+	tests := []struct {
+		name string
+		pod  *v1.Pod
+		want *v1.Pod
+	}{
+		{
+			name: "pod with no predefined preferred affinity",
+			pod:  NewPod().WithContainersImages(fake.SingleArchAmd64Image).Build(),
+			want: NewPod().WithContainersImages(fake.SingleArchAmd64Image).WithPreferredDuringSchedulingIgnoredDuringExecution(
```
aleskandro: ^

AnnaZivkovic: Done simplifying unit tests

pkg/e2e/operator/pod_placement_config_test.go
Prashanth684: as long as we have logs indicating why nodeAffinity wasn't added for any reason at all - we don't need the label set i think. That being said, the feature as a whole is active, so setting to `disabled` isn't a bad idea either

controllers/podplacement/events.go
```diff
@@ -18,7 +18,7 @@ const (
 	ArchitecturePredicatesConflictMsg        = "All the scheduling predicates already include architecture-specific constraints"
 	ArchitecturePredicateSetupMsg            = "Set the supported architectures to "
 	ArchitecturePreferredPredicateSetupMsg   = "Set the architecture preferences in the nodeAffinity"
-	ArchitecturePreferredPredicateSkippedMsg = "The node affinity already includes architecture preferences"
+	ArchitecturePreferredPredicateSkippedMsg = "Preferred architecture in node affinity was not set"
```
aleskandro: Why are we changing this?

controllers/podplacement/pod_reconciler_test.go
```diff
@@ -564,6 +604,38 @@ var _ = Describe("Controllers/Podplacement/PodReconciler", func() {
...
+						Name: utils.SchedulingGateName,
+					}), "scheduling gate not removed")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.SchedulingGateLabel, utils.SchedulingGateLabelValueRemoved),
+						"scheduling gate annotation not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.ImageInspectionErrorCountLabel, strconv.Itoa(MaxRetryCount)), "image inspection error count not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet),
+						"preferred node affinity label not found")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.NodeAffinityLabel, utils.LabelValueNotSet),
+						"node affinity label not found")
+				}).WithTimeout(e2e.WaitShort).WithPolling(time.Millisecond*250).Should(Succeed(), "failed to remove scheduling gate from pod")
```
aleskandro: Can we reduce the time or use the default one?

pkg/e2e/operator/pod_placement_config_test.go
aleskandro: Is this resolved?

aleskandro (commented):
Hi @AnnaZivkovic just a few minor comments. can we also look at why the enterprise contract verification is failing and fix it in a separate PR if unrelated to this one?

controllers/podplacement/events.go
AnnaZivkovic: Because we are changing how we set the label. I was thinking that a more generic label would be better but I think the only way for this to show up is if the node affinity already includes architecture preference.  Ill change it back.

pkg/e2e/operator/pod_placement_config_test.go
AnnaZivkovic: Yes but it is part of this pr https://github.com/openshift/multiarch-tuning-operator/pull/564

AnnaZivkovic:
/

> Hi @AnnaZivkovic just a few minor comments. can we also look at why the enterprise contract verification is failing and fix it in a separate PR if unrelated to this one?

The failures are unrelated. 
```
Reason: Task "coverity-availability-check-oci-ta" is used by pipeline task "coverity-availability-check" is or will be
  unsupported as of 2025-03-31T00:00:00Z. Starting with version 0.2, the coverity-availability-check-oci-ta task is deprecated.
  Please use coverity-availability-check instead.
```

AnnaZivkovic:
[ci/prow/ocp419-e2e-gcp](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_multiarch-tuning-operator/525/pull-ci-openshift-multiarch-tuning-operator-main-ocp419-e2e-gcp/1911865565878685696) is failing because of unrelated issues. It has been tested with a earlier version of 4.19 and passes. 
The [bug](https://issues.redhat.com/browse/MULTIARCH-5405) causing the problem is due to a regression when  attempting to mount a HostPath volume for the `/etc/docker` folder. More discussion can be found [here](https://redhat-internal.slack.com/archives/C06ECDGH4LE/p1744252187694029?thread_ts=1744248579.367939&cid=C06ECDGH4LE)

AnnaZivkovic:
[Red Hat Konflux / multiarch-tuning-operator-enterprise-contract / multiarch-tuning-operator](https://github.com/openshift/multiarch-tuning-operator/pull/525/checks?check_run_id=40530669872)
We have 6 of the following error 

```
Reason: Task "coverity-availability-check-oci-ta" is used by pipeline task "coverity-availability-check" is or will be
  unsupported as of 2025-03-31T00:00:00Z. Starting with version 0.2, the coverity-availability-check-oci-ta task is deprecated.
  Please use coverity-availability-check instead.
...
```
Link to issue: https://issues.redhat.com/browse/MULTIARCH-5406

controllers/podplacement/pod_reconciler.go
```diff
@@ -142,6 +142,11 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 	}
 	// If the pod has been processed successfully or the max retries have been reached, remove the scheduling gate.
 	if err == nil || pod.maxRetries() {
+		if status := pod.Labels[utils.PreferredNodeAffinityLabel]; status == utils.LabelValueNotSet {
```
aleskandro:
```suggestion
		if pod.Labels[utils.PreferredNodeAffinityLabel] == utils.LabelValueNotSet {
```

aleskandro:

AnnaZivkovic:
/override ocp419-e2e-gcp

konflux prowjob is still a work in progress

AnnaZivkovic:

konflux prowjob is still a work in progress

PR #566, opened 2025-04-09
MULTIARCH-5010: enoexec monitoring

aleskandro:
This enhancement proposes an eBPF-based monitoring system that can detect ENOEXEC errors. Upon detection, it will emit Kubernetes events, raise alerts, and expose metrics to the monitoring stack. This enables administrators to quickly identify and diagnose architecture compatibility issues without digging through logs, facilitating smoother migrations and providing continuous observability for future workloads.


docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Introducing an eBPF-based monitoring solution for ENOEXEC (aka Exec Format Error)
+
+## Summary
+
+The Multiarch Tuning Operator can take care of setting up the node affinities of pods that are created in the cluster to 
+ensure that they are scheduled on the most appropriate nodes based on the compatible architecture of the images in the pods.
+
+However, when a node selector is already set for the architecture, or when the pods load binaries that are not compatible with the
+architecture of the node where the pod is scheduled, failures can occur and either the pod will keep restarting (CrashLoopBackOff), or
```
Prashanth684: one other case is the integrity of the image itself - it is incorrectly built and tagged for one architecture, but the binaries and libs inside are a different architecture.

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+apiVersion: multiarch.openshift.io/v1beta1
+kind: ENoExecEvent
+metadata:
+  name: <uuid>
+  namespace: <multiarch-tuning-operator-namespace>
+spec:
+status:
+  nodeName: <node-name>
+  podName: <pod-name>
+  containerID: <container-id>
```
Prashanth684, line 93: could we also add the node arch and the command arch here?

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+### ENoExecEvent definition
+
+`ENoExecEvent` is a new CRD that will be used to record the `ENOEXEC` events detected by the `enoexec-event-daemon` DaemonSet. The CRD will have the following structure in yaml format:
+
+```yaml
+apiVersion: multiarch.openshift.io/v1beta1
+kind: ENoExecEvent
+metadata:
+  name: <uuid>
```
Prashanth684, line 86: is this random or something like `<container-name><pod-name><node-id> `

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+- The container id can be used by the `enoexec-event-handler` to infer the name of the container affected by the issue by searching for the container id in the pod's status field
+- The name of the binary that triggered the `ENOEXEC` error can be obtained from the `enoexec-event-daemon`'s eBPF program by using the `bpf_get_current_comm` helper. The name may not include the full path of the binary, but it can be used to identify the binary that triggered the error.
+
+
+### Risks and Mitigations
+
+#### Security considerations
+
+In order to run the `enoexec-event-daemon`, the pod must be run with the additional capabilities: CAP_BPF, CAP_PERFMON, and CAP_SYS_RESOURCE. However, as we
```
Prashanth684: we could have one container run as root with these privileges which runs the bpf program and pipes over grpc to another container which manages the rest of the processing.

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+also need to attach the tracepoint and CoreOS kernels enable set perf_event_paranoid to 2, disallowing most tracepoint usage for non-root users.
+We need to run the container running the eBPF program as root. This is an assessed risk.
+
+To mitigate this risk, we might use a dedicated container and minimal container image for the eBPF program and share a FIFO pipe with the main container running as a non-root user and responsible for interacting with the Kubernetes API. The main container would be responsible for consuming events in the FIFO pipe and creating the `ENoExecEvent` objects. We will evaluate this option in the implementation phase.
```
Prashanth684: ok..just noticed this...yes we need to do this..we don't necessarily need a separate container image though if we restrict the entry point.....need to think on this

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+A new section will be added to the Multiarch Tuning Operator documentation to explain the new plugin and its configuration.
+
+## Implementation History
+
+## Infrastructure Needed
+
+## Open Questions
+
+- Should we harden the security of the `enoexec-event-daemon` DaemonSet by running a non-privileged container that interacts with the Kubernetes API and a privileged one for the eBPF program, possibly using a minimal image that only includes the eBPF program and dependencies (no shell, no package manager, etc.)?
```
Prashanth684: definitely should at least have a separate container with privileges which runs the bpf program

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+```yaml
+apiVersion: multiarch.openshift.io/v1beta1
+kind: ClusterPodPlacementConfig
+metadata:
+  name: my-namespace-config
+spec:
+  # [...]
+  plugins:
+    eNoExecMonitoring: ## New plugin
+      enabled: true
```
Prashanth684, line 160: what is the behaviour when toggling enabled or removing the plugin? once installed will the daemonset remain, but be disabled or will it be removed and re-installed everytime?

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: this question is answered below - but we do have the information of the pod name and container name, we could use that to construct the name as well, but I am not very opinionated on this.

Prashanth684:
lgtm overall. will provide final approval after some clarifications

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: Do we want to include this? It's a very rare case that has mostly happened in  our ci for a bug in the ci-operator controllers.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: - nodeArch can be retrieved from the handler watching this object.
- the command arch is hard. We catch the command in the kernel space as our tracepoint is attached to the exit of the execve system calls and the full path (or the process as viewed in the user-space) cannot be retrieved always (it's a very short time and we can incur in data races for everything we do at user-space).

aleskandro: i can change it, but the names of an object cannot be longer than 63 chars iirc. Also, being this an internal resource, we do not want users to play with it.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: The objects related to the plugin are fully removed. I'll specify this in the uninstallation section.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: I preferred a unix FIFO pipe (stored in an emptydir volume) as it's more efficient and faster to implement for enabling two containers in the same pod communicate in a one-way channel.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: I'll remove the open question and fix above.

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: right..it'll mostly happen internally not with well verified images

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: it's a good to have certainly not needed

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+## Summary
+
+The Multiarch Tuning Operator can take care of setting up the node affinities of pods that are created in the cluster to 
+ensure that they are scheduled on the most appropriate nodes based on the compatible architecture of the images in the pods.
+
+However, when a node selector is already set for the architecture, or when the pods load binaries that are not compatible with the
+architecture of the node where the pod is scheduled, failures can occur and either the pod will keep restarting (CrashLoopBackOff), or
+other processes will be unable to start and trigger an `ENOEXEC` error.
+
+This is particularly critical in scenarios where production-grade clusters are migrated to a multi-architecture setup and they
```
prb112: This is very common when running a non-x86 cluster as well. This could aid in porting workloads from x86 to alt-arch.

docs/enhancements/MTO-0004-enoexec-monitoring.md
prb112: to add... We also see this when an improper mirroring occurs (stripping out some entries in a manifest and not reflecting those changes in the node affinity predicates)

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+ensure that they are scheduled on the most appropriate nodes based on the compatible architecture of the images in the pods.
+
+However, when a node selector is already set for the architecture, or when the pods load binaries that are not compatible with the
+architecture of the node where the pod is scheduled, failures can occur and either the pod will keep restarting (CrashLoopBackOff), or
+other processes will be unable to start and trigger an `ENOEXEC` error.
+
+This is particularly critical in scenarios where production-grade clusters are migrated to a multi-architecture setup and they 
+start running nodes of different architectures.
+
+This enhancement proposes to introduce a new eBPF-based monitoring solution that will be able to detect the `ENOEXEC` errors in the nodes,
```
prb112:
```suggestion
This enhancement proposes a new eBPF-based monitoring solution detects the `ENOEXEC` errors in the nodes,
```

prb112: Perhaps add a link to https://github.com/torvalds/linux/blob/1a1d569a75f3ab2923cb62daf356d102e4df2b86/include/uapi/asm-generic/errno-base.h#L12

Also while looking at this link see `ENOEXEC 	the vcpu hasn't been initialized or the guest tried to execute instructions from device memory (arm64)` https://github.com/torvalds/linux/blob/1a1d569a75f3ab2923cb62daf356d102e4df2b86/Documentation/virt/kvm/api.rst#L405

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+trigger alerts and events in the pod events stream and provide metrics to the monitoring stack to allow the cluster administrators to
+detect and troubleshoot these issues without further investigation in the logs, easing the migration journey and ensuring continuous monitoring
+for the future workloads.
+
+### User Stories
+- As a cluster administrator, I want to know when an ENOEXEC occurs during an EXECVE syscall, either when starting a pod or when the 
+main process in the pod executes a binary built for an architecture mismatching the one of the node.
```
prb112: 2nd use case - Migration from Single to Multi?

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Goals
+- Provide a new plugin - `eNoExecMonitoring` - in the `ClusterPodPlacementConfig` CRD that will enable the eBPF-based monitoring solution
```
prb112: Is there a clearer name than `eNoExecMonitoring` ?

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+  to detect the `ENOEXEC` errors in the nodes and trigger alerts and events in the pod events stream.
+- To monitor the `ENOEXEC` errors in the cluster monitoring stack
+- To quickly identify pods that are failing due to `ENOEXEC` errors and troubleshoot the issue
```
prb112: Will it highlight the container? or just the Pod level? I assume the Pod only.

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Non-Goals
+- Automatically resolve the cause of `ENOEXEC` errors
+- Identify the architecture of the binary that is causing the `ENOEXEC` error and adjust the pod's node affinity accordingly
```
prb112: It might be script related or malformed content.

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Proposal
+
+We introduce a new plugin - `eNoExecMonitoring` - in the cluster-scoped `ClusterPodPlacementConfig` CRD and a new CRD, `ENoExecEvent`.
+
+The plugin let the operator deploy a publisher-subscriber architecture consisting of:
+1. `enoexec-event-daemon`: A DaemonSet that runs an eBPF-based monitoring solution on each node in the cluster
+2. `enoexec-event-handler`: A Deployment that watches for `ENoExecEvent` objects and (a) records the event in the pod's events log, (b) record the event in a Prometheus counter, (c) delete the `ENoExecEvent` object.
+3. An AlertRule that triggers an alert when the Prometheus counter value is different than the past 24h value
```
prb112: I assume it won't be limited to a namespace?

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Architecture
+
+<img src="https://raw.githubusercontent.com/openshift/multiarch-tuning-operator/1412ef6f47c6d577041f1e3f0bce2c183a4f2886/docs/enhancements/enoexec-monitoring-architecture.svg" alt="multiarch-tuning-operator enoexec monitoring architecture" width="100%" style="max-width: 60vw;display:block"/>
```
prb112: Super helpful image

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+
+The architecture of the enoexec monitoring plugin - depicted in fig. 1 - consists of a leader-elected single consumer in a publish-subscribe system that uses the Kubernetes API and the KV Store (etcd) as the message broker. Publishers are the instances
+of the `enoexec-event-daemon` DaemonSet that run on each node in the cluster and are responsible for creating the `ENoExecEvent` objects when a `ENOEXEC` error is detected in the node. The consumer is the leader-elected instance of the `enoexec-event-handler` Deployment that watches for the `ENoExecEvent` objects and records the event in the pod's events log, in a Prometheus counter and, finally, deletes the `ENoExecEvent` object.
```
prb112: Based on the arch, is it used one time ? removed after migration or available all the time?

docs/enhancements/MTO-0004-enoexec-monitoring.md
prb112: nodeArch would be helpful - +1 for it. Also, if we can get a list of images used, that'd be even better. (some of the failures I see are transient - eg Jobs).

prb112: actually .metadata.ownerReferences would be helpful if they exist in the Pod

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+```yaml
+    operators.operatorframework.io/internal-objects: '["enoexecevents.multiarch.openshift.io"]' 
+```
+
+
+### enoexec-event-daemon
+
+The `enoexec-event-daemon` runs on each node in the cluster and is responsible for detecting the `ENOEXEC` errors in the nodes and creating the `ENoExecEvent` objects when an error is detected.
+
+It runs with a dedicated service account bound to a namespace-scoped role that only allows it to create and get `ENoExecEvent` objects in the namespace where the Multiarch Tuning Operator is running. For each failure, the `ENoExecEvent` creates a new object in the namespace where the Multiarch Tuning Operator is running, named as a random string generated by FNV, and including the fields described in the previous section.
```
prb112: Once per failure in CrashBackOffLoop, or just one failure?

docs/enhancements/MTO-0004-enoexec-monitoring.md
prb112: when RHEL10 is released, the kernel supports fine grained permissions. While this change? mention as it could be two separate code paths per OS type in 4.20

prb112:
Great draft, thanks for including me in the review.

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: it should be available all the time if enabled

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: `ExecFormatErrorMonitor` ?

prb112: I think it's readable

docs/enhancements/MTO-0004-enoexec-monitoring.md
prb112: Right, I was thinking of customer use-case when I wrote the above. 

Thinking it through... and where I was going ... if this is always resident, shouldn't it be pushed down to the crun/systemd level?

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+The `enoexec-event-handler` is a Deployment that watches for `ENoExecEvent` objects and (a) records the event in the pod's events log, (b) record the event in a Prometheus counter, (c) delete the `ENoExecEvent` object.
+
+## Changes to the `ClusterPodPlacementConfig` CRD
+
+The eNoExecMonitoring plugin will be added to the `ClusterPodPlacementConfig` CRD. The plugin will be disabled by default and will
+only include the basePlugin's `enabled` field.
+
+```yaml
+apiVersion: multiarch.openshift.io/v1beta1
+kind: ClusterPodPlacementConfig
```
Prashanth684, line 153: is there a way that we can have the tuning operator be a "reporting" only tool by enabling this plugin, but having an option to disable the base feature of adding  nodeAffinity even on a cluster scoped level? I am wondering if that aligns with what the operator is meant to be.

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: good point. It should be once per failure, unless that failure has been "cleared" and occurs again

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -0,0 +1,245 @@
...
+- Automatically resolve the cause of `ENOEXEC` errors
+- Identify the architecture of the binary that is causing the `ENOEXEC` error and adjust the pod's node affinity accordingly
+
+## Proposal
+
+We introduce a new plugin - `eNoExecMonitoring` - in the cluster-scoped `ClusterPodPlacementConfig` CRD and a new CRD, `ENoExecEvent`.
+
+The plugin let the operator deploy a publisher-subscriber architecture consisting of:
+1. `enoexec-event-daemon`: A DaemonSet that runs an eBPF-based monitoring solution on each node in the cluster
+2. `enoexec-event-handler`: A Deployment that watches for `ENoExecEvent` objects and (a) records the event in the pod's events log, (b) record the event in a Prometheus counter, (c) delete the `ENoExecEvent` object.
```
Prashanth684: will alerts be cleared if the problem is fixed?

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: we will record the container id also

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: Hi @prb112 do you have a way to reproduce what you mentioned?

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: All these data can be retrieved by the consumer that has access to the API.

The only case where the nodeArch wouldn't be retrieved is when the node gets deleted between the occurrence of the error/generation of the event and its processing by the consumer.

Or, are you wondering whether these events objects should provide information that the users can inspect it by getting the enoexecevents objects?

If so, we should probably change the logic to store the events in the same namespace as the pod one instead of storing them in the operator's namespace (this means the roles to read/write enoexecevents will change to clusterroles).

aleskandro: > actually .metadata.ownerReferences would be helpful if they exist in the Pod

Would you send the event to the highest-level object that generated the pod? It can be anything, including an object from a CRD like Tekton's TaskRuns.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: I'm not getting the point here. Are you talking of whether the ENoExecEvent persists after processing it?

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: In the POC, the event is created everytime a enoexec occurs. 

So, if the pod is in crashloopBackOff because of enoexec, a new event is generated everytime the pod restarts.

There are other cases where the pod can seem running correctly but a wrong binary is run 'sometimes' and does not let the pod crash.

Assume a script like this in the entrypoint. It 'simulates' a polling loop that either process a data or not. in the example, i run a wrong binary in an iteration with a 30% probability, the error is hidden and it doesn't let the script to fail. The pod will not go into crashloobBackoff, but 30% of the time will not work correctly.

```
              #!/bin/bash
              while :; do
                if [ $(( RANDOM % 100)) -le 30 ]; then
                  echo "Processing yaml with yq for aarch64 (node architecture: $(uname -m))"
                  /arm64/yq -n '{"foo": "bar"}' 2>/dev/null
                  [ $? != 0 ] && echo "[Debug for demo] enoexec occurred (hidden in the process logs)" >&2
                else
                  echo "Nothing to process"
                fi
                sleep 1
              done
```

aleskandro: Open to suggestions on how to improve this.... I was thinking of having a TTL for the ENoExecEvent and using `hash(namespace,pod,containerId)` instead of `newUUID()` to set its name and ensure no new event is created until the TTL of the first one expires.

aleskandro: But I also think this happens on the corev1/Event side (the API increases the count everytime this occurs): https://github.com/kubernetes/api/blob/master/core/v1/types.go#L7176-L7245

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: The last discussions we had involved automatically enabling the required affinity feature at installation time instead of having an opt-in for it.

aleskandro: (users can already set the main feature as an opt-in - or disable it - by using the namespace selector)

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: As of now, the container deploying the ebpf program is considered privileged anyway and will have to run as root because of 
```
└ $ cat /proc/sys/kernel/perf_event_paranoid
2 # On fedora 41.
```

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: i would avoid the separate image too... just mentioning it because besides the entrypoint, one can still open a shell in the container and be root.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: they can be cleared manually by the users or cleared automatically based on the TTL we want to define for it.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: Do you mean having a label for the affected namespace/pod in the counter so that the metric can be either accumulated among all or focus on specific namespaces and/or pods?

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: Open to suggestions about this. I tried looking for a short enough name that is meaningful and clear to the users... 

execFormatErrorMonitor can work for me too.

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: uhm is the KVM one still appearing at the exit of `execve`?

aleskandro: (thanks for the refs, working on them while refining)

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: Added the lables in the implementation details section. If I got what you meant, we can resolve.

docs/enhancements/MTO-0004-enoexec-monitoring.md
prb112: thanks - I'm not a hundred percent sure on KVM. Just one to have a think about.

docs/enhancements/MTO-0004-enoexec-monitoring.md
prb112: I think as long as we have some ownerReference the user can use it to associate with the initiator of the Pod (e.g., Job, Operator, Deployment)

aleskandro: As of now, the user is not meant to access those objects. If we want that, I'd change to let them be created in the pods namespace

aleskandro: (We'd also define a collection strategy to avoid filling etcd with all these objects)

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: Waiting for the discussions regarding whether to consider the ENoExecEvent objects as objects the users are meant to 'consume' for investigation before resolving this. If so, we may involve labels in the ENoExecEvent or something else to make sure users can identify the objects for their purposes somehow.

aleskandro:
/hold

(for the reviewers to /lgtm first).

prb112 (approved):
/lgtm

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: i agree that we want to keep this list minimal

docs/enhancements/MTO-0004-enoexec-monitoring.md
Prashanth684: we are talking of the feature in general - even after migration this feature will be needed as nodes can be added/removed, new workloads can be used which causes disruption. As for it being part of crun, that would be a conversation to have with the container folks and see if there is a plan for having bpf in a broader framework in crun.

prb112 (approved):
/lgtm

PR #570, opened 2025-04-15
triggering a prowjob from konflux test

Prucek:


.tekton/run-prowjob.yaml
```diff
@@ -0,0 +1,73 @@
...
+          value: master
+        - name: pathInRepo
+          value: task/run-prowjob.yaml
+      params:
+        - name: CLOUD_PROVIDER
+          value: "aws"
+        - name: GANGWAY_TOKEN
+          value: gangway-token
+        - name: DEPLOY_TEST_COMMAND
+          value: "CLEANUP=true USE_OLM=false ./hack/deploy-and-e2e.sh"
```
aleskandro, line 35: Is this going to be enough?

We usually run these two steps:

- https://github.com/openshift/release/blob/cccaf7cc0f3426ab71dfd74ce500f75376d3502c/ci-operator/step-registry/multiarch/tuning-operator/deploy/multiarch-tuning-operator-deploy-commands.sh
- https://github.com/openshift/release/blob/f4f1a884f79ae59bff745fff7ffa3fd4c4fd2f22/ci-operator/step-registry/multiarch/tuning-operator/e2e/multiarch-tuning-operator-e2e-commands.sh

aleskandro: I think deploy-and-e2e.sh might try to deploy another image (built by prow in the last postsubmit) than the konflux one?

Prucek: > Is this going to be enough?

You used to test it like this, right? And you already have this script in your repository, that's why I am using it. 
> I think deploy-and-e2e.sh might try to deploy another image

I think I am overriding those values in the ci-operator config, I already had green runs with your images, I pulled them to my registry and tried in this way: https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-ci-tools-master-konflux-test-azure/1908146831347421184

aleskandro: I see, OPERATOR_IMAGE is going to be a 'standard' variable for everyone.

.tekton/run-prowjob.yaml
```diff
@@ -0,0 +1,73 @@
...
+        - name: SNAPSHOT
+          value: $(params.SNAPSHOT)
+        - name: OPENSHIFT_VERSION
+          value: "4.17"
+        - name: ARCHITECTURE
+          value: "amd64"
```
aleskandro: we'd deploy a multiarch cluster here.

Prucek: But it should work even without it. I know it's probably not testing much, but that is not the goal now

Prucek: Thank you for your input!

aleskandro: Thanks for your work! I'm just saying we want this merged too, so later on when you're done and this is ready to be used, if you could change it to be multi, we'd have this done instead of looking into it later :)

deploy/base/kustomization.yaml
```diff
@@ -2,6 +2,7 @@ kind: Kustomization
 apiVersion: kustomize.config.k8s.io/v1beta1
 
 resources:
+  - ./config.openshift.io/imagedistedmirrorsets/multiarch-tuning-operator-fbc-staging
```
aleskandro: This should not stay in the base folder as not all the deployments need this object.

Also, this will fail too because you're deploying the catalogsource and operatorgroups to deploy via FBC, while your interest is to deploy the bundle.

See https://redhat-internal.slack.com/archives/C06ECDGH4LE/p1746600031337109?thread_ts=1745928262.112699&cid=C06ECDGH4LE on how this should be achieved:
1. New kustomization env that includes the imagedigestmirrorset and the namespace
2. A kustomization in your step or a parameter to fill its content that links to the new environment.

PR #585, opened 2025-04-24
MULTIARCH-5356: Konflux: Remove files causing error

AnnaZivkovic:
"The service account token is automatically mounted for a `Kubernetes.Pod` or `Kubernetes.ServiceAccount`. Auto-mounting the service account token means this shared bearer token will be written to the container file system at `/var/run/secrets/kubernetes.io/serviceaccount`. If an attacker were to compromise the container, this token can easily be used to elevate privileges, interact with the Kubernetes API, and pivot to other resources."
  
```
  Error: SIGMA.automounting_service_account_token (CWE-284):
/code/test/manifests/01-example-no-affinity.yaml:11: Sigma main event: The service account token is automatically mounted for a `Kubernetes.Pod` or `Kubernetes.ServiceAccount`. Auto-mounting the service account token means this shared bearer token will be written to the container file system at `/var/run/secrets/kubernetes.io/serviceaccount`. If an attacker were to compromise the container, this token can easily be used to elevate privileges, interact with the Kubernetes API, and pivot to other resources.
/code/test/manifests/01-example-no-affinity.yaml:11: remediation: Disable the automount feature by explicitly setting the `automountServiceAccountToken` attribute to `false`. If this field is missing, it should be added to the resource specification. There may be scenarios where the service account token must be automounted due to backwards compatibility with certain tooling. In these cases, one must evaluate the risk of using automounted tokens and implement mitigating controls if necessary.
#    9|       fail: "yes"
#   10|   spec:
#   11|->   securityContext:
#   12|       runAsNonRoot: true
#   13|       seccompProfile:

Error: SIGMA.automounting_service_account_token (CWE-284):
/code/test/manifests/02-example-affinity-no-node-affinity.yaml:10: Sigma main event: The service account token is automatically mounted for a `Kubernetes.Pod` or `Kubernetes.ServiceAccount`. Auto-mounting the service account token means this shared bearer token will be written to the container file system at `/var/run/secrets/kubernetes.io/serviceaccount`. If an attacker were to compromise the container, this token can easily be used to elevate privileges, interact with the Kubernetes API, and pivot to other resources.
/code/test/manifests/02-example-affinity-no-node-affinity.yaml:10: remediation: Disable the automount feature by explicitly setting the `automountServiceAccountToken` attribute to `false`. If this field is missing, it should be added to the resource specification. There may be scenarios where the service account token must be automounted due to backwards compatibility with certain tooling. In these cases, one must evaluate the risk of using automounted tokens and implement mitigating controls if necessary.
#    8|       result: "This pod get the affinity update from the controller. This affinity is just set to be an empty map at the beginning."
#    9|   spec:
#   10|->   securityContext:
#   11|       runAsNonRoot: true
#   12|       seccompProfile:

....
```


AnnaZivkovic:
Closing due to changing approach to exclude https://redhat-internal.slack.com/archives/C06ECDGH4LE/p1745480635153999

PR #587, opened 2025-04-28
WIP: MULTIARCH-5413: multiarch readiness report

aleskandro:
Still very unsure about this, but I'd open the PR to start some discussions and figure out what we want to achieve with this feature.


docs/enhancements/MTO-0005-multiarch-readiness-report.md
```diff
@@ -0,0 +1,221 @@
...
+see-also: []
+---
+
+# Multi-architecture cluster readiness report
+
+## Summary
+
+The Multiarch Tuning Operator automatically configures pod node affinities to ensure workloads are scheduled on nodes with compatible CPU architectures based on the architectures supported by the images they use.
+
+This enhancement proposes the introduction of a new custom resource that will allow users to generate a multi-architecture cluster readiness report. This report will provide insights into the architecture compatibility of workloads across the cluster, enabling administrators to identify potential issues and take corrective actions proactively before they impact the availability, reliability or performance of their clusters and Service Level Objectives (SLOs).
```
prb112, line 25: This is an excellent idea.

docs/enhancements/MTO-0005-multiarch-readiness-report.md
```diff
@@ -0,0 +1,221 @@
...
+    multiarch: true
+    supportedArchitectures:
+    - arm64
+    - amd64
+  summary:
+    multiarchImages: Int
+    singlearchImages: Int
+    multiarchWorkloads: Int
+    singlearchWorkloads: Int
+```
```
prb112, line 94: This is a rough calculation of the number of unique containers in a decent sized multi-arch cluster.

```
# oc get pods -A -oyaml | grep image: | awk '{print $NF}' | sort -u | wc -l
270
```

435 characters under image x 270 images = 117K. With all the other fields, wouldn't it start exceeding the max allowable etcd size? 

Should this be externalized to a PV?

docs/enhancements/MTO-0005-multiarch-readiness-report.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Changes to the ClusterPodPlacementConfig CRD
+
+None 
+
+## Changes to the Multiarch Tuning Operator
+
+A new controller will be added to the Multiarch Tuning Operator to handle the `MultiarchReadinessReport` custom resource. This controller will be responsible for creating a Job that will generate the report and update the status of the custom resource with the generated report.
+
+The Job - `multiarch-readiness-report-gen` - will be created in the same namespace as the Multiarch Tuning Operator. The job will use the Kubernetes API to gather information about the pods and their images across the cluster. The report will be stored in the status field of the custom resource.
```
prb112, line 117: Compliance Operator has a well established pattern that may be worth consulting - specifically the Platform level analysis.

docs/enhancements/MTO-0005-multiarch-readiness-report.md
```diff
@@ -0,0 +1,221 @@
...
+## Summary
+
+The Multiarch Tuning Operator automatically configures pod node affinities to ensure workloads are scheduled on nodes with compatible CPU architectures based on the architectures supported by the images they use.
+
+This enhancement proposes the introduction of a new custom resource that will allow users to generate a multi-architecture cluster readiness report. This report will provide insights into the architecture compatibility of workloads across the cluster, enabling administrators to identify potential issues and take corrective actions proactively before they impact the availability, reliability or performance of their clusters and Service Level Objectives (SLOs).
+
+### User Stories
+
+- As a cluster administrator that wants to migrate their cluster to multi-architecture nodes, I want to be informed about
+risks and issues related to the architecture compatibility of workloads across the cluster, so that I can take corrective actions proactively before they impact the availability, reliability or performance of my clusters and Service Level Objectives (SLOs).
```
Prashanth684, line 30: Another use case:

```
As a cluster administrator who wants to migrate their cluster to multi-architecture nodes, I want to be potentially informed about the optimal configuration of nodes based on architecture so I know how to balance the number of nodes created specific to each architecture.
```

docs/enhancements/MTO-0005-multiarch-readiness-report.md
Prashanth684, line 25: In addition to this, I also see the next step to a readiness report being recommendations on how nodes can be configured optimally. When i say optimally i mean rather than saying  "yes, you need to have nodes of all architectures", we can be more nuanced and say - "i would advise you to have x% of arm64 nodes, y% of ppc64le and z% of s390x", and tailor recommendations based on number of workloads that exclusively support only a single arch or a set of arches.

docs/enhancements/MTO-0005-multiarch-readiness-report.md
```diff
@@ -0,0 +1,221 @@
...
+- Automatically resolve issues related to architecture compatibility of workloads across the cluster.
+
+## Proposal
+
+We introduce a new cluster-scoped custom resource, `multiarchReadinessReport`, that will be used to generate a multi-architecture cluster readiness report. This custom resource will be created by a user willing to generate the report. The report will be generated by a Job created by the Multiarch Tuning Operator and will contain information about the architecture compatibility of workloads across the cluster.
+
+The implementation will be split into two phases:
+1. **Phase 1**: Implement the `multiarchReadinessReport` custom resource and the Job that will generate the report. The report will be generated in a JSON format and will be stored in the status field of the custom resource. It will contain information about the architecture compatibility of workloads across the cluster, including:
+   - The number of pods that are not compatible with each possible node architecture (i.e., x86_64, aarch64, ppc64le, s390x).
+   - A mapping of each image used by the pods to its supported architectures and workload that uses it.
```
Prashanth684, line 47: this is a lot of info that needs to be stored somewhere - preferrably on disk as it would encumber memory.

docs/enhancements/MTO-0005-multiarch-readiness-report.md
Prashanth684: yes, if the report is going to contain a detailed list of all pods mapped to workload image and architecture, then we would definitely need external storage. I think it would be good to also have an option for user to generate a "sparse" report or a "full" report. In case of sparse report, we don't store the data but rather increment counters for each architecture and that would be it.

aleskandro:
closing for now as we may want to develop a kubectl plugin instead of this.

PR #590, opened 2025-04-30
Update tags to match konflux naming standard

AnnaZivkovic:
Added a v to 1-0 as we will be using v1-0 or v1.0 form this point out


aleskandro:
force merging for the sake of completeness, but once v1-x is fully up, v1-0 is considered obsolete.

PR #572, opened 2025-04-17
MULTIARCH-5351: Switched to using pipeline referenced from Tekton bundle on Quay.io

lwan-wanglin:
Switched to using pipeline referenced from Tekton bundle on Quay.io

PR #591, opened 2025-05-01
Update tekton pipeline refs and add :devel tag for the update scripts…

aleskandro:
… to parse images when looking for updates

#572 changed the pipelines to make use of a single bundle resolver to resolve them instead of a per-task approach and hard-coding the pipeline in our repo. However, we missed adding the tag before the digest pinning the version, leading the update script to fail running and proposing updates to the pipeline.


aleskandro:
/override "Red Hat Konflux / run-prowjob / multiarch-tuning-operator" "Red Hat Konflux / run-prowjob-bundle / multiarch-tuning-operator-bundle"

PR #602, opened 2025-05-05
MULTIARCH-5416: Exec format error plugin

AnnaZivkovic:
The `execFormatErrorMonitor` plugin will be introduced as part of the ClusterPodPlacementConfig custom resource. It will be disabled by default and will initially expose only the enabled field, inherited from the base plugin definition.

```
apiVersion: multiarch.openshift.io/v1beta1
kind: ClusterPodPlacementConfig
metadata:
  name: my-namespace-config
spec:
  # [...]
  plugins:
    execFormatErrorMonitor: ## New plugin
      enabled: true
  # [...]

```


apis/multiarch/common/plugins/base_plugin.go
```diff
@@ -19,10 +19,13 @@ package plugins
 // +k8s:deepcopy-gen=package
 
 // Plugins represents the plugins configuration.
+// +kubebuilder:object:generate=true
 type Plugins struct {
-	// +kubebuilder:"validation:Required
+	// +kubebuilder:validation:Required
```
aleskandro: Hi Tori, iirc we should not set these fields as required.

apis/multiarch/common/plugins/base_plugin.go
```diff
@@ -19,10 +19,11 @@ package plugins
 // +k8s:deepcopy-gen=package
 
 // Plugins represents the plugins configuration.
+// +kubebuilder:object:generate=true
 type Plugins struct {
-	// +kubebuilder:"validation:Required
 	NodeAffinityScoring *NodeAffinityScoring `json:"nodeAffinityScoring,omitempty"`
-	// Future plugins can be added here.
+
+	ExecFormatErrorMonitor *ExecFormatErrorMonitor `json:"execFormatErrorMonitor,omitempty"`
```
Prashanth684, line 26: mostly a nit - but maybe we should call it "Detector" instead of "monitor" - there is a subtle nuance because we are detecting a condition, not actively monitoring it

AnnaZivkovic: I think that is a good name

AnnaZivkovic:
/hold
This depends on https://github.com/openshift/multiarch-tuning-operator/pull/612

apis/multiarch/common/plugins/base_plugin.go
```diff
@@ -19,10 +19,11 @@ package plugins
 // +k8s:deepcopy-gen=package
 
 // Plugins represents the plugins configuration.
+// +kubebuilder:object:generate=true
 type Plugins struct {
-	// +kubebuilder:"validation:Required
 	NodeAffinityScoring *NodeAffinityScoring `json:"nodeAffinityScoring,omitempty"`
-	// Future plugins can be added here.
+
+	ExecFormatErrorDetector *ExecFormatErrorDetector `json:"execFormatErrorDetector,omitempty"`
```
aleskandro: Let's decide on the name and keep it the same in the EP:

current name in the EP: `execFormatErrorMonitor`. 

cc @Prashanth684 suggestions?

apis/multiarch/common/plugins/execFormatErrorMonitor_plugin.go
```diff
@@ -0,0 +1,33 @@
...
+
+// +kubebuilder:object:generate=true
+package plugins
+
+const (
+	// PluginName for execFormatErrorDetector.
+	execFormatErrorDetectorPluginName = "execFormatErrorDetector"
+)
+
+// ExecFormatErrorDetector is the plugin that implements the base plugin interface.
```
aleskandro: missing godoc.

aleskandro (approved):

apis/multiarch/common/plugins/base_plugin.go
Prashanth684: not a strong preference at all - but i'd like to stick with Detector if it is not too much of a hassle to change it in the EP as well.

apis/multiarch/common/plugins/execFormatErrorMonitor_plugin.go
```diff
@@ -0,0 +1,35 @@
...
+package plugins
+
+const (
+	// PluginName for execFormatErrorMonitor.
+	execFormatErrorMonitorPluginName = "execFormatErrorMonitor"
+)
+
+// ExecFormatErrorMonitor is the plugin that implements the base plugin interface.
+// Enabling this plugin starts monitoring that detects ENOEXEC errors on nodes and emits alerts and Kubernetes events in
+// the affected pods's event stream.
```
aleskandro:
```suggestion
// ExecFormatErrorMonitor is a plugin that provides Exec Format Errors events reporting and monitoring
```

(Open to others but considering this is visible to the users)

Makefile
```diff
@@ -159,6 +159,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+	$(MAKE) fmt
```
aleskandro, line 162: can you clarify the reason here?

AnnaZivkovic: I had to add `make fmt` as a step to remove the empty `import ()` block that gets added when running `make generate`. If we keep the empty import then `hack/verify-diff.sh`  will fail saying that there is unnecessary code.

apis/multiarch/common/plugins/base_plugin.go
aleskandro: No strong preference from me either

Makefile
Prashanth684: adding this to the makefile would be good for posterity

Prashanth684:
@AnnaZivkovic is this ready to go? if so i can lgtm

apis/multiarch/common/plugins/base_plugin.go
aleskandro: I'm ok with both the names. Leaving the final decision to you all, @Prashanth684 and @AnnaZivkovic. (just ensure the EP is updated accordingly)

AnnaZivkovic:
@Prashanth684 it is ready

PR #610, opened 2025-05-08
MULTIARCH-5536: support glob-based registries in the pull secrets 

aleskandro:
https://kubernetes.io/docs/concepts/containers/images/#config-json specifies Kubernetes supports glob-style matching for the host part of the registry credentials when Kubelet pulls images. The OCI, instead, does not support globs in the auth.json, hence the containers/image lib wouldn't suffice to support such cases. This PR adds the support for it.

### DIFF
--- a/pkg/image/auth.go
+++ b/pkg/image/auth.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package image
 
-import "k8s.io/apimachinery/pkg/util/json"
+import (
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/json"
+)
 
 type authData struct {
 	Auth string `json:"auth"`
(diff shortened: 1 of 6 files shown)


pkg/image/auth.go
```diff
@@ -91,3 +98,121 @@ func (ac authCfg) unmarshallAuthsDataAndStore(authsBytes []byte) error {
...
+
+	// Swap the registry glob in the host part with the image reference's host part
+	irURL, _ := ParseSchemelessURL(imageReference)
+	regURL, _ := ParseSchemelessURL(registryGlob)
+	regURL.Host = irURL.Host
+	return strings.TrimPrefix(regURL.String(), "//"), true
+}
+
+// The following code come from the kubernetes official repo, but build constraints prevent
+// us from importing it directly.
```
Prashanth684: > but build constraints prevent us from importing it directly

what are these constraints?

pkg/image/auth.go
```diff
@@ -91,3 +98,121 @@ func (ac authCfg) unmarshallAuthsDataAndStore(authsBytes []byte) error {
 func (ac authCfg) marshallAuths() ([]byte, error) {
 	return json.Marshal(ac)
 }
+
```
Prashanth684, line 98: move these to new utils file

pkg/image/auth.go
aleskandro: I think https://github.com/kubernetes/kubernetes/blob/master/pkg/.import-restrictions

Prashanth684: hmm..not sure how this works...i don't see pkg being explicitly forbidden

aleskandro: I'll retry

aleskandro: But in general it seems not a suggested path: https://discuss.kubernetes.io/t/how-to-use-a-k8s-io-kubernetes-plugin-pkg-package/29640

I didn't find those functions in apimachinery or other packages imported more often

Prashanth684: it's fine. we can look at it later after we fix this issue. It's not something i want to block on.

pkg/image/auth_test.go
```diff
@@ -0,0 +1,337 @@
...
+		},
+		{
+			name: "no glob. Registry should not be processed for expansion",
+			args: args{
+				registry:       "kubernetes.io",
+				imageReference: "kubernetes.io/foo:bar",
+			},
+			expectedOk: false,
+		},
+	}
```
Prashanth684, line 79: can there be a `kubernetes.io*` to match `kubernetes.io/foo:bar` ?

aleskandro: Nope, glob matching on the path part is not covered by the additional k8s rules.

aleskandro:
/override 'Red Hat Konflux / run-prowjob / multiarch-tuning-operator'

aleskandro:
/override "Red Hat Konflux / run-prowjob / multiarch-tuning-operator"/

aleskandro:
/override "Red Hat Konflux / run-prowjob / multiarch-tuning-operator"/

aleskandro:

aleskandro:

PR #612, opened 2025-05-08
MULTIARCH-5416: Define ENoExecEvent CRD

AnnaZivkovic:
CRD  `ENoExecEvent` should have the format of 
```
apiVersion: multiarch.openshift.io/v1beta1
kind: ENoExecEvent
metadata:
  name: <uuid>
  namespace: <multiarch-tuning-operator-namespace>
spec:
status:
  nodeName: <node-name>
  podName: <pod-name>
  podNamespace: <pod-namespace>
  containerID: <container-id>
  command: <command>
```


AnnaZivkovic:
Can I get some help figuring out what tests we should have?

apis/multiarch/v1beta1/enoexecevent_types.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+// ENoExecEventSpec defines the desired state of ENoExecEvent
+type ENoExecEventSpec struct {
+	// No fields yet — empty spec
+}
+
+// ENoExecEventStatus defines the observed state of ENoExecEvent
+type ENoExecEventStatus struct {
+	NodeName     string `json:"nodeName,omitempty"`
```
aleskandro: Is there any validation we should consider for the fields here?

aleskandro (commented):
@AnnaZivkovic have a deeper look into the enhancement and its directions for this CRD to ensure nothing is missing. As an example, i don't see whether you declared it as an internal crd.

controllers/operator/enoexecevent_controller_test.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/multiarch-tuning-operator/pkg/testing/framework"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
+
+	"github.com/openshift/multiarch-tuning-operator/pkg/testing/builder"
+)
+
+var _ = Describe("Controllers/ENoExecEvent/ENoExecEventReconciler", Serial, Ordered, func() {
```
aleskandro: This should go in a separate suite, the one of the (yet to be implemented reconciler for this resource).

apis/multiarch/v1beta1/enoexecevent_types.go
AnnaZivkovic: The only validation approach I can think of is to set the annotation and verify if it takes effect. The main constraints I know about are the name formatting rules. ex pod names: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names.

controllers/operator/enoexecevent_controller_test.go
AnnaZivkovic: Thats why I put it there. I can "pre create" the directories

Prashanth684:
> @AnnaZivkovic have a deeper look into the enhancement and its directions for this CRD to ensure nothing is missing. As an example, i don't see whether you declared it as an internal crd.

i believe this would require a Validating Admission webhook which denies creates from any SA except the ppc controllers.

AnnaZivkovic:
I can add a validating webhook that will  denies creates from any SA except the ppc controllers.

apis/multiarch/v1beta1/enoexecevent_webhook.go
```diff
@@ -0,0 +1,78 @@
...
+
+func (r *ENoExecEventValidator) validate(ctx context.Context) (warnings admission.Warnings, err error) {
+	req, ok := admission.RequestFromContext(ctx)
+	if ok != nil {
+		enoexeceventlog.Error(nil, "unable to get admission request from context")
+		return nil, fmt.Errorf("could not retrieve request info from context")
+	}
+
+	user := req.UserInfo.Username
+	expectedSA := "system:serviceaccount:your-namespace:multiarch-tuning-operator-controller-manager"
```
aleskandro: I've never seen an implementation like this and it seems to turn upside-down the way RBAC rules are meant to be used (i.e., include permissions to SAs/Users instead of excluding). Is this to be considered valid?


Without this, Users that are not kubeadmin are already denied the creation of a ENoExecEvent object, unless they explicitly setup their RBAC rules to allow the manipulation of this resource.

I'd keep the ValidatingWebHook ready for the other tasks regarding validation, and I'd also work on setting this webhook as the one for multi-versioning the ENoExecEvent resource, but forcing this rule seems an anti-pattern at a first look.

aleskandro: Moreover, the enoexecevent resource will be managed by other SAs, probably two (one for creation only, another for reconciling).

apis/multiarch/v1beta1/enoexecevent_types.go
```diff
@@ -0,0 +1,60 @@
...
+
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ENoExecEventSpec defines the desired state of ENoExecEvent
+type ENoExecEventSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
```
aleskandro:
```suggestion
	// INSERT ADDITIONAL SPEC FIELDS - desired state of the object
```

apis/multiarch/v1beta1/enoexecevent_types.go
```diff
@@ -0,0 +1,60 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
```
aleskandro, line 22: did we include the annotation to make it an internal resource?

apis/multiarch/v1beta1/enoexecevent_webhook.go
aleskandro: cc @Prashanth684 given the other comment

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -21,13 +21,28 @@ metadata:
               ]
             }
           }
+        },
```
aleskandro: shouldn't stay here as this is an internal object.

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -496,3 +520,21 @@ spec:
...
+    deploymentName: webhook
+    failurePolicy: Fail
+    generateName: venoexecevent.kb.io
+    rules:
+    - apiGroups:
+      - multiarch.openshift.io
+      apiVersions:
+      - v1beta1
+      operations:
+      - CREATE
```
aleskandro: PATCH too?

config/manifests/kustomization.yaml
```diff
@@ -25,3 +25,13 @@ resources:
...
 #      path: /spec/template/spec/volumes/0
+
+patches:
+  - target:
+      group: operators.coreos.com
+      version: v1alpha1
+      kind: ClusterServiceVersion
+    patch: |-
+      - op: add
+        path: /metadata/annotations/operators.operatorframework.io~1internal-objects
```
aleskandro: There should be an annotation in the kubebuilder for achieving this, right?

main.go
```diff
@@ -224,6 +231,19 @@ func RunOperator(mgr ctrl.Manager) {
...
 	must((&multiarchv1beta1.ClusterPodPlacementConfig{}).SetupWebhookWithManager(mgr), unableToCreateController,
 		controllerKey, "ClusterPodPlacementConfigConversionWebhook")
+
+	if err := (&multiarchv1beta1.ENoExecEvent{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "ENoExecEvent")
+	}
+}
+
+func RunENoExecEventOperandControllers(mgr ctrl.Manager) {
+	if err := (&multiarchcontrollers.ENoExecEventReconciler{
```
aleskandro: The reconciler should be implemented as part of https://issues.redhat.com/browse/MULTIARCH-5417. I'd like us to split the PRs in smaller ones.

apis/multiarch/v1beta1/enoexecevent_webhook.go
Prashanth684: > Without this, Users that are not kubeadmin are already denied the creation of a ENoExecEvent object, unless they explicitly setup their RBAC rules to allow the manipulation of this resource.

This is correct. My thought was to tighten this as much as possible to not have a chance to expose the CRD, but then again that means questioning why it even needs to be a CRD - which it needs to be to leverage etcd storage etc. Given that only the admin will have access to it - i am ok to do away with this complexity.

config/manifests/kustomization.yaml
AnnaZivkovic: I am not sure what it would be. From what I can tell //+kubebuilder modifies CRD not CSV. I could add the annotation to the CRD but it would not end up in the annotations of  the CSV.

AnnaZivkovic: I was looking through the [community operators ](https://github.com/search?q=repo%3Aoperator-framework%2Fcommunity-operators%20internal-objects&type=code)to see how they were adding it. 
I found https://github.com/apache/camel-k/blob/a04c1443caa70936f4969f1aeb5a286349439ee1/pkg/resources/config/manifests/bases/camel-k.clusterserviceversion.yaml#L30 and 	csvBase.Annotations["operators.operatorframework.io/internal-objects"] = hiddenCRDsJ

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
AnnaZivkovic: I dont see any [mention](https://github.com/operator-framework/operator-lifecycle-manager/blob/b4ec5bd5caa84725ec40de79b40f93fa08686563/doc/contributors/design-proposals/internal-objects.md) that it should be added to the CR

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
AnnaZivkovic: I dont think patch is valid according to these [docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-rules)

AnnaZivkovic:
Removing validating webhook as it is out of scope for this pr.

config/manifests/kustomization.yaml
aleskandro: yeah, i thought this was from the kubebuilder. Adding it to the base csv should be better. xref https://github.com/openshift/multiarch-tuning-operator/pull/612#discussion_r2096346935

apis/multiarch/v1beta1/enoexecevent_types.go
aleskandro: yep, let's set them for podName, namespace and nodeName at least, and containerId if possible.

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
aleskandro: I think you'd mostly remove the config/samples changes.

PROJECT
```diff
@@ -18,7 +18,7 @@ resources:
   path: k8s.io/api/core/v1
   version: v1
 - api:
-    crdVersion: v1alpha1
+    crdVersion: v1
```
aleskandro, line 21:
```suggestion
    crdVersion: v1alpha1
```

PROJECT
```diff
@@ -30,7 +30,7 @@ resources:
     conversion: true
     webhookVersion: v1
 - api:
-    crdVersion: v1beta1
+    crdVersion: v1
```
aleskandro, line 33:
```suggestion
    crdVersion: v1beta1
```

PROJECT
```diff
@@ -40,4 +40,13 @@ resources:
   webhooks:
     conversion: true
     webhookVersion: v1
+- api:
+    crdVersion: v1
```
aleskandro, line 44:
```suggestion
    crdVersion: v1beta1
```

apis/multiarch/v1beta1/enoexecevent_types.go
```diff
@@ -0,0 +1,60 @@
+/*
```
aleskandro, line 1: Can we configure the other methods to enable multi-versioning? This should be the Hub version.

config/default/kustomization.yaml
```diff
@@ -34,31 +29,10 @@ bases:
...
-#    kind: Service
-#    version: v1
-#    name: webhook-service
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- ../crd
+- ../rbac
+- ../manager
+- ../webhook
```
aleskandro: I would keep the commented vars as we don't know yet how to approach the certificate management issue for non-OCP clusters. Also, although some of the keys here are deperecated, i would dedicate another PR to any refactoring and keep this for what it is only: adding what's needed for the enoexecevents to be defined.

PROJECT
AnnaZivkovic: https://book.kubebuilder.io/reference/project-config.html

> resources.api.crdVersion - The Kubernetes API version (apiVersion) used to do the scaffolding for the CRD resource.

v1beta1 is deprecated after doing the go upgrade it has to be migrated to v1. 
Now it might be worth pulling this out into a different story and doing this 
https://book.kubebuilder.io/migration/legacy/migration_guide_v2tov3

apis/multiarch/v1beta1/enoexecevent_types.go
AnnaZivkovic: yes

apis/multiarch/v1beta1/enoexecevent_types.go
AnnaZivkovic: I definitely think we should do this, but I dont think it makes since right now. We only have one version so we would be adding a conversion webhook that would do nothing. 

https://book.kubebuilder.io/multiversion-tutorial/conversion

config/samples/multiarch_v1beta1_enoexecevent.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: multiarch.openshift.io/v1beta1
```
aleskandro: to remove

config/default/kustomization.yaml
```diff
@@ -14,12 +14,12 @@ namePrefix: multiarch-tuning-operator-
 #  someName: someValue
 
 bases:
-- ../crd
-- ../rbac
-- ../manager
-# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
-# crd/kustomization.yaml
-- ../webhook
+  - ../crd
```
aleskandro: is this an automated change or we can revert?

apis/multiarch/v1beta1/enoexecevent_types.go
aleskandro: I see a partial implementation of the conversion webhook already at https://github.com/outrigger-project/multiarch-tuning-operator/pull/612/files. Having it done, even though it's only one version, allows us not to need to implement it later when doing small fixes.

apis/multiarch/v1beta1/enoexecevent_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	NodeName string `json:"nodeName,omitempty"`
+
+	// PodName must follow the RFC 1123 DNS subdomain format:
+	// - Max length: 253 characters
+	// - Characters: lowercase letters, digits, hyphens (`-`), and dots (`.`)
+	// - Must start and end with an alphanumeric character
+	// Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
```
aleskandro, line 45: how are you verifying these patterns? If there was another way available for these "standard" fields, I'd prefer it. Otherwise we need to implement some tests too.

main.go
```diff
@@ -169,6 +170,13 @@ func main() {
 	})
 	must(err, "unable to create manager")
 
+	if err = (&multiarchcontrollers.ENoExecEventReconciler{
```
aleskandro: can we comment this snippet. I'll work later on the ENoExecEventReconciler development. For now, I'd avoid it from running at all.

apis/multiarch/v1beta1/enoexecevent_types.go
AnnaZivkovic: I added a bunch of integration tests

controllers/enoexecevent/enoexecevent_controller_test.go
```diff
@@ -0,0 +1,228 @@
...
+					WithNodeName("test-name-").
+					WithNamespace(utils.Namespace()).
+					Build()
+				err = k8sClient.Create(ctx, enee)
+				Expect(err).To(HaveOccurred(), "Created ENoExecEvent that does not end with alphanumeric character", err)
+			})
+			It("should reject a PodName that exceeds 253 character", func() {
+				enee := builder.NewENoExecEvent().WithName("invalid-node").WithNamespace(utils.Namespace()).
+					WithNodeName("a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890.a234567890").Build()
+				err := k8sClient.Create(ctx, enee)
```
aleskandro: I think the error is here: with k8sClient.Create, you won't write the status field of the resource.

aleskandro: you may need to add k8sClient.Status().Update(ctx, enee)

controllers/enoexecevent/enoexecevent_controller_test.go
```diff
@@ -0,0 +1,249 @@
...
+				Expect(err).NotTo(HaveOccurred(), "failed to delete ENoExecEvent", err)
+				Eventually(framework.ValidateDeletion(k8sClient, ctx)).Should(Succeed(), "the ENoExecEvent should be deleted")
+			})
+		})
+		Context("is validating the fields of the ENoExecEvent CR", func() {
+			var enee *v1beta1.ENoExecEvent
+			var eneeName string
+			BeforeEach(func() {
+				// Use a unique name based on the Ginkgo test node ID
+				eneeName = fmt.Sprintf("invalid-node-%d", GinkgoParallelProcess())
```
aleskandro:
```suggestion
				eneeName = fmt.Sprintf("invalid-node-%d", GinkgoParallelProcess())
```

can we use a uuid here instead? You'd find some facility to generate names in the test suite already.

apis/multiarch/v1beta1/enoexecevent_types.go
```diff
@@ -0,0 +1,86 @@
...
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of the object
+}
+
+// ENoExecEventStatus defines the observed state of ENoExecEvent
+type ENoExecEventStatus struct {
+	// NodeName must follow the RFC 1123 DNS label format:
+	// - Max length: 63 characters
+	// - Characters: lowercase letters, digits, and hyphens (`-`)
+	// - Must start and end with an alphanumeric character
+	// Ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names
```
aleskandro, line 34: Can you link any other source you have used to define patterns and maxlength for node, pods and namespaces?

AnnaZivkovic: I tried to find documentation, but I dont think docker explicitly says that the container id  is 64 char long it is just a hash of a sha256 and is more of a convention

PR #614, opened 2025-05-09
Fix bump version script

aleskandro:
The bump-version.sh was missing a regexp to replace the `version` label in the Dockerfile


AnnaZivkovic:

PR #616, opened 2025-05-09
MULTIARCH-5538: Clear cache when global pull secret changes

Prashanth684:
When there is an update to the global pull secrets, we need to clear the cache as access might have changed. This needs to be done because the cache keys do not use the global pull secret.

### DIFF
--- a/pkg/e2e/podplacement/pod_placement_test.go
+++ b/pkg/e2e/podplacement/pod_placement_test.go
@@ -26,6 +26,7 @@ const (
 	helloOpenshiftPrivateMultiarchImageGlobal = "quay.io/multi-arch/tuning-test-global:1.1.0"
 	helloOpenshiftPrivateArmImageGlobal       = "quay.io/multi-arch/tuning-test-global:arm-1.1.0"
 	helloOpenshiftPrivateArmPpcImageGlobal    = "quay.io/multi-arch/tuning-test-global:arm-ppc64le-1.1.0"
+	helloOpenshiftPrivateArmPpcImageGlobal2   = "quay.io/multi-arch/tuning-test-global-2:arm-ppc64le-1.1.0"
 	helloOpenshiftPrivateMultiarchImageLocal  = "quay.io/multi-arch/tuning-test-local:1.1.0"
 	helloOpenshiftPrivateArmImageLocal        = "quay.io/multi-arch/tuning-test-local:arm-1.1.0"
 	helloOpenshiftPrivateArmPpcImageLocal     = "quay.io/multi-arch/tuning-test-local:arm-ppc64le-1.1.0"
@@ -37,6 +38,7 @@ const (
 var _ = Describe("The Pod Placement Operand", func() {
 	var (
 		podLabel                  = map[string]string{"app": "test"}
+		podLabelPullSecretUpdate  = map[string]string{"app": "test-post-update"}
 		schedulingGateLabel       = map[string]string{utils.SchedulingGateLabel: utils.SchedulingGateLabelValueRemoved}
 		schedulingGateNotSetLabel = map[string]string{utils.SchedulingGateLabel: utils.LabelValueNotSet}
 	)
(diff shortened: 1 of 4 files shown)


controllers/podplacement/global_pull_secret.go
```diff
@@ -87,6 +87,7 @@ func (s *GlobalPullSecretSyncer) onAddOrUpdate(obj interface{}) {
 	s.log.Info("The global pull secret was updated")
 	if pullSecret, err := utils.ExtractAuthFromSecret(secret); err == nil {
 		image.FacadeSingleton().StoreGlobalPullSecret(pullSecret)
+		image.FacadeSingleton().ClearCache()
```
aleskandro: This exposes the logic of the facade, violating the pattern.

The Facade's StoreGlobalPullSecret can perform the clearing, avoiding the need for facade users (e.g., this informer) to be aware of the internal logic (whether a cache exists and how it is handled, ...). This is important, especially in case we need to implement other Facade/Cache/Registry Inspector abstractions for different ways of providing credentials. Also, this prevents us from exporting all the methods.

```suggestion
```

pkg/image/facade.go
```diff
@@ -42,6 +42,10 @@ func (i *Facade) StoreGlobalPullSecret(pullSecret []byte) {
 	i.storeGlobalPullSecret(pullSecret)
 }
 
+func (i *Facade) ClearCache() {
+	i.inspectionCache.ClearCache()
+}
+
```
aleskandro:
```suggestion
	i.storeGlobalPullSecret(pullSecret)
	i.inspectionCache.ClearCache()
}

```

pkg/image/interfaces.go
```diff
@@ -26,6 +26,8 @@ type ICache interface {
 	// GetCompatibleArchitecturesSet takes an image reference. a list of secrets and the client to the cluster and
 	// returns a set of architectures that are compatible with the image reference.
 	GetCompatibleArchitecturesSet(ctx context.Context, imageReference string, skipCache bool, secrets [][]byte) (sets.Set[string], error)
+	// clearCache purges the cache of all entries
+	ClearCache()
```
aleskandro:
```suggestion
	// storeGlobalPullSecret takes a pull secret and stores it in the ImageFacade. It will be used by the controller
	// in charge of watching the global pull secret and to store it in the ImageFacade's relevant private field.
	// Then, the ImageFacade will be responsible for consuming it during the inspection.
storeGlobalPullSecret(pullSecret []byte)
```

and remove it at lines 34-38

pkg/testing/image/fake/cache.go
```diff
@@ -23,6 +23,10 @@ func (c *cacheProxy) GetCompatibleArchitecturesSet(ctx context.Context, imageRef
 	return nil, errors.New("image not found")
 }
 
+func (c *cacheProxy) ClearCache() {
```
aleskandro:
```suggestion
func (c *cacheProxy) storeGlobalPullSecret(_ []byte) {
```

pkg/testing/image/fake/facade.go
```diff
@@ -24,6 +24,10 @@ func (i *Facade) GetCompatibleArchitecturesSet(ctx context.Context, imageReferen
 	return i.inspectionCache.GetCompatibleArchitecturesSet(ctx, imageReference, skipCache, secrets)
 }
 
+func (i *Facade) ClearCache() {
+	i.inspectionCache.ClearCache()
+}
+
```
aleskandro:
```suggestion
```

you might also need to change the type at https://github.com/openshift/multiarch-tuning-operator/pull/616/files#diff-1096bc06a68b3abfb6acddf7c65ef2909bf3fafb75928bfcf731ef6d7dfe0fbaR19 to cacheProxy.

pkg/image/cache.go
```diff
@@ -77,6 +77,12 @@ func newCacheProxy() *cacheProxy {
 	}
 }
 
+func (c *cacheProxy) ClearCache() {
```
aleskandro:
```suggestion
func (c *cacheProxy) storeGlobalPullSecret(_ []byte) {
```

this way and by moving storeGlobalPullSecret from the RegistryInspector to ICache interface, the logic of storeGlobalPullSecret is left to the specific cache implementation, with the Purge of the cache being an internal-to-this-package optimization choice as we were discussing, hidden from the rest of the code that can eventually make use of other implementations of RegistryInspector/Cache pairs (e.g. for the new k8s SA-based auth feature I sent yesterday).

pkg/image/inspector.go
```diff
@@ -252,6 +252,11 @@ func writeMemFile(name string, b []byte) (int, error) {
 	return fd, nil
 }
 
+// clearCache clears the image inspection cache - not applicable here
+func (i *registryInspector) ClearCache() {
```
aleskandro: We wouldn't need a 'void' implementation here if re-using storeGlobalPullSecret to satisfy the interface design.

lwan-wanglin:
/retitle [WIP] [MULTIARCH-5538](https://issues.redhat.com//browse/MULTIARCH-5538): Clear cache when global pull secret changes

lwan-wanglin:
/retitle [WIP] [MULTIARCH-5538](https://issues.redhat.com//browse/MULTIARCH-5538): Clear cache when global pull secret changes

controllers/podplacement/global_pull_secret.go
Prashanth684: ah..right...you did mention that...been a while since i did OOP :)

pkg/image/cache.go
Prashanth684: yup i understand the change - but don't really like that we are using `storeGlobalPullSecret` to clear the cache - it's kinda confusing as it has nothing to do with caching - it satisfies the pattern, but maybe worth revisting to see if this can be structured better to accomodate more functions without violating the pattern

pkg/image/facade.go
```diff
@@ -42,6 +42,10 @@ func (i *Facade) StoreGlobalPullSecret(pullSecret []byte) {
 	i.storeGlobalPullSecret(pullSecret)
```
Prashanth684, line 43: so with this change - similar to `GetCompatibleArchitectureSet`, the `StoreGlobalPullSecret` will also have to go through cache and then call the registry inspector's `storeGlobalPullSecret`, correct? because now:
when new facade is created:
```
storeGlobalPullSecret: inspectionCache.registryInspector.storeGlobalPullSecret,
```
which cannot be the case anymore.

pkg/image/cache.go
Prashanth684: > by moving storeGlobalPullSecret from the RegistryInspector to ICache interface

this will cause a problem in the testing framework because ICache will have an unexported method and even if we change the fake cache to be CacheProxy type rather than ICache, the assignment [here](https://github.com/openshift/multiarch-tuning-operator/blob/main/controllers/podplacement/pod_model_test.go#L282) fails since that is also used to assign it to an ICache object later.

Prashanth684:
simplified the implementation quite a lot, but if we need to scale this, we need to design it in such a way that we create the interfaces by decoupling the internal functionality and the functions we want to expose externally.

pkg/image/cache.go
```diff
@@ -70,6 +70,13 @@ func (c *cacheProxy) GetRegistryInspector() IRegistryInspector {
 	return c.registryInspector
 }
 
+func (c *cacheProxy) clearCache() {
+	// Purge clears the cache completely.
+	// onEvict is called for each evicted key.
```
aleskandro:
```suggestion
// clearCache purges the image metadata cache
func (c *cacheProxy) clearCache() {
```

pkg/image/cache.go
```diff
@@ -70,6 +70,13 @@ func (c *cacheProxy) GetRegistryInspector() IRegistryInspector {
...
+	c.imageRefsCache.Purge()
+	// TBD: Reset the metrics?
```
aleskandro: Do we have metrics for the cache @AnnaZivkovic ?

pkg/image/facade.go
aleskandro: This looks great to me.

aleskandro:
> simplified the implementation quite a lot, but if we need to scale this, we need to design it in such a way that we create the interfaces by decoupling the internal functionality and the functions we want to expose externally.

Thanks @Prashanth684 . This looks ok as fix to me. Yes, we'd have a deeper look into this package as it was one of  the first we implemented for the PoC and might need some refactoring.

aleskandro:
@Prashanth684 can we implement E2Es and - if possible - some integration tests for this bugfix?

Other than that, it looks good to me.

pkg/image/cache.go
AnnaZivkovic: I think so. We have metrics here https://github.com/openshift/multiarch-tuning-operator/blob/6c0648eba8ce48e88aa2c29d8b47ae2dd63cceca/pkg/image/cache.go#L41-L54

Prashanth684:
Added e2e test. I was having some trouble adding an integration test because it looks like the global pull secret that is used for those tests doesn't actually do anything (it actually doesn't match and authenticates the user anonymously) - @lwan-wanglin or @AnnaZivkovic when you have some time could you look into the global/local pull secret tests for integration?

pkg/e2e/podplacement/pod_placement_test.go
```diff
@@ -1014,6 +1015,84 @@ var _ = Describe("The Pod Placement Operand", func() {
...
+				utils.PreferredNodeAffinityLabel, utils.NodeAffinityLabelValueSet,
+			), e2e.WaitShort).Should(Succeed())
+			By("The pod should have been set node affinity of arch info.")
+			Eventually(framework.VerifyPodNodeAffinity(ctx, client, ns, "app", "test", *expectedNSTs), e2e.WaitShort).Should(Succeed())
+			By("The pod should have the preferred affinities set in the ClusterPodPlacementConfig")
+			Eventually(framework.VerifyPodPreferredNodeAffinity(ctx, client, ns, "app", "test",
+				defaultExpectedAffinityTerms()), e2e.WaitShort).Should(Succeed())
+			// delete the pullsecret from the global pull secret
+			By("Update the global pull secret to remove the credentials for the private image")
+			updateGlobalPullSecret(true)
```
lwan-wanglin: I think we should consider removing the other global pull secret E2E test cases. First, they are already covered by this new test. Second, all PodPlacement E2E cases run in parallel. If one test deletes the global pull secret, it may cause other tests to fail.
The current concern is that we have two other test cases related to the global pull secret:, [1: should set the node affinity in pods with images requiring credentials set in the global pull secret](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/e2e/podplacement/pod_placement_test.go#L970) and [2: should set the node affinity in pods with images that require both global and local pull secrets](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/e2e/podplacement/pod_placement_test.go#L1142) 
Should we just add this additional test to the second test case mentioned above and remove the first one?  but that would make the test case hard to read.
Should we consider creating a new repository for the cache clean-up test?
cc @Prashanth684 @aleskandro

lwan-wanglin:
> Added e2e test. I was having some trouble adding an integration test because it looks like the global pull secret that is used for those tests doesn't actually do anything (it actually doesn't match and authenticates the user anonymously) - @lwan-wanglin or @AnnaZivkovic when you have some time could you look into the global/local pull secret tests for integration?

Are you saying that this [pull-secret](https://github.com/openshift/multiarch-tuning-operator/blob/main/controllers/podplacement/suite_test.go#L305-L309) for [registry](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/testing/image/fake/registry/registry.go#L46) doesn't take effect actually?

Prashanth684:
> > Added e2e test. I was having some trouble adding an integration test because it looks like the global pull secret that is used for those tests doesn't actually do anything (it actually doesn't match and authenticates the user anonymously) - @lwan-wanglin or @AnnaZivkovic when you have some time could you look into the global/local pull secret tests for integration?
> 
> Are you saying that this [pull-secret](https://github.com/openshift/multiarch-tuning-operator/blob/main/controllers/podplacement/suite_test.go#L305-L309) for [registry](https://github.com/openshift/multiarch-tuning-operator/blob/main/pkg/testing/image/fake/registry/registry.go#L46) doesn't take effect actually?

Exactly. I am not sure how that is being enforced at the registry level, but even without seeding the pull-secret , the access works

pkg/e2e/podplacement/pod_placement_test.go
aleskandro: is it rather possible to:
1. add a pull secret to the global pull secret
2. test that inspection works
3. remove that pull secret from the global pull secret
4. test that inspection does not return the previous results

This should allow us to run this test in parallel with others.

lwan-wanglin: That's already what Prashanth did. My concern is that if we run this newly added test case in parallel, we'll need to remove test cases 1 and 2 that I mentioned earlier.
Test case 1 can be removed without any issues, but test case 2 is trying to verify a pod that requires both a local image and a pull secret.

aleskandro: If we use a pull secret that is not involved in the other cases (as in it provides access to images the other cases do not use), the tests must work, or either them or the operator need a fix, correct me if I'm wrong.

Prashanth684: right we need a new image in `quay.io/multi-arch` with  a different pull secret that we can use for this test case.

Prashanth684: added a new image and pull-secret token just for this test

lwan-wanglin: Hi Alessandro, I misread your comment, if we add a new pull secret to global pull secret, it won't influnce other test cases

pkg/image/cache.go
Prashanth684: i'm guessing there's nothing to do here - as the next time `GetCompatibleArchitectures` is called, it'll reset it anyway

aleskandro: can you remove the comment so that we merge this?

```suggestion
```

Prashanth684:
@aleskandro approve?

aleskandro (commented):
/lgtm

pkg/image/cache.go
Prashanth684: done

PR #625, opened 2025-05-12
WIP  MULTIARCH-5365: Implement local PodPlacementConfig API

pkenchap:
The new CRD is defined and available as golang types and manifests based on
The CRD v1beta1 is set as hub for multi-version definitions


apis/multiarch/common/priority.go
```diff
@@ -0,0 +1,26 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package common
+
+// +k8s:deepcopy-gen=package
+
+// Priority represents the priority configuration.
+// priority field indicates the priority of in range 1-100.
```
prb112, line 23: You could cite https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L3308 as the defining definition

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -302,6 +302,78 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+		})
+		It("should fail creating the CPPC with multiple items for the same architecture in the plugins.nodeAffinityScoring.Platforms list", func() {
+			By("Creating a v1beta1 PodPlacementConfig")
+			err := client.Create(ctx,
+				NewPodPlacementConfig().
+					WithName(common.SingletonResourceObjectName).
+					WithPriority(100).
+					WithNodeAffinityScoring(true).
+					WithNodeAffinityScoringTerm(utils.ArchitectureAmd64, 50).
+					WithNodeAffinityScoringTerm(utils.ArchitectureAmd64, 100).
```
prb112, line 371: Are the same architectures meant to be used?

PR #629, opened 2025-05-15
Automated checking snapshot images

AnnaZivkovic:
This is checking that the bundle in the snapshot contains the same  operator image.  Can now run ./hack/snapshot.sh -v <version> -s <snapshot for version>
Will return 

- No match: The internal image SHA is NOT present in snapshot images.
- Match found: The internal image SHA is present in snapshot images.

can also use make to run ex. 
`make verify-snapshots VERSION=v1-0 SNAPSHOT=my-snapshot`

### DIFF
--- a/docs/ocp-release.md
+++ b/docs/ocp-release.md
@@ -189,9 +189,12 @@ If so, update hack/bump-version.sh to include any further patches required to up
 2. When planning to release a new version (e.g. v1.0) or a patch version (e.g., v1.0.z), select a snapshot from the
 corresponding Konflux application
 and ensure all post-submit checks have successfully passed.
-```shell
-oc get snapshots --sort-by .metadata.creationTimestamp -l pac.test.appstudio.openshift.io/event-type=push,appstudio.openshift.io/application=multiarch-tuning-operator-<version>
-```
+There are two ways to check if the snapshot contains the right images.
+   - Run `./hack/check-snapshots.sh` and skip to step 4
+   - Manually check with the following command and steps  2 and 3
+    ```shell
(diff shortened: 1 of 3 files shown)


docs/ocp-release.md
```diff
@@ -189,9 +189,12 @@ If so, update hack/bump-version.sh to include any further patches required to up
 2. When planning to release a new version (e.g. v1.0) or a patch version (e.g., v1.0.z), select a snapshot from the
 corresponding Konflux application
 and ensure all post-submit checks have successfully passed.
-```shell
-oc get snapshots --sort-by .metadata.creationTimestamp -l pac.test.appstudio.openshift.io/event-type=push,appstudio.openshift.io/application=multiarch-tuning-operator-<version>
-```
+There are two ways to check if the snapshot contains the right images.
+   - Run `./hack/check-snapshots.sh` and skip to step 4
```
aleskandro, line 193: can we add a makefile target for this such that we can launch it as 

```
make verify-snapshots [SNAPSHOT=..] [VERSION=..]
```
?

AnnaZivkovic: Can now run 
`make verify-snapshots VERSION=v1-0 SNAPSHOT=my-snapshot`

aleskandro (commented):
/lgtm
/approve

hack/check-snapshots.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/sh
```
aleskandro, line 1: is the mode of this file and the odthers set to 0755?

AnnaZivkovic: yes

PR #635, opened 2025-05-19
.tekton: add placeholder pipelines to make tests always pass

Prucek:
This PR adds temporary pipelines for the run-prowjob tasks so that in other PRs, the nonexistent pipelines don't fail and don't need to be overridden.


aleskandro (approved):
Thanks @Prucek

PR #644, opened 2025-05-30
MULTIARCH-5417: ENoExecEvent reconciler

aleskandro:
This PR implements the enoexec event reconciler to:
1. Publish an event in the pod affected by enoexec
2. Label the pod
3. Remove the ENoExecEvent temporary object

Blocked by 
- #643
- #612

### DIFF
--- a/controllers/enoexecevent/enoexecevent_controller.go
+++ b/controllers/enoexecevent/enoexecevent_controller.go
@@ -18,45 +18,153 @@ package enoexecevent
 
 import (
 	"context"
+	runtime2 "runtime"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/record"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	ctrl2 "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	multiarchv1beta1 "github.com/openshift/multiarch-tuning-operator/apis/multiarch/v1beta1"
+	"github.com/openshift/multiarch-tuning-operator/pkg/models"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
 )
 
-// ENoExecEventReconciler reconciles a ENoExecEvent object
-type ENoExecEventReconciler struct {
+// Reconciler reconciles a ENoExecEvent object
+type Reconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
+	clientSet *kubernetes.Clientset
+	Scheme    *runtime.Scheme
+	recorder  record.EventRecorder
+}
+
+func NewReconciler(client client.Client, clientSet *kubernetes.Clientset, scheme *runtime.Scheme, recorder record.EventRecorder) *Reconciler {
+	return &Reconciler{
+		Client:    client,
+		clientSet: clientSet,
+		Scheme:    scheme,
+		recorder:  recorder,
+	}
 }
 
 //+kubebuilder:rbac:groups=multiarch.openshift.io,resources=enoexecevents,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=multiarch.openshift.io,resources=enoexecevents/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=multiarch.openshift.io,resources=enoexecevents/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=pods,verbs=get;list;update
+//+kubebuilder:rbac:groups="",resources=nodes,verbs=get;list
 
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
-// TODO(user): Modify the Reconcile function to compare the state specified by
-// the ENoExecEvent object against the actual cluster state, and then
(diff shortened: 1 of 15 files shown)


controllers/enoexecevent/enoexecevent_controller_test.go
```diff
@@ -5,33 +5,244 @@ import (
...
 				Expect(err).NotTo(HaveOccurred(), "failed to create ENoExecEvent", err)
 				By("Deleting the ENoExecEvent")
 				err = k8sClient.Delete(ctx, enee)
 				Expect(err).NotTo(HaveOccurred(), "failed to delete ENoExecEvent", err)
-				Eventually(framework.ValidateDeletion(k8sClient, ctx)).Should(Succeed(), "the ENoExecEvent should be deleted")
 			})
 			It("should create a ENoExecEvent CR with set fields", func() {
 				By("Creating the ENoExecEvent")
-				enee := builder.NewENoExecEvent().WithName("test-name").WithNamespace(utils.Namespace()).Build()
+				enee := builder.NewENoExecEvent().WithName("test-name").WithNamespace(testNamespace).Build()
```
lwan-wanglin, line 245: I have a question about the test case: why isn't the ENoExecEvent automatically deleted by Reconcile() in this case?

pkg/models/pod_test.go
```diff
@@ -545,3 +545,139 @@ func TestPod_IsFromDaemonSet(t *testing.T) {
...
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		want    string
+		wantErr bool
+	}{
+		{
+			name: "valid container ID matches a container (cri-o)",
```
AnnaZivkovic, line 566: Did we want to add a tests for docker://<> as well?

AnnaZivkovic:
Can we add a unit tests to check if the container has invalid characters as well? For example
`containerd://abcdef1234567890abcdef1234567890a-cdef1234567890abcdef1234567890`
`containerd://abcdef1234567890abcdef1234567890abcdef1234567890abcdef12ghijklmz`

pkg/models/pod.go
```diff
@@ -153,3 +155,24 @@ func (pod *Pod) IsFromDaemonSet() bool {
 	}
 	return false
 }
+
+func (pod *Pod) ContainerNameFor(containerID string) (string, error) {
+	// The containerID is in the format: "runtime://<64-hex-chars>"
+	matched, err := regexp.MatchString(`^.+://[a-f0-9]{64}$`, containerID)
```
AnnaZivkovic, line 160: Should we add a TODO here to note that we’re assuming it’s 64 characters long in accordance with this story (https://issues.redhat.com/browse/MULTIARCH-5547)? This would help us track and revisit the assumption later.

controllers/enoexecevent/enoexecevent_controller.go
```diff
@@ -18,45 +18,153 @@ package enoexecevent
...
+		logger.Info("Pod is not scheduled on the node where the ENoExecEvent was generated", "podName",
+			pod.Name, "namespace", pod.Namespace, "nodeName", enoExecEvent.Status.NodeName)
+		return ctrl.Result{}, nil
+	}
 
-	// TODO(user): your logic here
+	// Retrieve the node
+	node, err := r.clientSet.CoreV1().Nodes().Get(ctx, enoExecEvent.Status.NodeName, metav1.GetOptions{})
+	if err != nil {
+		// When the error is "not found", the ENoExecEvent may refer a new pod with the same name scheduled on a different node.
```
AnnaZivkovic, line 125: Just trying to follow what is happening here. Maybe a pod was recreated on another node (rolling update/eviction).  Hence why the pod was found but the node is not. So if the node doesn’t exist anymore, it’s probably safe to ignore this ENoExecEvent and let the caller clean it up?

pkg/models/pod_test.go
aleskandro: There is one for containerd already to cover a different runtime class in the "protocol" part of the uri.

pkg/models/pod.go
aleskandro: I'd rather add a link to this function in that story to make sure we look at this again. TODO comments need to be looked up in the codebase and they get easily ignored.

aleskandro:
> Can we add a unit tests to check if the container has invalid characters as well? For example `containerd://abcdef1234567890abcdef1234567890a-cdef1234567890abcdef1234567890` `containerd://abcdef1234567890abcdef1234567890abcdef1234567890abcdef12ghijklmz`

Isn't this covered by your work about enabling the API? We shouldn't even get such case in the code involved in this PR.

controllers/enoexecevent/enoexecevent_controller_test.go
aleskandro: I changed the namespace to a different namespace than the operator's one so that the reconciler does not work on it (otherwise tests can fail based on how fast the reconciler is to delete the object before the test concludes).

aleskandro: This should be resilient to the future validating webhook (in case we will deny creation of these objects in other namespaces than the operator's one) as  the integration tests will not hit the validating webhook policies.

controllers/enoexecevent/enoexecevent_controller.go
aleskandro: What can happen here is that:
1. A pod P that causes ENOEXEC is created and scheduled in node N
2. The ENoExevEvent is created for pod P, node N, container C
3. Assume the reconcile is busy with other ENoExecEvents (there are many and the service time increases on the reconciler side)
4. While the reconciler has not processed the object, the user deletes the pod and creates another with the same name.
5. When the reconciler processes the ENoExecObject, it could cast the event to the newly created pod that is not really affected by the exec format error anymore.


So what I try to do here is to validate that the data in the ENoExecEvent object are still valid: the pod is bound to the node on it, both the node and pod exist, the container id should be found in the status.containerStatuses field.

PR #646, opened 2025-06-05
MULTIARCH-5546: Fix bundle image detection logic for multi-arch scheduling

### LINKED ISSUE #645: Operator installation fails on s390x architecture via OLM on OpenShift
Environment: Openshift 4.18.14 on s390x architecture, z16 generation, with multiarch-tuning-operator v1.1.1.
Problem exists only when multiarch-tuning-operator is installed, it does NOT exist without this operator.

Image

When installing OLM operator, e.g. `Streams for Apache Kafka` into arbitrary namespace, installation fails due to the failure in the OLM operator bundle unpack `Job` (OLM auto-generates `Job` into `openshift-marketplace` namespace), with the error that the Pod cannot be scheduled due to the unmatched node affinity.

```
 Warning  FailedScheduling                  5m55s  default-scheduler          0/11 nodes are available: 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 3 node(s) had untolerated taint {node.ocs.openshift.io/storage: true}, 5 node(s) didn't match Pod's node affinity/selector. preemption: 0/11 nodes are available: 11 Preemption is not helpful for scheduling. Warning  FailedScheduling                  50s    default-scheduler          0/11 nodes are available: 3 node(s) had untolerated taint {node-role.kubernetes.io/master: }, 3 node(s) had untolerated taint {node.ocs.openshift.io/storage: true}, 5 node(s) didn't match Pod's node affinity/selector. preemption: 0/11 nodes are available: 11 Preemption is not helpful for scheduling. Normal   ArchAwarePredicateSet             5m56s  multiarch-tuning-operator  Set the supported architectures to {amd64} Normal   ArchAwareSchedGateRemovalSuccess  5m56s  multiarch-tuning-operator  Successfully removed the multiarch.openshift.io/scheduling-gate scheduling gate
```

Image

Image

So it seems based on this event:

```
Set the supported architectures to {amd64}
```

That looks multiarch-tuning-operator is defaulting the Pod arch to be amd64, despite running on s390x architecture, since I see `nodeAffinity` being auto-injected:

```yaml
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
          - matchExpressions:
              - key: kubernetes.io/arch
                operator: In
                values:
                  - amd64
```

aleskandro:
Bundle images are typically shipped as single-architecture images, as they do not contain any architecture-specific content. OLM uses these images as container images with an external entrypoint binary, usually initialized via a volume and an init container.

Previously, the registry inspector logic was partially updated to treat bundle images as supporting all architectures, ensuring they do not interfere with scheduling predicate evaluation. However, this logic relied on a limited set of labels to detect bundle images, which caused some valid bundles to be missed.

This commit expands the set of recognized labels used to detect bundle images. If any of the known labels are present in the image config metadata, the image is now treated as a bundle and assumed to support all architectures. This ensures consistent behavior during installation of operators in clusters that do not run amd64 nodes.

Other container images in the same OLM pod will still be used to restrict the supported architecture set if necessary.

Note: There appears to be no reliable way to detect a bundle image solely from image metadata, apart from these labels. As such, this detection continues to rely on label presence. Also, we have no e2e tests in place for this as they should rely on external content we cannot control.

Closes #645

### DIFF
--- a/pkg/image/inspector.go
+++ b/pkg/image/inspector.go
@@ -33,6 +33,7 @@ import (
 	"github.com/containers/image/v5/signature"
 	"github.com/containers/image/v5/types"
 	"github.com/opencontainers/go-digest"
+	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
 
 	"golang.org/x/sys/unix"
 
@@ -40,7 +41,20 @@ import (
 )
 
 const (
-	operatorSDKBuilderBundleAnnotation = "operators.operatorframework.io.metrics.builder"
+	osdkMetricsAnnotation              = "operators.operatorframework.io.metrics.builder"
+	osdkMediaTypeAnnotation            = "operators.operatorframework.io.bundle.mediatype.v1"
+	osdkManifestsAnnotation            = "operators.operatorframework.io.bundle.manifests.v1"
+	osdkBundleMetadataAnnotation       = "operators.operatorframework.io.bundle.metadata.v1"
+	osdkBundlePackageAnnotation        = "operators.operatorframework.io.bundle.package.v1"
+	osdkBundleChannelsAnnotation       = "operators.operatorframework.io.bundle.channels.v1"
+	osdkBundleDefaultChannelAnnotation = "operators.operatorframework.io.bundle.channel.default.v1"
+)
+
+var (
+	// https://github.com/operator-framework/operator-registry/blob/c4b5f1196/docs/design/operator-bundle.md
+	operatorSDKBuilderBundleAnnotationSet = sets.New[string](
+		osdkMetricsAnnotation, osdkMediaTypeAnnotation, osdkManifestsAnnotation, osdkBundleMetadataAnnotation,
+		osdkBundlePackageAnnotation, osdkBundleChannelsAnnotation, osdkBundleDefaultChannelAnnotation)
 )
 
 type registryInspector struct {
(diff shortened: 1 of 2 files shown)


AnnaZivkovic:
Could you link to the documentation that explains where these labels originate and how we can be certain that they are exclusively bundle labels?

pkg/image/inspector.go
```diff
@@ -33,14 +33,28 @@ import (
...
 
 	"golang.org/x/sys/unix"
 
 	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
 )
 
 const (
-	operatorSDKBuilderBundleAnnotation = "operators.operatorframework.io.metrics.builder"
+	osdkMetricsAnnotation              = "operators.operatorframework.io.metrics.builder"
+	osdkMediaTypeAnnotation            = "operators.operatorframework.io.bundle.mediatype.v1"
```
Prashanth684, line 45: i am not sure if any of these are required, so it is good to check for all of them - but maybe we can just check for manifests, metadata and package. Not a big deal in any case.

Prashanth684:
> Could you link to the documentation that explains where these labels originate and how we can be certain that they are exclusively bundle labels?

https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md?#bundle-annotations

aleskandro:
Already added. PTAL

pkg/image/inspector.go
aleskandro: I'd say it's not a big deal as they are a fixed number and the loop is anyway o(n) + o(1) for the set lookup with n the number of labels in the image

PR #647, opened 2025-06-05
NO-JIRA: Fix channel metadata information in the bundle images

aleskandro:
Replace tech-preview with stable in the channels and default channel bundle's metadata fields.


bundle.konflux.Dockerfile
```diff
@@ -17,8 +17,8 @@ LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
 LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=multiarch-tuning-operator
-LABEL operators.operatorframework.io.bundle.channels.v1=tech-preview
-LABEL operators.operatorframework.io.bundle.channel.default.v1=tech-preview
+LABEL operators.operatorframework.io.bundle.channels.v1=stable
+LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
```
Prashanth684, line 21: if we have had tech preview till now - does this mean that upgrades did not work?

aleskandro: They do work anyway. I'm not sure if this value is actually used elsewhere for now

Prashanth684: yeah the subscription is probably what matters...this is just metadata i guess..good to have it fixed.

AnnaZivkovic: I dont think these were actually being used

PR #649, opened 2025-06-09
WIP: MULTIARCH-5419: enoexec-event-daemon, eBPF program, ...

aleskandro:


enoexec-daemon/main.go
```diff
@@ -0,0 +1,135 @@
...
+	// Buffer Size must be a multiple of page size
+	if os.Getpagesize() <= 0 {
+		must(fmt.Errorf("invalid page size"), "failed to get page size")
+	}
+	pageSize := uint32(os.Getpagesize()) // [bytes]
+	// The payload is 24 bytes. Other 8 bytes are used for the header.
+	// 32 [bytes/event].
+	// See https://github.com/outrigger-project/multiarch-tuning-operator/blob/eabed5c4e54/enhancements/MTO-0004-enoexec-monitoring.md
+	maxEvents := 256
+	bufferSize := pageSize * uint32(math.Ceil(float64(maxEvents*32)/float64(pageSize)))
```
Prashanth684, line 79: if pagesize > (maxEvents*32) then use pagesize or (maxevents*32), that's all right? and even in that case if the requested is less than page size, won't it by default allocate a page?

enoexec-daemon/internal/tracepoint/tracepoint_program.go
```diff
@@ -0,0 +1,171 @@
...
+	//              |      +---------+          |
+	//              |      | process |          |
+	//              |     _| pid=42  |_         |
+	//         __(fork) _/ | tgid=42 | \_ (new thread) _
+	//        /     |      +---------+          |       \
+	//+---------+   |                           |    +---------+
+	//| process |   |                           |    | process |
+	//| pid=43  |   |                           |    | pid=44  |
+	//| tgid=43 |   |                           |    | tgid=42 |
+	//+---------+   |                           |    +---------+
```
Prashanth684, line 71: just to refresh my memory on this since it's been a while - all threads share same tgid, but different pid? and all process have different pid and tgid?

enoexec-daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,212 @@
...
+		"real_parent_tgid", realParentTGID, "current_task_tgid", currentTaskTGID)
+	for _, pid := range []int32{currentTaskTGID, realParentTGID} {
+		podUUID, containerUUID, err := getPodContainerUUIDFor(pid)
+		if err != nil {
+			// Log the error and continue processing other pids as this is not a critical error
+			// (e.g., the pid might not exist anymore due to a delay in processing this record)
+			log.V(5).Info("Failed to get pod and container UUIDs for pid", "pid", pid, "error", err)
+			continue
+		}
+		podName, podNamespace, err := getPodNameFromUUID(tp.ctx, podUUID)
```
Prashanth684, line 191: so how does it work if some binary on the host or let's say a systemd process returns an enoexec - maybe some script or binary fails with the same error, this will intercept and process it as it runs on the host? it will return and continue, but would be nice to identify and not return an error to alarm users.

enoexec-daemon/internal/tracepoint/tracepoint_helpers.go
```diff
@@ -0,0 +1,91 @@
...
+
+	runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"
+
+	"github.com/openshift/multiarch-tuning-operator/enoexec-daemon/internal/utils"
+)
+
+// getPodContainerUUIDFor retrieves the pod UUID and container ID for a given process ID (pid).
+// It reads the cgroup file for the process and extracts the pod UUID and container ID from it.
+// The container ID is returned in the format "cri-o://<container_id>".
+// The pod UUID is extracted from the cgroup path and returned as a string.
```
Prashanth684, line 22: ah .. i see that this is creating a dependency on the runtime mentioned in the other proposal by Lin. this should be noted as needing to adapt when supporting distributions that don't use cri-o

enoexec-daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,212 @@
...
+		tp.outputChannel <- evt
+	}
+}
+
+func (tp *Tracepoint) processRecord(record *ringbuf.Record) (*types.ENOEXECInternalEvent, error) {
+	log := logr.FromContextOrDiscard(tp.ctx)
+	if len(record.RawSample) < 8 {
+		return nil, fmt.Errorf("record too short: %d bytes, expected at least 8 bytes", len(record.RawSample))
+	}
+	realParentTGID := int32(binary.LittleEndian.Uint32(record.RawSample[:4]))
```
Prashanth684, line 179: is this conversion consistent for all arches, especially s390x

enoexec-daemon/internal/storage/k8s_enoexecevent.go
```diff
@@ -0,0 +1,151 @@
...
+	// Convert the ENOEXECInternalEvent to a multiarchv1beta1.ENOExecEvent.
+	// This is a placeholder for actual conversion logic.
+	// generate a unique name for the event with uuid
+	uid, err := uuid.NewUUID()
+	if err != nil {
+		// Handle error (e.g., log it, return an error, etc.)
+		return nil, err
+	}
+	name := uid.String()
+	if len(name) > 63 {
```
Prashanth684, line 123: given that the UUID is a 16 byte array: https://pkg.go.dev/github.com/google/uuid#section-readme this can never happen?

enoexec-daemon/main.go
aleskandro: I'm currently ignoring these comments, as that snippet is from the poc and I'm still working on the pr, looking on whether there is a way to split it or just refine what's here and do other tasks in follow up PRs.

enoexec-daemon/internal/tracepoint/tracepoint.go
aleskandro: if the process is not running in a pod, we should ignore the event.

PR #650, opened 2025-06-10
MULTIARCH-5421: Lifecycle of the exec format error plugin

AnnaZivkovic:
depends on https://github.com/outrigger-project/multiarch-tuning-operator/pull/648, https://github.com/outrigger-project/multia

When the `execFormatErrorMonitor` plugin is enabled via the ClusterPodPlacementConfig, the operator now deploys and manages the necessary components to monitor, handle, and report on executable format errors.

Key Changes:

- ENOEXEC Controller: A new controller is enabled to manage the lifecycle of the custom resource `execFormatErrorMonitor` 
- DaemonSet and Deployment: 
     - ` enoexec-event-daemon` (DaemonSet): Deployed on each node. This DaemonSet runs as a privileged container in order to load an eBPF program that detects ENOEXEC errors and creates an ENoExecEvent custom resource
     - ` enoexec-event-handler` (Deployment): Consumes ENoExecEvent resources and creates corresponding Kubernetes events on the affected pods.
- RBAC Implementation: Creates dedicated ServiceAccounts for the daemon and the handler. Implements the necessary ClusterRoles, Roles, ClusterRoleBindings, and RoleBindings to grant the precise permissions required for each component, as outlined in the design. This includes permissions for the daemon to create ENoExecEvent resources and for the handler to read them and update pods cluster-wide.

Main Function Flags: The operator's main.go has been updated to include the flag `enable-enoexec-event-controllers` to enable and configure the new controller. 

Known Issues: A race condition can occur when deleting resources. If the Enoexexc CR still exists and the objects get deleted

### DIFF
--- a/cmd/main-binary/main.go
+++ b/cmd/main-binary/main.go
@@ -126,6 +126,12 @@ func main() {
 			},
 		}
 	}
+	if enableENoExecEventControllers {
+		leaderID = fmt.Sprintf("enoexecevent-controllers-%s", leaderID)
+		cacheOpts.DefaultNamespaces = map[string]cache.Config{
+			utils.Namespace(): {},
+		}
+	}
 
 	// Rapid Reset CVEs. For more information see:
 	// - https://github.com/advisories/GHSA-qppj-fm5r-hxr3
@@ -272,17 +278,17 @@ func RunENoExecEventControllers(mgr ctrl.Manager) {
 		mgr.GetClient(),
 		clientset,
 		mgr.GetScheme(),
-		mgr.GetEventRecorderFor("enoexecevent-controller"),
+		mgr.GetEventRecorderFor(utils.EnoexecControllerName),
 	).SetupWithManager(mgr), unableToCreateController, controllerKey, "ENoExecEventController")
 }
 
 func validateFlags() error {
-	if !enableOperator && !enableClusterPodPlacementConfigOperandControllers && !enableClusterPodPlacementConfigOperandWebHook {
-		return errors.New("at least one of the following flags must be set: --enable-operator, --enable-ppc-controllers, --enable-ppc-webhook")
+	if !enableOperator && !enableClusterPodPlacementConfigOperandControllers && !enableClusterPodPlacementConfigOperandWebHook && !enableENoExecEventControllers {
+		return errors.New("at least one of the following flags must be set: --enable-operator, --enable-ppc-controllers, --enable-ppc-webhook, --enable-enoexec-event-controllers")
 	}
 	// no more than one of the flags can be set
-	if btoi(enableOperator)+btoi(enableClusterPodPlacementConfigOperandControllers)+btoi(enableClusterPodPlacementConfigOperandWebHook) > 1 {
-		return errors.New("only one of the following flags can be set: --enable-operator, --enable-ppc-controllers, --enable-ppc-webhook")
+	if btoi(enableOperator)+btoi(enableClusterPodPlacementConfigOperandControllers)+btoi(enableClusterPodPlacementConfigOperandWebHook)+btoi(enableENoExecEventControllers) > 1 {
+		return errors.New("only one of the following flags can be set: --enable-operator, --enable-ppc-controllers, --enable-ppc-webhook, --enable-enoexec-event-controllers")
 	}
 	return nil
 }
--- a/controllers/operator/clusterpodplacementconfig_controller.go
+++ b/controllers/operator/clusterpodplacementconfig_controller.go
@@ -79,6 +79,7 @@ const (
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;update;patch;create;delete
 //+kubebuilder:rbac:groups=core,resources=services/status,verbs=get
 //+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
 
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;update;patch;create;delete
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts/status,verbs=get
--- a/controllers/operator/clusterpodplacementconfig_controller.go
+++ b/controllers/operator/clusterpodplacementconfig_controller.go
@@ -249,6 +249,12 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 	clusterPodPlacementConfig *multiarchv1beta1.ClusterPodPlacementConfig) error {
 	// The ClusterPodPlacementConfig is being deleted, cleanup the resources
 	log := ctrllog.FromContext(ctx).WithValues("operation", "handleDelete")
+
+	err := r.handleEnoexecDelete(ctx)
+	if err != nil {
+		return err
+	}
+
 	// The error by the updateStatus function, if any, is ignored, as the deletion should always proceed.
 	// We execute the update here because this function returns multiple times before the whole deletion process is completed.
 	// Executing it here ensures that the conditions are updated throughout the deletion process.
@@ -285,7 +291,7 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 		log.Error(err, "Unable to delete resources")
 		return err
 	}
-	_, err := r.ClientSet.CoreV1().Services(utils.Namespace()).Get(ctx, utils.PodPlacementWebhookName, metav1.GetOptions{})
+	_, err = r.ClientSet.CoreV1().Services(utils.Namespace()).Get(ctx, utils.PodPlacementWebhookName, metav1.GetOptions{})
 	// We look for the webhook service to ensure that the webhook has stopped communicating with the API server.
 	// If the error is nil, the service was found. If the error is not nil and the error is not NotFound, some other error occurred.
 	// In both the cases we return an error, to requeue the request and ensure no race conditions between the verification of the
@@ -401,6 +407,69 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 	return err
 }
 
+func (r *ClusterPodPlacementConfigReconciler) handleEnoexecDelete(ctx context.Context) error {
+	log := ctrllog.FromContext(ctx)
+	log.Info("Disabling ENoExecEvent Controller")
+	objsToDelete := []utils.ToDeleteRef{
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().Roles(utils.Namespace()),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().RoleBindings(utils.Namespace()),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().ClusterRoles(),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().ClusterRoleBindings(),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().Roles(utils.Namespace()),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().RoleBindings(utils.Namespace()),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().ClusterRoles(),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.RbacV1().ClusterRoleBindings(),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.CoreV1().ServiceAccounts(utils.Namespace()),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.CoreV1().ServiceAccounts(utils.Namespace()),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.AppsV1().Deployments(utils.Namespace()),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.AppsV1().DaemonSets(utils.Namespace()),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+	}
+
+	log.Info("Deleting the ENoExecEvent resources")
+	// NOTE: err aggregates non-nil errors, excluding NotFound errors
+	if err := utils.DeleteResources(ctx, objsToDelete); err != nil {
+		log.Error(err, "Unable to delete resources")
+		return err
+	}
+	return nil
+}
+
 // reconcile reconciles the ClusterPodPlacementConfig operand's resources.
 func (r *ClusterPodPlacementConfigReconciler) reconcile(ctx context.Context, clusterPodPlacementConfig *multiarchv1beta1.ClusterPodPlacementConfig) error {
 	log := ctrllog.FromContext(ctx)
@@ -412,53 +481,26 @@ func (r *ClusterPodPlacementConfigReconciler) reconcile(ctx context.Context, clu
 		log.Error(err, "Unable to ensure namespace labels")
 		return errorutils.NewAggregate([]error{err, r.updateStatus(ctx, clusterPodPlacementConfig)})
 	}
-	requiredSCCHostmountAnyUID, seLinuxOptionsType, err := r.getCorrectHostmountAnyUIDSCC(ctx)
+	clusterPodPlacementConfigObjects, err := r.buildPodPlacementConfigObjects(clusterPodPlacementConfig, ctx)
 	if err != nil {
-		log.Error(err, "Unable to set correct hostmount SCC", "requiredSCCHostmoundAnyUID", requiredSCCHostmountAnyUID)
-		return errorutils.NewAggregate([]error{err, r.updateStatus(ctx, clusterPodPlacementConfig)})
+		return err
 	}
-	objects := []client.Object{
-		// The finalizer will not affect the reconciliation of ReplicaSets and Pods
-		// when updates to the ClusterPodPlacementConfig are made.
-		buildService(utils.PodPlacementControllerName),
-		buildService(utils.PodPlacementWebhookName),
-		buildClusterRoleController(), buildClusterRoleWebhook(), buildRoleController(),
-		buildServiceAccount(utils.PodPlacementWebhookName), buildServiceAccount(utils.PodPlacementControllerName),
-		buildClusterRoleBinding(utils.PodPlacementControllerName, rbacv1.RoleRef{
-			APIGroup: rbacv1.GroupName,
-			Kind:     clusterRoleKind,
-			Name:     utils.PodPlacementControllerName,
-		}, []rbacv1.Subject{
-			{
-				Kind:      serviceAccountKind,
-				Name:      utils.PodPlacementControllerName,
-				Namespace: utils.Namespace(),
-			},
-		}),
-		buildRoleBinding(utils.PodPlacementControllerName, rbacv1.RoleRef{
-			APIGroup: rbacv1.GroupName,
-			Kind:     roleKind,
-			Name:     utils.PodPlacementControllerName,
-		}, []rbacv1.Subject{
-			{
-				Kind: serviceAccountKind,
-				Name: utils.PodPlacementControllerName,
-			},
-		}),
-		buildClusterRoleBinding(utils.PodPlacementWebhookName, rbacv1.RoleRef{
-			APIGroup: rbacv1.GroupName,
-			Kind:     clusterRoleKind,
-			Name:     utils.PodPlacementWebhookName,
-		}, []rbacv1.Subject{
-			{
-				Kind:      serviceAccountKind,
-				Name:      utils.PodPlacementWebhookName,
-				Namespace: utils.Namespace(),
-			},
-		}),
-		buildControllerDeployment(clusterPodPlacementConfig, requiredSCCHostmountAnyUID, seLinuxOptionsType),
-		buildWebhookDeployment(clusterPodPlacementConfig),
+
+	execFormatErrorObjects := []client.Object{}
+	if clusterPodPlacementConfig.Spec.Plugins != nil && clusterPodPlacementConfig.Spec.Plugins.ExecFormatErrorMonitor != nil && clusterPodPlacementConfig.Spec.Plugins.ExecFormatErrorMonitor.IsEnabled() {
+		execFormatErrorObjects, err = r.buildENoExecEventObjects(ctx, clusterPodPlacementConfig)
+		if err != nil {
+			return err
+		}
+	} else {
+		err := r.handleEnoexecDelete(ctx)
+		if err != nil {
+			return err
+		}
 	}
+
+	objects := append(clusterPodPlacementConfigObjects, execFormatErrorObjects...)
+
 	// We ensure the MutatingWebHookConfiguration is created and present only if the operand is ready to serve the admission request and add/remove the scheduling gate.
 	shouldEnsureMWC := clusterPodPlacementConfig.Status.CanDeployMutatingWebhook()
 	shouldDeleteMWC := !shouldEnsureMWC && !clusterPodPlacementConfig.Status.IsMutatingWebhookConfigurationNotAvailable()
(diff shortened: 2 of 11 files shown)


enoexec-daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,212 @@
...
+
+func NewTracepoint(ctx context.Context, eventsMaxEntries uint32) (*Tracepoint, chan *types.ENOEXECInternalEvent, error) {
+	tp := &Tracepoint{
+		eventsMaxEntries: eventsMaxEntries,
+		ctx:              ctx,
+		progSpec: &ebpf.ProgramSpec{
+			Name:     "multiarch_tuning_enoexec_tracepoint",
+			Type:     ebpf.TracePoint,
+			AttachTo: "syscalls:sys_enter_execve",
+			License:  "GPL",
```
prb112: total curiosity question - would this be apache-v2.0 since the repo uses this license?

enoexec-daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,212 @@
...
+	events           *ebpf.Map
+	tgidOffset       *int32
+	realParentOffset *int32
+	prog             *ebpf.Program
+	link             link.Link
+	outputChannel    chan *types.ENOEXECInternalEvent
+	eventsMaxEntries uint32
+}
+
+func NewTracepoint(ctx context.Context, eventsMaxEntries uint32) (*Tracepoint, chan *types.ENOEXECInternalEvent, error) {
```
prb112: this is an excellent approach

enoexec-daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+}
+
+func (tp *Tracepoint) processRecord(record *ringbuf.Record) (*types.ENOEXECInternalEvent, error) {
+	log := logr.FromContextOrDiscard(tp.ctx)
+	if len(record.RawSample) < 8 {
+		return nil, fmt.Errorf("record too short: %d bytes, expected at least 8 bytes", len(record.RawSample))
+	}
+	realParentTGID := int32(binary.LittleEndian.Uint32(record.RawSample[:4]))
+	currentTaskTGID := int32(binary.LittleEndian.Uint32(record.RawSample[4:]))
```
prb112: You may want to get the s390x team to attempt this code. Dominik W might want to look at it to see if this limits the code base to LE processors.

enoexec-daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,212 @@
...
+			log.V(5).Info("Failed to get pod name from UUID", "pod_uuid", podUUID)
+			continue
+		}
+		log.Info("Found pod/container UUIDs in record", "pod_name", podName,
+			"pod_namespace", podNamespace, "container_id", containerUUID)
+		return &types.ENOEXECInternalEvent{
+			PodName:      podName,
+			PodNamespace: podNamespace,
+			ContainerID:  containerUUID,
+			// Binary:      "", // TODO: get binary name from record
```
prb112: oh - the binary would be super helpful

enoexec-daemon/internal/tracepoint/tracepoint_helpers.go
```diff
@@ -0,0 +1,91 @@
...
+	return podUID, containerID, nil
+}
+
+// getPodNameFromUUID retrieves the pod name and namespace from the CRI-O runtime
+// using the pod UUID. It connects to the CRI-O socket and lists all pod sandboxes,
+// searching for the pod with the given UUID. If found, it returns the pod name and namespace.
+// If the pod is not found, it returns an empty string for both name and namespace.
+// It returns an error if the connection to the CRI-O socket fails or if other critical operations fail.
+// If the pod is not found, it returns empty strings for both name and namespace without an error.
+func getPodNameFromUUID(ctx context.Context, uid string) (string, string, error) {
```
prb112: fascinating...

prb112:
This feature has wide impact, so exciting to see the code. Great Job.

controllers/enoexecevent/handler/enoexecevent_controller.go
```diff
@@ -39,17 +41,22 @@ import (
...
 type Reconciler struct {
 	client.Client
-	clientSet *kubernetes.Clientset
-	Scheme    *runtime.Scheme
-	recorder  record.EventRecorder
+	clientSet     *kubernetes.Clientset
+	dynamicClient *dynamic.DynamicClient
+	Scheme        *runtime.Scheme
+	recorder      record.EventRecorder
+	events        events.Recorder
```
aleskandro: what's the purpose of the two recorders?

AnnaZivkovic: that is left over code. Ill remove it

cmd/main-binary/main.go
```diff
@@ -126,6 +126,12 @@ func main() {
 			},
 		}
 	}
+	if enableENoExecEventControllers {
+		leaderID = fmt.Sprintf("enoexecevent-controllers-%s", leaderID)
+		cacheOpts.DefaultNamespaces = map[string]cache.Config{
+			"openshift-multiarch-tuning-operator": {},
```
lwan-wanglin:
```suggestion
			utils.Namespace(): {},
```
Can this be changed to avoid hardcoding the name?

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -122,7 +125,7 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, nil
 	case !clusterPodPlacementConfig.DeletionTimestamp.IsZero():
 		// Only execute deletion if the object is being deleted and the finalizer is present
-		return ctrl.Result{}, r.handleDelete(ctx, clusterPodPlacementConfig)
+		return ctrl.Result{}, errors.Join(r.handleDelete(ctx, clusterPodPlacementConfig), r.handleEnoexecDelete(ctx))
```
lwan-wanglin: I see you've already added `r.handleEnoexecDelete(ctx)` to `r.handleDelete()`. Do we still need to run it here?

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -400,6 +409,69 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
...
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.CoreV1().ServiceAccounts(utils.Namespace()),
+			ObjName:               utils.EnoexecDaemonSet,
+		},
+		{
+			NamespacedTypedClient: r.ClientSet.AppsV1().Deployments(utils.Namespace()),
+			ObjName:               utils.EnoexecControllerName,
+		},
+		{
```
aleskandro, line 458: If we delete all at once, there may be leftover ENoExecEvent objects in the Cluster.
We should basically ensure that the Daemonset is removed first and the reconcilers are removed only after all the ENoExecEvents are processed and deleted.

I'll open a story as a follow up for this PR.

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -411,53 +483,35 @@ func (r *ClusterPodPlacementConfigReconciler) reconcile(ctx context.Context, clu
...
-				Name:      utils.PodPlacementWebhookName,
-				Namespace: utils.Namespace(),
-			},
-		}),
-		buildControllerDeployment(clusterPodPlacementConfig, requiredSCCHostmountAnyUID, seLinuxOptionsType),
-		buildWebhookDeployment(clusterPodPlacementConfig),
+
+	execFormatErrorObjects := []client.Object{}
+	if clusterPodPlacementConfig.Spec.Plugins != nil && clusterPodPlacementConfig.Spec.Plugins.ExecFormatErrorMonitor != nil {
+		if clusterPodPlacementConfig.Spec.Plugins.ExecFormatErrorMonitor.IsEnabled() {
```
aleskandro:
```suggestion
	if clusterPodPlacementConfig.Spec.Plugins != nil && clusterPodPlacementConfig.Spec.Plugins.ExecFormatErrorMonitor != nil && clusterPodPlacementConfig.Spec.Plugins.ExecFormatErrorMonitor.IsEnabled() {
```

aleskandro: In the current implementation, you first check that `Plugins` and the plugin's stanza are not nil pointers. Then, in an inner if block, you verify whether the plugin is enabled to create the objects. If not, you delete the objects. However, if the user deletes the entire stanza without setting enabled=false first, do the objects remain in the cluster?

We should check all the three conditions together for creating the objects. Then, the else statement will work even when a user entirely deletes the plugins key.

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -411,53 +483,35 @@ func (r *ClusterPodPlacementConfigReconciler) reconcile(ctx context.Context, clu
...
+			execFormatErrorObjects, err = r.buildENoExecEventObjects(ctx)
+			if err != nil {
+				return err
+			}
+		} else {
+			// Check if the Deployment exists
+			exists, err := r.doesDeploymentExist(utils.EnoexecControllerName, ctx, log)
```
aleskandro: What if we deleted the enoexeccontroller, but the daemonset or a service account, for some reason (temporary failure in the API server when sending the delete request), is still there?

What happens if we just call handleEnoexecDelete? The removal requests will be sent but no error should arise if the objects do not exist.

pkg/testing/framework/cluster_pod_placement_config.go
```diff
@@ -99,3 +126,20 @@ func ValidateCreation(cl client.Client, ctx context.Context) func(gomega.Gomega)
 		)
 	}
 }
+
+// getObjectsForMode returns the list of objects to be validated based on the provided mode.
+func getObjectsForMode(mode []int) []client.Object {
```
aleskandro:
```suggestion
func getObjectsFor(what []PluginObjectsSet) []client.Object {
```

I would use `PluginObjectsSet` instead of `mode`.

The calls to the objects would seem more meaningful to me:

```
getObjectsFor(ENoExecPlugin)
getObjectsFor(MainPlugin)
getObjectsFor(ENoExecPlugin, MainPlugin)
```

Where MainPlugin can be a constant of type `PluginObjectsSet = int` and defined as you did in the enumeration above.

pkg/testing/framework/cluster_pod_placement_config.go
```diff
@@ -23,6 +23,16 @@ type ConditionTypeStatusTuple struct {
...
 }
+type ValidationMode int
+
+const (
+	// ValidateDefault checks only the core pod placement resources.
+	ValidateDefault ValidationMode = iota
+	// ValidateEnoexecOnly checks only the ENoExecEvent resources.
+	ValidateEnoexecOnly
+	// ValidateAll checks both the core and ENoExecEvent resources.
+	ValidateAll
```
aleskandro: I wouldn't define an explicit ValidateAll. I'm ok with having that default to reduce the needed changes in the other tests, but wouldn't 'advertise' this to a future developer working on a third plugin as I  think it might not scale well.

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,174 @@
...
+						{
+							Name:            name,
+							Image:           utils.Image(),
+							ImagePullPolicy: corev1.PullIfNotPresent,
+
+							Args: args,
+							Command: []string{
+								"/enoexec-daemon",
+							},
+							Env: []corev1.EnvVar{},
```
lwan-wanglin: @AnnaZivkovic @aleskandro I'm not familiar with this. Do we need the Downward API to get `nodeName` and `namespace` in the pod env？

controllers/operator/clusterpodplacementconfig_controller.go
AnnaZivkovic: I dont think I have a definite answer to that. My though is that the plugin controls weather or not to enable the Enoexec. So if the cppc is deleted the Enoexec should be as well.  Same for creation, we must first enable the Enoexec in the cppc for the objects to get created.

controllers/operator/clusterpodplacementconfig_controller.go
AnnaZivkovic: @aleskandro would that be done though finalizers?

aleskandro: Yes.

controllers/operator/enoexecevent_objects.go
aleskandro: Hi @lwan-wanglin yes we do. Thanks for raising this.

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,174 @@
...
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: utils.NewPtr(true),
+								Privileged:               utils.NewPtr(true),
+								// The bpftrace tool requires root privileges to interact with the kernel.
+								RunAsUser:              utils.NewPtr(int64(0)),
+								ReadOnlyRootFilesystem: utils.NewPtr(true),
+							},
+							VolumeMounts: []corev1.VolumeMount{
```
aleskandro, line 157: We'd also have /var/run/crio mounted i think.

AnnaZivkovic: Looks like if we want to mount the socket we need /var/run/crio/crio.sock

AnnaZivkovic:
This pr depends on https://github.com/outrigger-project/multiarch-tuning-operator/pull/662

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: I think what Lin is saying is correct. Why are we calling handleEnoexecDelete twice?

aleskandro: handleDelete -> handleEnoexecDelete (called inside handleDelete) -> other deletion stuff that handleDeletedoes -> handleEnoexecDelte (called as the second parameter to errors.Join())

AnnaZivkovic:
Running into this error in the e2e
```
Message:         Failed to create ClusterRole.rbac.authorization.k8s.io/enoexec-event-handler-controller: clusterroles.rbac.authorization.k8s.io "enoexec-event-handler-controller" is forbidden: user "system:serviceaccount:openshift-multiarch-tuning-operator:multiarch-tuning-operator-controller-manager" (groups=["system:serviceaccounts" "system:serviceaccounts:openshift-multiarch-tuning-operator" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
{APIGroups:[""], Resources:["configmaps"], Verbs:["update" "patch" "create" "delete"]}
{APIGroups:["coordination.k8s.io"], Resources:["leases"], Verbs:["list" "watch" "get" "update" "patch" "create" "delete"]}
```

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,214 @@
...
+					HostPID:                      true,
+					Containers: []corev1.Container{
+						{
+							Name:            name,
+							Image:           utils.Image(),
+							ImagePullPolicy: corev1.PullIfNotPresent,
+
+							Args: args,
+							Command: []string{
+								"/enoexec-daemon",
```
aleskandro, line 127: We should add the following parameter too:

```
-initial-log-level="${the log level}"
```

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,214 @@
...
+									ValueFrom: &corev1.EnvVarSource{
+										FieldRef: &corev1.ObjectFieldSelector{
+											APIVersion: "v1",
+											FieldPath:  "spec.nodeName",
+										},
+									},
+								},
+							},
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: utils.NewPtr(true),
```
aleskandro:
```suggestion
								AllowPrivilegeEscalation: utils.NewPtr(true),
```


Do we need this?

controllers/operator/clusterpodplacementconfig_controller.go
AnnaZivkovic: Good catch. Ill remove the first call

controllers/operator/enoexecevent_objects.go
AnnaZivkovic: I added the loglevel from he cppc 
`fmt.Sprintf("--initial-log-level=%d",logVerbosity),`

controllers/operator/enoexecevent_objects.go
AnnaZivkovic: No we do not need this.

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -527,6 +572,169 @@ func (r *ClusterPodPlacementConfigReconciler) updateStatus(ctx context.Context,
...
+func (r *ClusterPodPlacementConfigReconciler) buildENoExecEventObjects(ctx context.Context, clusterPodPlacementConfig *multiarchv1beta1.ClusterPodPlacementConfig) ([]client.Object, error) {
+	log := ctrllog.FromContext(ctx)
+	// Check if the Deployment exists
+	exists, err := r.doesDeploymentExist(utils.EnoexecControllerName, ctx, log)
+	if err != nil {
+		return []client.Object{}, err
+	}
+	if exists {
+		return []client.Object{}, nil
+	}
```
aleskandro: If the deployment is created and other objects are not created due to some temporary error, this verification here leads to a non-reliable reconciliation. 

If we recreate the objects, and they are the same, K8S should understand it and do nothing as  they are not different from what already exists.

```suggestion
```

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -527,6 +572,169 @@ func (r *ClusterPodPlacementConfigReconciler) updateStatus(ctx context.Context,
...
+		if apierrors.IsNotFound(err) {
+			log.Info("Deployment not found", "name", deploymentName, "namespace", utils.Namespace())
+			return false, nil
+		}
+		log.Error(err, "Unable to fetch the deployment")
+		return false, err
+	}
+
+	return true, nil
+}
```
aleskandro:
```suggestion
```

We should not need this function.

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,216 @@
...
+					Containers: []corev1.Container{
+						{
+							Name:            name,
+							Image:           utils.Image(),
+							ImagePullPolicy: corev1.PullIfNotPresent,
+
+							Args: args,
+							Command: []string{
+								"/enoexec-daemon",
+								fmt.Sprintf("--initial-log-level=%d",
```
aleskandro, line 128:
```suggestion
								fmt.Sprintf("--initial-log-level=%d",
```

just to confirm, is the double dash `--` correctly handled?

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,216 @@
...
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+// buildEnoexecDeployment returns a minimal Deployment object matching your YAML
+func buildDeploymentENoExecEvent(logVerbosity int) *appsv1.Deployment {
```
aleskandro:
```suggestion
func buildDeploymentENoExecEventHandler(logVerbosity int) *appsv1.Deployment {
```

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -485,4 +485,138 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+				err = client.Create(ctx, ns)
+				Expect(err).NotTo(HaveOccurred())
+				//nolint:errcheck
+				defer client.Delete(ctx, ns)
+				By("Creating a pod designed to trigger an ENOEXEC error")
+				ps := NewPodSpec().
+					WithContainersImages(helloOpenshiftPublicMultiarchImage).
+					WithNodeSelectors(map[string]string{utils.ArchLabel: utils.ArchitectureAmd64}).
+					WithCommand([]string{"/bin/bash"}).
+					WithArgs("-c", `#!/bin/bash
```
aleskandro: Can we move this string to a global constant outside this function or change  it such that it doesn't impact readability due to the missing indentation?

pkg/testing/builder/pod_spec.go
```diff
@@ -49,6 +49,22 @@ func (ps *PodSpecBuilder) WithContainers(containers ...*v1.Container) *PodSpecBu
 	return ps
 }
 
+func (ps *PodSpecBuilder) WithCommand(command []string) *PodSpecBuilder {
```
aleskandro:
```suggestion
func (ps *PodSpecBuilder) WithCommand(command ...string) *PodSpecBuilder {
```

can we change this to use a variadic parameter like WithArgs?

pkg/testing/framework/cluster_pod_placement_config.go
```diff
@@ -99,3 +130,18 @@ func ValidateCreation(cl client.Client, ctx context.Context) func(gomega.Gomega)
 		)
 	}
 }
+
+// getObjectsFor returns the list of objects to be validated based on the provided mode.
+func getObjectsFor(pluginObjectsSet []PluginObjectsSet) []client.Object {
```
aleskandro:
```suggestion
func getObjectsFor(pluginObjectsSet ...PluginObjectsSet) []client.Object {
```

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: (An additional update to the cppc, for example changing the log level, would also lead to no reconciliation of the objects here).

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -485,4 +485,138 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+				By("Create an ephemeral namespace for the test pod")
+				ns := framework.NewEphemeralNamespace()
+				err = client.Create(ctx, ns)
+				Expect(err).NotTo(HaveOccurred())
+				//nolint:errcheck
+				defer client.Delete(ctx, ns)
+				By("Creating a pod designed to trigger an ENOEXEC error")
+				ps := NewPodSpec().
+					WithContainersImages(helloOpenshiftPublicMultiarchImage).
+					WithNodeSelectors(map[string]string{utils.ArchLabel: utils.ArchitectureAmd64}).
```
aleskandro, line 603: Is it possible to run this test for each architecture available in the cluster?

This can be done by using a DescribeTable node to loop through the 4 architectures (entries) and adding a check at the beginning that:
1. Gets all the nodes and maps them into a set of available architectures
2. Skip the test if the set does not include the current entry's architecture

aleskandro: This way we test that the program works on arm and  amd in CI and, when run on other architectures like P/Z, that it works there too.

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -36,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
 
+	"github.com/go-logr/logr"
```
aleskandro: given the other  comment , this should go an d we'd use ctrllog for most of the cases instead of logr.

controllers/operator/enoexecevent_objects.go
aleskandro: just to confirm, is the double dash -- correctly handled?

controllers/enoexecevent/handler/suite_test.go
```diff
@@ -31,7 +31,6 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
```
aleskandro: do we need the changes to this file?

controllers/operator/enoexecevent_objects.go
AnnaZivkovic: I think it works but I keep getting this error in the daemon set pods, but it seems like an permissions problem?
`failed to IncreaseLevel: invalid increase level, as level "debug" is allowed by increased level, but not by existing core`

pkg/testing/framework/cluster_pod_placement_config.go
AnnaZivkovic: I think it is much cleaner to keep it as a list because of the way they are passed to validateCreation/Deletion 

```
ValidateCreation(cl client.Client, ctx context.Context, pluginObjectsSet ...PluginObjectsSet) func(gomega.Gomega) {
	return func(g gomega.Gomega) {
		ginkgo.By("Verify all objects exist")
		for _, obj := range getObjectsFor(pluginObjectsSet) {
```

Now we pass a list into `getObjectsFor` vs the individual elements

controllers/operator/enoexecevent_objects.go
aleskandro: All right. That error is fixed in https://github.com/outrigger-project/multiarch-tuning-operator/pull/669

Thanks Tori

controllers/operator/enoexecevent_objects.go
```diff
@@ -0,0 +1,216 @@
...
+									Type: utils.NewPtr(corev1.HostPathDirectory),
+								},
+							},
+						},
+						{
+							Name: "crio",
+							VolumeSource: corev1.VolumeSource{
+								HostPath: &corev1.HostPathVolumeSource{
+									Path: "/var/run/crio/crio.sock",
+									Type: utils.NewPtr(corev1.HostPathSocket),
```
aleskandro, line 200: Great :)

aleskandro (commented):
/lgtm

PR #654, opened 2025-06-30
MULTIARCH-5419: Scaffolding for the enoexec-daemon

aleskandro:
This commit implements the initial scaffolding for the enoexec-daemon as an additional gomodule. The CI scripts are updated to run for any additional modules and the Containerfiles now include a build of the enoexec-daemon in the runtime image.

In the future, we may move this binary to a dedicated container image.


enoexec-daemon/internal/types/event.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+// ENOEXECInternalEvent is an event that is emitted when the eBPF tracepoint
+// detects an ENOEXEC error. This event contains information about the pod that
+// encountered the error, and is sent to the internal event bus to be processed by
+// the consumer process running the storage implementation and then adapted into an ENoExecEvent K8S object.
+type ENOEXECInternalEvent struct {
+	PodName      string `yaml:"podName,omitempty"`
```
Prashanth684, line 8: nodeName?

PR #655, opened 2025-06-30
MULTIARCH-5419: Scaffolding for the enoexec-daemon

aleskandro:
This commit implements the initial scaffolding for the enoexec-daemon as an additional gomodule. The CI scripts are updated to run for any additional modules and the Containerfiles now include a build of the enoexec-daemon in the runtime image.

In the future, we may move this binary to a dedicated container image.


enoexec-daemon/internal/types/event.go
```diff
@@ -0,0 +1,12 @@
+package types
+
+// ENOEXECInternalEvent is an event that is emitted when the eBPF tracepoint
+// detects an ENOEXEC error. This event contains information about the pod that
+// encountered the error, and is sent to the internal event bus to be processed by
+// the consumer process running the storage implementation and then adapted into an ENoExecEvent K8S object.
+type ENOEXECInternalEvent struct {
+	PodName      string `yaml:"podName,omitempty"`
```
Prashanth684, line 8: didn't we want to add node name as well?

enoexec-daemon/internal/storage/storage.go
```diff
@@ -0,0 +1,52 @@
+package storage
+
+import (
+	"context"
+
+	"github.com/openshift/multiarch-tuning-operator/enoexec-daemon/internal/types"
```
Prashanth684: needs to be `outrigger-project`?

enoexec-daemon/internal/types/event.go
aleskandro: The node name can be retrieved in many other ways here: we are running this code on that node, there should be at least env variables to have the name of the node where this is running

enoexec-daemon/internal/storage/storage.go
aleskandro: yep... let me send an update.

enoexec-daemon/internal/types/event.go
Prashanth684: yes, but do we want it to be included as part of this event reporting?

aleskandro: not in this type, but yes when we adapt it into https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/apis/multiarch/v1beta1/enoexecevent_types.go#L37. This will happen on another thread.

PR #656, opened 2025-07-09
MULTIARCH-5596: add document for mto installation on crio based kind clusters

lwan-wanglin:
add document for mto installation on crio based kind clusters before we fix the OCP depedences

### DIFF
--- a/docs/support-non-okd.md
+++ b/docs/support-non-okd.md
@@ -1,0 +1,282 @@
+# Install Multiarch Tuning Operator in a kind Cluster with CRI-O
+
+There are several OpenShift-specific dependencies required to run the Multiarch Tuning Operator on non-OpenShift clusters. We already have an enhancement and an epic [MULTIARCH-5324](https://issues.redhat.com/browse/MULTIARCH-5324) planned to remove these dependencies.
+In the meantime, if you want to use the operator before that enhancement is complete, this document provides instructions for installing the Multiarch Tuning Operator on a `kind` cluster that uses `CRI-O` as the container runtime.
+It includes steps for creating the cluster, installing dependencies, and deploying the operator.
+
+## Prerequisites
+
+- Golang
+- Docker (Community Edition)
+- kind
+- Helm
+- kubectl
+
+## Create kind Cluster with CRI-O
+
+The following is an example setup:
+
+### Ensure Kind is installed
+
+```bash
+go install sigs.k8s.io/kind@v0.29.0
+kind version
+```
+more information see [installation-and-usage](https://kind.sigs.k8s.io/#installation-and-usage)
+
+### Create a CRI-O–based node image
+You can follow the official guide [crio-in-kind](https://github.com/cri-o/cri-o/blob/main/tutorials/crio-in-kind.md) to build a node image that uses CRI-O as the container runtime.
+
+### Create a Kind configuration file using the CRI-O–based node image
+To enable image policy and registry access, the operator depends on standard container runtime configuration files, such as:
+`/etc/containers/policy.json`
+`/etc/containers/registries.conf`
+As these files are not always present by default, we recommend mounting the corresponding host paths into the container to provide the necessary configuration context.
+Use the node image created in the previous step to define the cluster configuration. Save the following content into a file named `kind-crio-config.yaml`:
+```yaml
+cat <<EOF > kind-crio-config.yaml
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+  - role: control-plane
+    image: kindnode/crio:v1.30
+    extraMounts:
+    - hostPath: /etc/containers/registries.conf
+      containerPath: /etc/containers/registries.conf
+    - hostPath: /etc/containers/policy.json
+      containerPath: /etc/containers/policy.json
+    extraPortMappings:
+      - containerPort: 6443
+        hostPort: 6443
+        listenAddress: "127.0.0.1"
+    kubeadmConfigPatches:
+      - |
+        kind: InitConfiguration
+        nodeRegistration:
+          criSocket: unix:///var/run/crio/crio.sock
+      - |
+        kind: JoinConfiguration
+        nodeRegistration:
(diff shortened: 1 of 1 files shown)


docs/support-non-okd.md
```diff
@@ -0,0 +1,270 @@
...
+
+Clone the repository and run the following command to install the operator and its operand:
+```bash
+make deploy IMG=registry.ci.openshift.org/origin/multiarch-tuning-operator:main
+
+kubectl annotate ValidatingWebhookConfiguration multiarch-tuning-operator-validating-webhook-configuration \
+cert-manager.io/inject-ca-from=openshift-multiarch-tuning-operator/multiarch-tuning-operator-validating-webhook
+
+kubectl create -f - <<EOF
+apiVersion: multiarch.openshift.io/v1beta1
```
Prashanth684, line 261: hmm..our CRDs still use `multiarch.openshift.io`, but changing it to `multiarch.outrigger.io` is probably a breaking change and then our downstream will differ

docs/support-non-okd.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Create the ConfigMap with the CA bundle:
+```bash
+kubectl create configmap multiarch-tuning-operator-trusted-ca \
+  --from-file=ca-bundle.crt=./ca-certificates.crt \
+  -n openshift-multiarch-tuning-operator
+```
+
+## Add Pull Secret
+The operator is currently hardcoded to watch the `pull-secret` Secret in the `openshift-config` namespace. As a result, if you're running on a non-OpenShift cluster or the namespace does not exist, you need to manually create the `openshift-config` namespace and add a `pull-secret` Secret to it.
```
Prashanth684: ideally, this can be configured by the user to provide their own namespace for the pullsecret [here](https://github.com/outrigger-project/multiarch-tuning-operator/blob/799f91a33ceb2f1b503f97016ce9e44b30322909/config/manager/manager.yaml#L69) through the `global-pul-secret-namespace` option

docs/support-non-okd.md
```diff
@@ -0,0 +1,270 @@
...
+    - mutating-webhook.openshift-multiarch-tuning-operator.svc
+  issuerRef:
+    name: multiarch-tuning-operator-issuer
+    kind: ClusterIssuer
+  commonName: mutating-webhook.openshift-multiarch-tuning-operator.svc
+EOF
+```
+
+## Add CA Bundle ConfigMap (if needed)
+In OpenShift, resources labeled with `config.openshift.io/inject-trusted-cabundle: "true"` are automatically injected with the cluster’s trusted CA bundle.
```
Prashanth684, line 209: this is only if a custom CA/proxy is used, correct?

Prashanth684:
This comment is not related only to this document - but something I just noticed: It would be nice to not have `openshift` in the CRD/namespace names. Now this will cause a deviation from downstream which we have to handle, but something to think about for the future

docs/support-non-okd.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Run below command to create a basic crio kind cluster:
+
+```bash
+kind create cluster --name cri-o --config kind-crio-config.yaml
+```
+
+## Create certificate for operator traffic encryption
+To ensure the operator traffic is secure, we use certificates to encrypt the communication.
+In OpenShift (OCP) clusters, certificate-related resources are automatically managed by the ocp operator. However, in non-OCP environments, you must manage these certificates manually. Also you can use a `cert-manager` operator to assist with certificate provisioning and management.
```
Prashanth684, line 88: you mean service-ca operator?

Prashanth684: OLM manages the certs too for the webhooks

Prashanth684: this also makes me wonder if cert-manager should be the default upstream way of handling certs rather than the OCP service-ca operator (in a non OLM case)

lwan-wanglin: In my opinion, using cert-manager is a good approach for managing certificates, including creation and rotation.
We’ve also documented it as an option in the enhancement proposal.

docs/support-non-okd.md
lwan-wanglin: Probably not — even images from quay.io can't be trusted if the CA bundle is missing.

docs/support-non-okd.md
lwan-wanglin: Ah, right. the current instructions assume you’re using the image with our default settings.
If we want users to modify those settings and build the image themselves, I will update the steps accordingly.

docs/support-non-okd.md
aleskandro: > this also makes me wonder if cert-manager should be the default upstream way of handling certs rather than the OCP service-ca operator (in a non OLM case)

Yeah, I think that's where we will land. This Documentation though is meant to allow using it now, before the changes to improve the deployment in vanilla k8s are implemented.

aleskandro:
> This comment is not related only to this document - but something I just noticed: It would be nice to not have `openshift` in the CRD/namespace names. Now this will cause a deviation from downstream which we have to handle, but something to think about for the future

@lwan-wanglin, have we added anything about this in the  EP?

docs/support-non-okd.md
aleskandro: Yeah, this is only needed if the user makes use of a registry that has no SSL certificate validated through a CA that is verifiable by the root CA certs available by default in the image's OS.

docs/support-non-okd.md
aleskandro: That's a bigger change that we should track as a story in the related epic.

lwan-wanglin:
> have we added anything about this in the EP?
added it to "Open Questions" section.

docs/support-non-okd.md
lwan-wanglin: Hi @Prashanth684, I've updated the content to include the option for configuring a custom namespace.

aleskandro (commented):
/lgtm

PR #657, opened 2025-07-14
MULTIARCH-5419: Implement Kubernetes storage for ENOEXEC events

aleskandro:
This commit introduces K8sENOExecEventStorage as a storage backend that creates ENoExecEvent custom resources in the Kubernetes cluster.

The backend runs in its goroutine and consumes data sent through a channel of ENOEXECInternalEvent objects. The expected writer for this channel is the eBPF tracepoint component.

When an event is received, this storage implementation creates a corresponding ENoExecEvent object in the cluster. To prevent API server and key-value store flooding, especially in scenarios such as CrashLoopBackOff or incorrect binaries retrying in a loop, a token bucket-based rate limiter (golang.org/x/time/rate) is used to throttle requests.

Additionally, each event processing operation is wrapped in a timeout context. This defends against prolonged blocking or denial-of-service (DoS) conditions under high load or malicious input.

Throttling parameters are currently hardcoded and will be refined or made configurable during system fine-tuning.

Note: this is a first implementation of a storage backend and that, as discussed in the EP, we might implement a backend storage that leverages an IPC mechanism to seaparate more the event generation and the creation of the K8S CRs by deploying the two components as a two-containers DaemonSet.

This PR also includes two refactoring commits that simplify the structure of the repo by unifying the enoexec-event daemon code into the main module, still maintaining the possibility of building it as a separate binary. I could not separate that change without disrupting the storage implementation, so I'm proposing it in the same PR.


controllers/enoexecevent/daemon/daemon.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+func RunDaemon(ctx context.Context, cancel context.CancelFunc) error {
+	log, err := logr.FromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get logger from context: %w", err)
+	}
+	ch := make(chan *types.ENOEXECInternalEvent, 256)
+	storageImpl, err := storage.NewK8sENOExecEventStorage(ctx,
+		rate.NewLimiter(5, 10), ch, os.Getenv("NODE_NAME"), os.Getenv("NAMESPACE"), time.Minute,
```
lwan-wanglin, line 24: where do the `NODE_NAME` and `NAMESPACE` variables come from?

aleskandro: They will be provided in the pod specs by the operator via downwardAPI.

PR #658, opened 2025-07-14
MULTIARCH-5419: Add scaffolding for eBPF tracepoint to detect ENOEXEC errors in pods

aleskandro:
This PR introduces the initial scaffolding for an eBPF tracepoint used to capture ENOEXEC errors occurring in containers managed by Kubernetes.

The tracepoint is encapsulated in the internal Tracepoint struct, which provides:

- Initialization of eBPF program specifications

- Setup of the ring buffer for communication between kernel and user space
- Marshaling and unmarshaling of data transmitted through the ring buffer
- Methods to extract additional metadata (e.g., pod name, container UUIDs) in user space for Kubernetes object tracking
- A placeholder (dummy) tracepoint eBPF program — the actual implementation will be proposed in a subsequent PR

Additional notes:

- Endianness validation (e.g., on s390x) will be addressed after the program is functional on Little Endian architectures
- CRI runtime compatibility (other than the current default) is not yet supported — an issue should be opened to track this

Stacked on #657


controllers/enoexecevent/daemon/daemon.go
```diff
@@ -0,0 +1,91 @@
...
+	bufferSize := pageSize * uint32(math.Ceil(float64(maxEvents*32)/float64(pageSize)))
+
+	log.Info("Buffer size calculated", "buffer_size", bufferSize, "page_size", pageSize, "max_events", maxEvents)
+
+	log.Info("Initializing channel")
+	ch := make(chan *types.ENOEXECInternalEvent, maxEvents)
+
+	log.Info("Initializing storage")
+	storageImpl, err := storage.NewK8sENOExecEventStorage(ctx,
+		rate.NewLimiter(5, 10), ch, os.Getenv("NODE_NAME"), os.Getenv("NAMESPACE"), time.Minute,
```
Prashanth684, line 49: so after the 1 min timeout expires, events are dropped?

controllers/enoexecevent/daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,220 @@
...
+	link     link.Link
+
+	tgidOffset       *int32
+	realParentOffset *int32
+	bufferSize       uint32 // Size of the ring buffer in bytes
+
+	ch chan *types.ENOEXECInternalEvent
+}
+
+func NewTracepoint(ctx context.Context, ch chan *types.ENOEXECInternalEvent, bufferSize uint32) (*Tracepoint, error) {
```
Prashanth684, line 37: isn't this dependent on #661 ? should #661 be used as the main PR?

controllers/enoexecevent/daemon/daemon.go
aleskandro: Yes I think a timeout for that processing is needed to avoid memory leaks.

About the values set, they are mostly draft for now. The default value will be changed in a fine-tuning PR once the feature is testable and we can infer the values.

In the future, we can also have these values editable from the ClusterPodPlacementConfig object.

controllers/enoexecevent/daemon/internal/tracepoint/tracepoint.go
aleskandro: #661 is built on top of #658 and #657.

The order of implementation is:
- #657: storage implementation. That is the consumer of the channel where the internal event is sent. The role of this consumer is to create an object in the k8s API (without knowing about the ebpf nuances)
- #658: ebpf scaffolding. It implements the consumer of the ebpf event at user-space. Its role is to consume an event submitted by the ebpf program, enrich it with additional information required in the final k8s object, and send this information over to the channel where the storage consumer at #657 is listening for new events
- #661: ebpf program: it implements the ebpf program: the producer.


Alternatively, we can merge all at once (but #661 is still wip).

I have split the whole work in 3 PRs (others will come for testing and other refinements) to ease the reviews. The 3 PRs are already big enough.

aleskandro: So you'd not review this, but #657 first.

controllers/enoexecevent/daemon/internal/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,220 @@
...
+			continue
+		}
+		podName, podNamespace, err := getPodNameFromUUID(tp.ctx, podUUID)
+		if err != nil {
+			// Errors from getPodNameFromUUID are critical if err is not nil, as it indicates a failure to connect to the CRI-O runtime
+			log.V(5).Info("Failed to get pod name from UUID", "pod_uuid", podUUID, "error", err)
+			return nil, fmt.Errorf("failed to get pod name from UUID %s: %w", podUUID, err)
+		}
+		if podName == "" {
+			// If podName is empty, it means the pod was not found in the CRI-O runtime, that is not a critical error
```
Prashanth684, line 191: so in this case the event was captured, but the pod was removed - i guess this can happen when we wait in the queue for long?

Prashanth684: would also be nice to capture these as metrics

PR #661, opened 2025-07-15
MULTIARCH-5419: enoexec eBPF program

aleskandro:
This PR implements the eBPF program to catch exec format errors and minor fixes from the initial manual testing.

This is related to, but not blocked by, MULTIARCH-5602.

The command is not being caught because it usually is the name of the process running the command that leads to failure. Therefore, we do not retrieve the failing command name itself, but rather its caller.

Minor fixes and optimizations are also proposed.

### DIFF
--- a/controllers/enoexecevent/daemon/daemon.go
+++ b/controllers/enoexecevent/daemon/daemon.go
@@ -33,11 +33,11 @@ func RunDaemon(ctx context.Context, cancel context.CancelFunc) error {
 	maxEvents := 256
 	// The buffer size has to be a multiple of the page size.
 	// We calculate the required buffer size based on the maximum number of events as
-	// size_max = maxEvents * 32 [bytes].
+	// size_max = maxEvents * 24 [bytes].
 	// We obtain the number of pages required to store the events rounding up the number of pages required
 	// to store size_max bytes: required_pages = Ceil(size_max [bytes] / pageSize [bytes]).
 	// Finally, we multiply required_pages by the page size to get the buffer size.
-	bufferSize := pageSize * uint32(math.Ceil(float64(maxEvents*32)/float64(pageSize)))
+	bufferSize := pageSize * uint32(math.Ceil(float64(maxEvents*24)/float64(pageSize)))
 
 	log.Info("Buffer size calculated", "buffer_size", bufferSize, "page_size", pageSize, "max_events", maxEvents)
 
--- a/controllers/enoexecevent/daemon/daemon.go
+++ b/controllers/enoexecevent/daemon/daemon.go
@@ -3,7 +3,6 @@ package daemon
 import (
 	"context"
 	"fmt"
-	"math"
 	"os"
 	"sync"
 	"time"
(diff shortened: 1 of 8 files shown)


controllers/enoexecevent/daemon/daemon.go
```diff
@@ -21,39 +20,29 @@ func RunDaemon(ctx context.Context, cancel context.CancelFunc) error {
...
-	// We calculate the required buffer size based on the maximum number of events as
-	// size_max = maxEvents * 32 [bytes].
-	// We obtain the number of pages required to store the events rounding up the number of pages required
-	// to store size_max bytes: required_pages = Ceil(size_max [bytes] / pageSize [bytes]).
-	// Finally, we multiply required_pages by the page size to get the buffer size.
-	bufferSize := pageSize * uint32(math.Ceil(float64(maxEvents*32)/float64(pageSize)))
-
-	log.Info("Buffer size calculated", "buffer_size", bufferSize, "page_size", pageSize, "max_events", maxEvents)
+	var (
+		maxEvents uint32     = 256
```
prb112, line 24: Does 256 align with the maxPods default per node?

controllers/enoexecevent/daemon/internal/tracepoint/tracepoint.go
```diff
@@ -168,9 +187,12 @@ func (tp *Tracepoint) processRecord(record *ringbuf.Record) (*types.ENOEXECInter
 	if len(record.RawSample) < 8 {
 		return nil, fmt.Errorf("record too short: %d bytes, expected at least 8 bytes", len(record.RawSample))
 	}
-	// TODO: Is this ok on s390x?
-	realParentTGID := int32(binary.LittleEndian.Uint32(record.RawSample[:4]))
-	currentTaskTGID := int32(binary.LittleEndian.Uint32(record.RawSample[4:]))
+	var order binary.ByteOrder = binary.LittleEndian
+	if binary.LittleEndian.Uint16([]byte{1, 0}) != 1 {
+		order = binary.BigEndian
+	}
```
prb112, line 193: Great.

controllers/enoexecevent/daemon/internal/tracepoint/tracepoint_program.go
```diff
@@ -18,43 +68,162 @@ func (tp *Tracepoint) initializeProgSpec() error {
...
+		asm.Mov.Reg(asm.R1, asm.R7).WithSymbol(CleanupLabel),
+		asm.Mov.Imm(asm.R2, 0),
+		asm.FnRingbufDiscard.Call(),
+	}
+}
+
+// exit exits the eBPF program with a return value of 0.
+func exit() asm.Instructions {
+	return asm.Instructions{
+		asm.Mov.Imm(asm.R0, 0).WithSymbol(ExitLabel), // Set return value to 0
```
prb112, line 226: Made met think of https://github.com/kubernetes/enhancements/issues/4960
There are different stop signals. Would this impact anything? I doubt it based on the asm.

docs/enhancements/MTO-0004-enoexec-monitoring.md
```diff
@@ -244,6 +244,8 @@ In large-scale deployments spanning many nodes, the architecture is intentionall
 
 ### Open Questions
 
+- Is it actually possible to retrieve the name of the process' command that fails to run?
```
prb112, line 247: Wouldn't the staxrox collector have some code related to this?

prb112 (approved):
/lgtm

controllers/enoexecevent/daemon/daemon.go
aleskandro: Hi @prb112, it does not. I'll be changing these values in a fine-tuning PR. I'm also considering making them configurable through the CR, but let me see what happens when I load a cluster with 'bad' containers for a bit and then follow up with a refinement PR.

controllers/enoexecevent/daemon/internal/tracepoint/tracepoint_program.go
aleskandro: I think we shuold be ok with the stop signals and how the ebpf program is attached, but I'll take a better look while finalizing the related epic.Thanks @prb112

docs/enhancements/MTO-0004-enoexec-monitoring.md
aleskandro: I'll take a look at this too, thanks!

aleskandro:
/override pull-ci-outrigger-project-multiarch-tuning-operator-main-ocp418-e2e-gcp

controllers/enoexecevent/daemon/daemon.go
prb112: When I had a crashed pod with cpuManager, it created 10K pods. See `OCPBUGS-48107: Deployment with OOMKilled Pod results in retriable failure loop creating thousands of ContainerStatusUnknown pods`. I think it's fixed in 4.21 (You may have to find the duplicate error)

aleskandro: we also want to avoid (D)DoS from this component being stressed by something,e.g., a deployment with many containers running a binary leading to enoexec and flooding api server and etcd to the aim of causing disruptions... so I think we'd expect that, in some extreme cases, events can be dropped and assume that such dropped errors among many come from a set of newly created resources that can be identified by the non dropped packets... or that eventually less errors are caught and the remaining ones are not dropped, allowing identification and remediation.

Does it make sense?

PR #662, opened 2025-07-21
NO-JIRA: Prevent webhook panic on partially configured plugins

AnnaZivkovic:
```
message: 'admission webhook "pod-placement-scheduling-gate.multiarch.openshift.io" denied the request: panic: runtime error: invalid memory address or nil pointer dereference [recovered]'
```

A nil pointer dereference could occur in the pod placement webhook when evaluating `cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled()`.

If the ClusterPodPlacementConfig had other plugins enabled but NodeAffinityScoring was nil, the check for cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() would panic.

This commit adds the necessary nil checks to the conditional statement, ensuring the entire struct path is validated before being accessed. This prevents the webhook from panicking and denying pod creation requests.

### DIFF
--- a/controllers/podplacement/scheduling_gate_mutating_webhook.go
+++ b/controllers/podplacement/scheduling_gate_mutating_webhook.go
@@ -80,7 +80,7 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
 	log := ctrllog.FromContext(ctx).WithValues("namespace", pod.Namespace, "name", pod.Name)
 
 	cppc := clusterpodplacementconfig.GetClusterPodPlacementConfig()
-	if cppc != nil && cppc.Spec.Plugins != nil && cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() {
+	if cppc != nil && cppc.Spec.Plugins != nil && cppc.Spec.Plugins.NodeAffinityScoring != nil && cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() {
 		pod.EnsureLabel(utils.PreferredNodeAffinityLabel, utils.LabelValueNotSet)
 	}
 	pod.EnsureLabel(utils.NodeAffinityLabel, utils.LabelValueNotSet)
--- a/pkg/e2e/operator/pod_placement_config_test.go
+++ b/pkg/e2e/operator/pod_placement_config_test.go
@@ -425,6 +425,46 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
 			By("The pod should not have any preferred affinities")
 			Eventually(framework.VerifyPodPreferredNodeAffinity(ctx, client, ns, "app", "test", nil), e2e.WaitShort).Should(Succeed())
 		})
+		It("Should create a pod when plugin NodeAffinityScoring is nil", func() {
+			var err error
+			By("Creating a v1beta1 ClusterPodPlacementConfig")
+			err = client.Create(ctx,
+				NewClusterPodPlacementConfig().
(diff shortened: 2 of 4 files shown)


controllers/podplacement/scheduling_gate_mutating_webhook.go
```diff
@@ -80,7 +80,7 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
 	log := ctrllog.FromContext(ctx).WithValues("namespace", pod.Namespace, "name", pod.Name)
 
 	cppc := clusterpodplacementconfig.GetClusterPodPlacementConfig()
-	if cppc != nil && cppc.Spec.Plugins != nil && cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() {
+	if cppc != nil && cppc.Spec.Plugins != nil && cppc.Spec.Plugins.NodeAffinityScoring != nil && cppc.Spec.Plugins.NodeAffinityScoring.IsEnabled() {
```
aleskandro, line 83: Can we add a test about this if this logic is being tested already?

AnnaZivkovic: This logic is being tested indirectly already. But we only have the cases where the plugin is nil or it is NodeAffinityScoring. Ill add a tests for this case

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -422,6 +423,69 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
 			By("The pod should not have any preferred affinities")
 			Eventually(framework.VerifyPodPreferredNodeAffinity(ctx, client, ns, "app", "test", nil), e2e.WaitShort).Should(Succeed())
 		})
+		It("Should create a deployment when plugin NodeAffinityScoring is nil", func() {
+			var err error
+			By("Creating a v1beta1 ClusterPodPlacementConfig")
+			err = client.Create(ctx,
+				NewClusterPodPlacementConfig().
+					WithName(common.SingletonResourceObjectName).
+					WithExecFormatErrorMonitor(true).
```
lwan-wanglin:
```suggestion
					WithPlugins().
```
We can use `WithPlugins()` to create a non-nil Plugins here

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -422,6 +423,69 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+			}), ppc)
+			Expect(err).NotTo(HaveOccurred(), "failed to get the v1beta1 ClusterPodPlacementConfig", err)
+			Eventually(framework.ValidateCreation(client, ctx)).Should(Succeed())
+			By("Create an ephemeral namespace")
+			ns := framework.NewEphemeralNamespace()
+			err = client.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer client.Delete(ctx, ns)
+			By("Create a deployment using the container")
```
aleskandro: Isn't creating a single pod sufficient from here? We are testing that the webhook doesn't fail when the cppc has non-nil plugins but nil nodeAffinityScoring: the webhook should run that code with a single pod

AnnaZivkovic: Yes a pod would be enough. This is a bit over the top

aleskandro:
lgtm, pending on https://github.com/outrigger-project/multiarch-tuning-operator/pull/662#discussion_r2227501193

PR #668, opened 2025-07-31
MULTIARCH-5594: Fix endianess detection

aleskandro:
This PR fixes the endianess detection algorithm. Tested on s390x and amd64.


aleskandro (commented):
/unhold

PR #670, opened 2025-08-01
MULTIARCH-5368: add validate webhook for local podplacementconfig priority

lwan-wanglin:
The pr implement the following:
- add validate webhook for local podplacementconfig `priority` filed
- add validate webhook for `NodeAffinityScoring` field

### DIFF
--- a/controllers/podplacementconfig/podplacementconfig_controller.go
+++ b/controllers/podplacementconfig/podplacementconfig_controller.go
@@ -1,0 +1,62 @@
+/*
+Copyright 2023 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package podplacementconfig
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	multiarchv1beta1 "github.com/openshift/multiarch-tuning-operator/apis/multiarch/v1beta1"
+)
+
+// PodPlacementConfigReconciler reconciles a PodPlacementConfig object
+type PodPlacementConfigReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=multiarch.openshift.io,resources=podplacementconfigs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=multiarch.openshift.io,resources=podplacementconfigs/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=multiarch.openshift.io,resources=podplacementconfigs/finalizers,verbs=update
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the PodPlacementConfig object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
+func (r *PodPlacementConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	// TODO(user): your logic here
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PodPlacementConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&multiarchv1beta1.PodPlacementConfig{}).
+		Complete(r)
+}
--- a/controllers/podplacementconfig/podplacementconfig_webhook.go
+++ b/controllers/podplacementconfig/podplacementconfig_webhook.go
@@ -1,0 +1,85 @@
+package podplacementconfig
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"sync"
+
+	admissionv1 "k8s.io/api/admission/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/openshift/multiarch-tuning-operator/apis/multiarch/common"
+	multiarchv1beta1 "github.com/openshift/multiarch-tuning-operator/apis/multiarch/v1beta1"
+	"github.com/openshift/multiarch-tuning-operator/controllers/podplacement/metrics"
+)
+
+// +kubebuilder:webhook:path=/validate-multiarch-openshift-io-v1beta1-podplacementconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=multiarch.openshift.io,resources=podplacementconfigs,verbs=create;update;delete,versions=v1beta1,name=validate-podplacementconfig.multiarch.openshift.io,admissionReviewVersions=v1
+
+type PodPlacementConfigWebhook struct {
+	apiReader client.Reader
+	decoder   admission.Decoder
+	once      sync.Once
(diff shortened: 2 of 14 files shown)


controllers/podplacementconfig/podplacementconfig_controller.go
```diff
@@ -0,0 +1,62 @@
...
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+// TODO(user): Modify the Reconcile function to compare the state specified by
+// the PodPlacementConfig object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.1/pkg/reconcile
+func (r *PodPlacementConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
Prashanth684, line 49: nothing to reconcile here - can we omit this file?

lwan-wanglin: I'm not sure where to add the [rbac](https://github.com/outrigger-project/multiarch-tuning-operator/pull/670/files#diff-b634b43220974b624355c40a0641a30f434e41eb8df6e1455e9c1098568efd51R36-R38) listing RBAC permissions for the new PodPlacementConfig API, so I added it in the Reconciler file. Could you advise where it should go?

Prashanth684: hmm..that's true and the reconciler actually needs to create the role/rolebinding for the rbac also i guess

aleskandro: Yeah, at this point, let's leave it as is... Punith proposed it in his PR too, so I guess we will feel the need for this file more and more while the implementation is executed.

aleskandro (commented):
/lgtm

lwan-wanglin:
/hold
one of new added e2e test case failed.

controllers/podplacementconfig/podplacementconfig_controller.go
AnnaZivkovic: If we know we will need a controller I think it is simplest to add it when creating the api then leaving a todo in the reconciler noting that we will implement it in a different story. This way we get the rbac permissions in

controllers/podplacementconfig/podplacementconfig_webhook.go
```diff
@@ -0,0 +1,107 @@
...
+				fmt.Errorf("failed to list existing PodPlacementConfigs in namespace %q: %w", req.Namespace, err))
+		}
+
+		var oldPPC multiarchv1beta1.PodPlacementConfig
+		var priorityChanged bool
+		if req.Operation == admissionv1.Update {
+			if err := w.decoder.DecodeRaw(req.OldObject, &oldPPC); err != nil {
+				return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to decode old object: %w", err))
+			}
+			priorityChanged = oldPPC.Spec.Priority != newPPC.Spec.Priority
```
aleskandro: can we admit the ppc here if the priority did not change and move the retrieval of the list of PPCs in lines 53-58 down to line 69 before the for loop?

lwan-wanglin: Hi @aleskandro, I have fixed it, PTAL! thanks

controllers/podplacementconfig/podplacementconfig_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+		if err := w.apiReader.List(ctx, &existingPPCs, client.InNamespace(req.Namespace)); err != nil {
+			return admission.Errored(http.StatusInternalServerError,
+				fmt.Errorf("failed to list existing PodPlacementConfigs in namespace %q: %w", req.Namespace, err))
+		}
+
+		for _, existing := range existingPPCs.Items {
+			if existing.Spec.Priority != newPPC.Spec.Priority {
+				continue
+			}
+			if req.Operation == admissionv1.Update && existing.Name == req.Name && !priorityChanged {
```
aleskandro:
```suggestion
			if req.Operation == admissionv1.Update && existing.Name == req.Name && !priorityChanged {
```

Do we still need this?

aleskandro: maybe we can remove !priorityChanged?

aleskandro: I'd also compact it with the above one:

```
if existing.Spec.Priority != newPPC.Spec.Priority || (req.Operation == admissionv1.Update && existing.Name == req.Name) {
```

controllers/podplacementconfig/podplacementconfig_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+
+	case admissionv1.Delete:
+		var oldPPC multiarchv1beta1.PodPlacementConfig
+		if err := w.decoder.DecodeRaw(req.OldObject, &oldPPC); err != nil {
+			return admission.Errored(http.StatusBadRequest, fmt.Errorf("failed to decode old object on delete: %w", err))
+		}
+
+		// TODO(user): your logic here
+
+		return admission.Allowed("delete allowed")
```
aleskandro: Can we remove this and just rely on the default  case?

controllers/podplacementconfig/podplacementconfig_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+	scheme    *runtime.Scheme
+}
+
+func (w *PodPlacementConfigWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
+	w.once.Do(func() {
+		w.decoder = admission.NewDecoder(w.scheme)
+	})
+
+	switch req.Operation {
+	case admissionv1.Create, admissionv1.Update:
```
aleskandro, line 34: I think we can distinguish the two cases here instead of branching inside for the specific behavior, by moving 42-52 and 54-64 to dedicated functions.

aleskandro: I would also suggest the two dedicated functions should be introduced in the model side (in the PodPlacementConfig type).

so that we can call them like

```
ok [bool], err [error] := newPPC.plugins.NodeAffinityScoring.validateArchitecturesSet() 
ok [bool], err [error] := newPPC.validatePriorityUpdate(old, list) bool, error
ok [bool], err [error] := newPPC.validatePriorityNew(list) bool, error
```

You could also wrap the verification of the plugins and plugin-specific pointers  in a method in the newPPC (cc @AnnaZivkovic for the other optimization)
```
newPPC.validatePluginsUpdate(old, new, list) // this will call newPPC.plugins.NodeAffinityScoring.validateArchitecturesSet() if needed
```

controllers/podplacementconfig/podplacementconfig_webhook.go
lwan-wanglin: yes, your are right, I will remove !priorityChanged

controllers/podplacementconfig/podplacementconfig_webhook.go
lwan-wanglin: ok

controllers/podplacementconfig/podplacementconfig_webhook.go
lwan-wanglin: firstly, I added the webhook file in hte model side, but I can't use client.Client in it, when I run `make generate`, I will hit this error 
```
/Users/lwan/go/src/github.com/openshift/multiarch-tuning-operator/apis/multiarch/v1beta1/clusterpodplacementconfig_webhook.go:39:2: invalid field type: interface{GroupVersionKindFor(obj k8s.io/apimachinery/pkg/runtime.Object) (k8s.io/apimachinery/pkg/runtime/schema.GroupVersionKind, error); IsObjectNamespaced(obj k8s.io/apimachinery/pkg/runtime.Object) (bool, error); RESTMapper() k8s.io/apimachinery/pkg/api/meta.RESTMapper; Scheme() *k8s.io/apimachinery/pkg/runtime.Scheme; sigs.k8s.io/controller-runtime/pkg/client.Reader; sigs.k8s.io/controller-runtime/pkg/client.Writer; sigs.k8s.io/controller-runtime/pkg/client.StatusClient; sigs.k8s.io/controller-runtime/pkg/client.SubResourceClientConstructor}
```

AnnaZivkovic: Those are great functions to have. 
The only one I can think of at the moment converting the label Selector. Which I dont think we need in this pr. 
`selector [labels.Selector], err [error] := newPPC.getLabelSelector() labels.Selector, error`

lwan-wanglin: fixed it

controllers/podplacementconfig/suite_test.go
```diff
@@ -164,3 +181,40 @@ func runManager() {
 		Succeed(), "manager is not ready yet")
 	suiteLog.Info("Manager is ready")
 }
+
+func getPodPlacementConfigValidatingWebHook() *v1.ValidatingWebhookConfiguration {
+	return &v1.ValidatingWebhookConfiguration{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "validating-webhook-configuration",
```
AnnaZivkovic, line 188: Can put this name in utils? 
`Name: utils.PodPlaecementValidatingWebhookConfiguration,`

PR #672, opened 2025-08-07
MULTIARCH-5605: Safe removal of the ENoExec plugin objects

AnnaZivkovic:
Introducing two finalizers to the `enoexec-event-handler-controller` Deployment to ensure a robust, ordered, and safe shutdown of the ENoExecEvent components.

Previously, when the ExecFormatErrorMonitor plugin was disabled or the operator was uninstalled, there was no guarantee on the deletion order of the enoexec components. This could lead to two potential issues:

1. The Deployment could be deleted before its associated DaemonSet.
2. The Deployment and DaemonSet could be deleted while ENoExecEvent custom resources still existed, preventing those events from ever being processed by the controller.

### DIFF
--- a/controllers/operator/clusterpodplacementconfig_controller.go
+++ b/controllers/operator/clusterpodplacementconfig_controller.go
@@ -139,6 +139,46 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
 		}
 		return ctrl.Result{Requeue: true}, nil
 	}
+	if clusterPodPlacementConfig.PluginsEnabled(common.ExecFormatErrorMonitorPluginName) {
+		if !controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.ExecFormatErrorFinalizerName) {
+			// Add the finalizer to the object
+			log.V(1).Info("Adding ENoExecEvent finalizer to the ClusterPodPlacementConfig")
+			controllerutil.AddFinalizer(clusterPodPlacementConfig, utils.ExecFormatErrorFinalizerName)
+			if err := r.Update(ctx, clusterPodPlacementConfig); err != nil {
+				log.Error(err, "Unable to update finalizers in the ClusterPodPlacementConfig")
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{Requeue: true}, nil
+		}
+
+		// Attempt to fetch the ENoExecEvent Deployment.
+		eNoExecEventDeployment := &appsv1.Deployment{}
+		err := r.Get(ctx, client.ObjectKey{
+			Name:      utils.EnoexecControllerName,
+			Namespace: utils.Namespace(),
+		}, eNoExecEventDeployment)
+		if err != nil {
+			// If the deployment is not found, it may be created later.
+			// If any other error occurs, log it and requeue.
+			log.Error(err, "Unable to fetch EnoexecController Deployment")
+			if client.IgnoreNotFound(err) != nil {
+				return ctrl.Result{}, err
+			}
+		} else {
+			if !controllerutil.ContainsFinalizer(eNoExecEventDeployment, utils.ExecFormatErrorFinalizerName) {
+				log.V(1).Info("Adding finalizers to the ENoExecEvent Deployment for the ENoExecEvents")
+				controllerutil.AddFinalizer(eNoExecEventDeployment, utils.ExecFormatErrorFinalizerName)
+				if err := r.Update(ctx, eNoExecEventDeployment); err != nil {
+					log.Error(err, "Unable to update finalizers on the EnoexecController Deployment")
+					// Requeue on conflict, as another process might have updated the object.
+					return ctrl.Result{}, err
+				}
+				// After a successful update, requeue to ensure the next state is processed.
+				return ctrl.Result{Requeue: true}, nil
+			}
+		}
+	}
+
 	err = r.dependentsStatusToClusterPodPlacementConfig(ctx, clusterPodPlacementConfig)
 	if err != nil {
 		log.Error(err, "Unable to retrieve the status of the PodPlacementConfig dependencies")
@@ -251,7 +291,7 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 	// The ClusterPodPlacementConfig is being deleted, cleanup the resources
 	log := ctrllog.FromContext(ctx).WithValues("operation", "handleDelete")
 
-	err := r.handleEnoexecDelete(ctx)
+	err := r.handleEnoexecDelete(ctx, clusterPodPlacementConfig)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -138,6 +138,49 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
...
+			// If the deployment is not found, it may be created later.
+			// If any other error occurs, log it and requeue.
+			log.Error(err, "Unable to fetch EnoexecController Deployment")
+			if client.IgnoreNotFound(err) != nil {
+				return ctrl.Result{}, err
+			}
+		} else {
+			needsUpdate := false
+			// Check if either finalizer is missing.
+			if !controllerutil.ContainsFinalizer(eNoExecEventDeployment, utils.ExecDeploymentFinalizerName) {
```
lwan-wanglin: Do we have a plan to add an eNoExec finalizer to ClusterPodPlacementConfig to prevent its deletion before the associated eNoExec object is deleted?

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -645,5 +669,105 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+			err = client.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer client.Delete(ctx, ns)
+			By("Creating a eNoExecEventwith")
+			enee := NewENoExecEvent().
+				WithName("test-enoexecevent").
+				WithCommand("foo-binary").
+				WithNodeName("test-name").
+				WithPodNamespace(ns.Name).
```
lwan-wanglin, line 711: should an ENoExecEvent CR outside the operator namespace allow to prevent the deletion of the deployment, depending on the current implementation logic [enoexecevent_controller.go](https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/controllers/enoexecevent/handler/enoexecevent_controller.go#L76-L79)

AnnaZivkovic: Interesting...I have it set that way so that the ENoExecEvent CR does not get deleted for the test. But that is problematic as we can end up with "orphan" ENoExecEvents that do not ever get deleted which would require the user to manually delete them to uninstall or disable ENoExecEventMonitor. Would an error/retry count be the way to go?

controllers/operator/clusterpodplacementconfig_controller.go
AnnaZivkovic: The way the [current logic ]( https://github.com/AnnaZivkovic/multiarch-manager-operator/blob/326dfc5b32b3f9cd2aeeb7c976122af3427a9e7d/controllers/operator/clusterpodplacementconfig_controller.go#L296-L298) is set up prevents the cppc or any of its associated objects getting deleted before the ENoExecEvents objects. So i dont think we need a finalizer there

pkg/e2e/operator/pod_placement_config_test.go
lwan-wanglin: In my opinion, we should not consider `ENoExecEvent `CRs in namespaces other than the operator namespace in deletion code. We should specify the namespace explicitly in `err = r.List(ctx, enoexecEventList, client.InNamespace(utils.Namespace()))  ` instead of just calling `r.List(ctx, enoexecEventList).`

For the testing part, I think you provided a good idea: we can add a different finalizer to block `ENoExecEvent `deletion.
What do you think?

controllers/operator/clusterpodplacementconfig_controller.go
lwan-wanglin: got it

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -409,32 +452,110 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 }
 
 func (r *ClusterPodPlacementConfigReconciler) handleEnoexecDelete(ctx context.Context) error {
-	log := ctrllog.FromContext(ctx)
-	log.Info("Disabling ENoExecEvent Controller")
-	objsToDelete := []utils.ToDeleteRef{
+	log := ctrllog.FromContext(ctx, "operation", "handleEnoexecDelete")
```
lwan-wanglin, line 453: Should we delete the `DaemonSet` first before checking the remaining `ENoExecEvent` CRs?
I mean, if we don’t delete the` DaemonSet` first, it will keep creating new `ENoExecEvent` CRs.

AnnaZivkovic: That is a good idea

pkg/e2e/operator/pod_placement_config_test.go
Prashanth684: > should an ENoExecEvent CR outside the operator namespace

is it possible to prevent this object from being created anywhere else at admission time?

Prashanth684: >  We should specify the namespace explicitly in err = r.List(ctx, enoexecEventList, client.InNamespace(utils.Namespace()))   instead of just calling r.List(ctx, enoexecEventList).

We can live with this for now - but ideally being an internal CR, the creation itself should be restricted except by us (the operator)

AnnaZivkovic: I added a finalizer to block ENoExecEvent deletion for the e2e test and I made it so that we only check the operators namespace. Any ENoExecEvents not in the right namespace were not created by the operator and I dont think we should block on them.

AnnaZivkovic: @Prashanth684  To ensure only our operator can create ENoExecEvent CRs, I think we'll need to implement a Validating Admission Webhook. At the moment it is hidden from the user in the UI

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: just for a final confirmation. Is the other finalizer currently added to the CPPC?

lwan-wanglin: we currently have [finalizers.multiarch.openshift.io/pod-placement]( https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/controllers/operator/clusterpodplacementconfig_controller.go#L135) in the CPPC

controllers/operator/clusterpodplacementconfig_controller.go
```diff
@@ -409,32 +452,111 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
...
+		return client.IgnoreNotFound(err)
+	}
+
+	log.Info("Removing the ExecDeployment finalizer from the ENoExec Deployment")
+	if controllerutil.RemoveFinalizer(deployment, utils.ExecDeploymentFinalizerName) {
+		if err := r.Update(ctx, deployment); err != nil {
+			log.Error(err, "Unable to remove finalizer from deployment.", "name", deployment.Name)
+			return err
+		}
+	}
```
aleskandro: I think we should not check the deployment here and we can remove lines 503-519.

I also think the deployment should just be deleted at line [481](https://github.com/outrigger-project/multiarch-tuning-operator/pull/672/files#diff-2ccc8a1f8d7e3da0a4eb974bab7660f7a801c392bc16d1114181815f1efe290aR481) instead of 577-580.

Since the finalizer will be in place, the deployment will have the deletionTimestamp set, but will continue to exist and run while the DaemonSet is fully deleted.

Then, we run the verification at line 521: when no other EnoExecEventList are found, we remove the finalizer in the deployment. At this point, the deployment will conclude its deletion lifecycle and we can remove the RBAC objects related to it?

pkg/e2e/operator/pod_placement_config_test.go
aleskandro: @Prashanth684 we could, but we are currently using other namespaces for some specific test cases where we need to avoid the controller processes the object.

I think it's harmless to leave this possibility. That would be a non-used object anyway and there are many other resources (even core resources) that can be created even if they lead to no actual change of state in the cluster.

@AnnaZivkovic , are we making sure that the deletion handler does not check for ENoExecEvents outside the operator's namespace?

controllers/operator/clusterpodplacementconfig_controller.go
aleskandro: By looking at the logic below.

If we do not add the finalizer to the CPPC, the user deletes the CPPC, and the operator's controller is restarted for any reason in this process, we may lose the deletion event (as the CPPC will not exist anymore at restart time due to the missing finalizer).

We should probably add the finalizer to the CPPC too.

controllers/operator/clusterpodplacementconfig_controller.go
AnnaZivkovic: @aleskandro If we dont remove the finalizer for the daemon set here, where would we do it? 
I like deleting the deployment along with the daemon set.

pkg/e2e/operator/pod_placement_config_test.go
AnnaZivkovic: @aleskandro  yes. This is the only check for ENoExecEvents
`err = r.List(ctx, enoexecEventList, client.InNamespace(utils.Namespace()))`

controllers/operator/clusterpodplacementconfig_controller.go
AnnaZivkovic: I did not think about that. Ill add another finalizer to the cppc

PR #673, opened 2025-08-11
Refactor: Add generic PluginEnabled helper method

AnnaZivkovic:
Introduced a generic PluginEnabled method on the Plugins struct to provide a centralized and extendable way to check if a specific plugin is enabled.


aleskandro (commented):
/lgtm

PR #675, opened 2025-08-14
MULTIARCH-5611: fix image inspector issue for image with shortname

lwan-wanglin:
This PR fixes the following:

- Adds a check for shortname image resolution. The `shortnames.Resolve` function needs to create a .cache file when it runs, so an empty directory is created for cache storage. Otherwise, it would attempt to create .cache in the root filesystem and fail with the below error:
```
"Resolve failed","controller":"pod","controllerGroup":"","controllerKind":"Pod","Pod":{"name":"test-deployment-7c89554898-kjnvr","namespace":"lwan-test"},"namespace":"lwan-test","name":"test-deployment-7c89554898-kjnvr","reconcileID":"82e77c59-1fcd-4720-aaf2-ea3876fbae57","image":"openshifttest/hello-openshift:1.2.0","error":"mkdir /.cache: read-only file system"
```


pkg/image/inspector.go
```diff
@@ -243,6 +258,69 @@ func marshaledImagePullSecrets(imageReference string, secrets [][]byte) ([]byte,
...
+	}
+
+	// 2. Attempt to pull each candidate in order
+	var pullErrs []error
+	for i, cand := range resolved.PullCandidates {
+		fqName := cand.Value.String()
+		log.V(1).Info("Attempting pull", "candidate", i, "fullName", fqName)
+
+		// Ensure the reference starts with //
+		if !strings.HasPrefix(fqName, "//") {
```
AnnaZivkovic: Do we need this check? Can [ParseReference](https://github.com/containers/image/blob/186df8702a61dd6ed16a88532ed9a10ddc679132/docker/archive/transport.go#L56C19-L56C25) handle the //?

hack/ci-test.sh
```diff
@@ -23,7 +23,7 @@ REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
 OPENSHIFT_CI=${OPENSHIFT_CI:-""}
 export ARTIFACT_DIR=${ARTIFACT_DIR:-$(mktemp -d)}
 GINKGO=${GINKGO:-"go run ${REPO_ROOT}/vendor/github.com/onsi/ginkgo/v2/ginkgo"}
-GINKGO_ARGS=${GINKGO_ARGS:-"-vv --randomize-all --randomize-suites -p -race -trace --keep-going --timeout=40m "}
+GINKGO_ARGS=${GINKGO_ARGS:-"-vv --randomize-all --randomize-suites -p -race -trace --keep-going --timeout=60m "}
```
AnnaZivkovic, line 26: I was wondering what the context for increasing this to 60 is? Are we seeing specific tests timing out? If so do we know why?

go.mod
```diff
@@ -33,6 +33,11 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.19.0
 )
 
+require (
+	github.com/chzyer/readline v1.5.1 // indirect
+	github.com/manifoldco/promptui v0.9.0 // indirect
+)
+
```
AnnaZivkovic: Can you move these into the indirect block?

hack/ci-test.sh
lwan-wanglin: As the number of test cases continues to grow, such as exec and local podplacementconfig, the execution time will inevitably increase. I checked the logs, and the current execution time has already reached 45–47 minutes. extend the timeout setting is not a harmful operation.

go.mod
lwan-wanglin: sure

pkg/image/inspector.go
lwan-wanglin: Yes, otherwise it will hit the error below when calling `docker.ParseReference`. In fact, we are already handling it this way, as seen here [pod_model.go#L230-L239](https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/controllers/podplacement/pod_model.go#L230-L239)
```
docker: image reference registry.access.redhat.com/openshifttest/hello-openshift:1.2.0 does not start with // * docker: image reference docker.io/openshifttest/hello-openshift:1.2.0 does not start with // Warning ArchAwareInspectionError 27s multiarch-tuning-operator Failed to retrieve the supported architectures after multiple retries: 2 errors occurred while pulling: * docker: image reference registry.access.redhat.com/openshifttest/hello-openshift:1.2.0 does not start with // * docker: image reference docker.io/openshifttest/hello-openshift:1.2.0 does not start with //
```

lwan-wanglin: it calls this [ParseReference](https://github.com/containers/image/blob/186df8702a61dd6ed16a88532ed9a10ddc679132/docker/docker_transport.go#L51) which first checks whether the image reference starts with `//`, then parses the normalized image name.

pkg/image/inspector.go
```diff
@@ -91,19 +93,32 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
 		AuthFilePath:                authFile.Name(),
+		RegistriesDirPath:           RegistryCertsDir(),
 		SystemRegistriesConfPath:    RegistriesConfPath(),
-		SystemRegistriesConfDirPath: RegistryCertsDir(),
+		SystemRegistriesConfDirPath: RegistriesConfDir(),
 		SignaturePolicyPath:         PolicyConfPath(),
 		DockerPerHostCertDirPath:    DockerCertsDir(),
 	}
+	// we expect the imageReference to start with `//`. Let's remove it for shortnames func
+	if strings.HasPrefix(imageReference, "//") {
```
aleskandro: If we expect the imageReference to always start with `//`, can we skip the if statement and just calculate the trimmedRef? 

At most, if we want to double check the imageReference starts with `//`, i'd do something like:

```
if !strings.HasPrefix(imageReference, "//") {
  return nil, errors.New("invalid value for the imageReference argument")
}
```

In the current version, if the imageReference does not start with `//`, we silently skip checking for shortnames.

pkg/image/inspector.go
```diff
@@ -243,6 +258,69 @@ func marshaledImagePullSecrets(imageReference string, secrets [][]byte) ([]byte,
...
+	}
+
+	// Optionally log a user-friendly description of the resolution process
+	if desc := resolved.Description(); desc != "" {
+		log.V(2).Info("Shortname resolution", "details", desc)
+	}
+
+	// 2. Attempt to pull each candidate in order
+	var pullErrs []error
+	for i, cand := range resolved.PullCandidates {
```
aleskandro, line 263: This method has a lot in common with what we do later in GetCompatibleArchitectureSet...


I wonder if we can rather change GetCompatibleArchitectureSet to:

1. Check for shortnames and compute the candidates as a list
2. If not a shortname, build the same list with only one item (the provided full imageReference)
3. Loop through the list like in this line, but running directly the logic to compuute the architectures.
4. The first one that 'succeeds' (the image is found and the architecture set is computed) leads to a in-loop `return` statement
5. If none succeed, we exit the loop and return error.

This way, we should save a few requests to the registry (that here are potentially multiplied by the number of possible registries for the shortnames and finally re-done for the resolved image to compute the compatible architectures)

lwan-wanglin: it's a good suggestion, let me fix it.

aleskandro (commented):
/lgtm

PR #676, opened 2025-08-16
MULTIARCH-5369:  Implement logic in the pod reconciler to work with the namespace-scoped PPC

AnnaZivkovic:
This PR updates the pod reconciler to support the new namespace-scoped PodPlacementConfig CRD, implementing the logic defined in [enhancement](https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/docs/enhancements/MTO-0002-local-pod-placement.md).

It introduces a strictly ordered reconciliation flow to ensure that scheduling constraints are applied with the correct precedence
Given a pod hitting the pod reconciliation loop:
- Verify the pod has the scheduling gate (return if not)
- Get all the pod placement configurations in the pod's namespace, sorted by descending priority
- For each pod placement configuration:
   - Check if the pod matches the label selector
   - Apply the configuration to the pod.
   - If a proposed configuration overlaps with the current pod specification, it is not applied and an event is recorded in the pod's event log. This ensures that the operator does not overwrite the user's configuration and that a PodPlacementConfig CRD with a higher priority gets precedence over the ones with lower priority.
- Apply the cluster-scoped ClusterPodPlacementConfig configuration to the pod. As the namespace-scoped configuration has already been applied, the cluster-scoped configuration will not overwrite any overlapping namespace-scoped configuration. We can also garantee that the cluster-scoped ClusterPodPlacementConfig's strong predicates will be applied to the pods as they are not handled by the namespace-scoped resource, which is focused on the scheduling scoring plugins.


internal/controller/podplacement/pod_model.go
```diff
@@ -166,12 +167,8 @@ func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequi
 }
 
 // SetPreferredArchNodeAffinity sets the node affinity for the pod to the preferences given in the ClusterPodPlacementConfig.
-func (pod *Pod) SetPreferredArchNodeAffinity(cppc *v1beta1.ClusterPodPlacementConfig) {
-	// Prevent overriding of user-provided kubernetes.io/arch preferred affinities or overwriting previously set preferred affinity
-	if pod.isPreferredAffinityConfiguredForArchitecture() {
```
lwan-wanglin, line 171: Hi @AnnaZivkovic,  Why are we removing this check? Should we ignore Pods if the user has already set a Preferred Affinity for architecture? 
If the intention is to cover both local PPC and CPPC, could we introduce a third option to distinguish whether the Preferred Affinity was user-defined or modified by our code? For example, we might rely on the label `multiarch.openshift.io/preferred-node-affinity`—treat Pods without this label(or `not-set`) or with the value `set` differently.

AnnaZivkovic: @lwan-wanglin that is a good idea! ill add a value to notate how the label got set. 

This check will need to be removed or modified because it can not find the difference between having the preferred node affinity setp with the CPPC or the PPC. With the original check it would not check the CPPC preferred affinity if the PPC set anything. Tho I think removing it is the wrong move now. Ill give it a more fine grained approach

AnnaZivkovic:
/test ci-index-multiarch-tuning-operator-bundle

Testing unrelated to pr. Looking for infrastructure errors

internal/controller/podplacement/pod_model.go
Prashanth684: imo whatever is configured in the PPC should always take precedence over what is configured in the CPPC in case of a conflict

PR #677, opened 2025-08-25
MULTIARCH-5584: Support image references with both tag and digest

AnnaZivkovic:
The image inspector now supports inspecting image references that specify both a tag and digest. When mirroring images that are frequently updated under the same tag (e.g., operator catalogs), there's a risk of pulling a different image version than intended if the tag is updated during the mirroring process. 
This code follows https://github.com/openshift/oc-mirror/pull/911 and will default to the digest if both digest and tag are present.

### DIFF
--- a/pkg/image/inspector.go
+++ b/pkg/image/inspector.go
@@ -93,6 +93,13 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 	// than do this everytime
 	sysregistriesv2.InvalidateCache()
 
+	// check if image reference has both tag and digest
+	imageReference, err = parseImageReference(imageReference)
+	if err != nil {
+		log.Error(err, "Couldn't parse image reference")
+		return nil, err
+	}
+
 	sys := &types.SystemContext{
 		AuthFilePath:                authFile.Name(),
 		RegistriesDirPath:           RegistryCertsDir(),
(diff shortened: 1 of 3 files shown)


pkg/image/inspector.go
```diff
@@ -196,6 +198,40 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 	return supportedArchitectures, nil
 }
 
+// parseImageReference standardizes an image name, removing the digest
+// if a tag is also present to ensure parser compatibility.
+func parseImageReference(imageName string) string {
+	repo, tag, sha := "", "", ""
```
aleskandro:
```suggestion
	var repo, tag, sha string
```

pkg/image/inspector.go
```diff
@@ -196,6 +198,40 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 	return supportedArchitectures, nil
 }
 
+// parseImageReference standardizes an image name, removing the digest
```
aleskandro: We should rather use the digest instead of the tag when both are present.

pkg/image/inspector.go
```diff
@@ -196,6 +198,40 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
+// if a tag is also present to ensure parser compatibility.
+func parseImageReference(imageName string) string {
```
aleskandro: Can we optimize and/or simplify this function further?

For example,
```
digestSplit := split(imageName, "@sha256:")
switch len(digestSplit) {
case 0:
  return error
case 1:
  // this is a tag-only image
  return imageName
}

// Since the length is 2, imageName is either a digest-only image or both the tag and the digest have been provided.

tagSplit := split(digestSplit[1], ":")
switch len(tagSplit) {
case 0:
  return error
case 1:
  // Since the length is 1, no tag has been found: imageName is a digest-only image.
  return imageName
}

// Since the length of tagSplit is also 2, we can now state that the imageName provides both the tag and the digest, and we should return the digest-only version. 

// digestSplit consists of two elements: "<repository>:<tag>" (0) and the digest (1).
// tagSplit consists of two elements: the repository name (0) and the tag (1).

return tagSplit[0] + "@sha256:" + diigestSplit[1]
```

pkg/image/inspector.go
aleskandro: also, let's use `digest` instead of `sha`.

AnnaZivkovic: I would, except I get the warning `Variable 'digest' collides with imported package name`

pkg/image/inspector.go
aleskandro:
```suggestion
// parseImageReference normalizes an imageName into a reference suitable for use
// with the inspection library. It returns one of the following:
//   1. A tag-only reference if no digest is present
//   2. A digest-only reference if a digest is present, dropping the tag when both
//      a tag and digest are specified in the pod's container image field
```

pkg/image/inspector.go
```diff
@@ -196,6 +202,29 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 	return supportedArchitectures, nil
 }
 
+// parseImageReference standardizes an image name, removing the digest
+// if a tag is also present to ensure parser compatibility.
+func parseImageReference(imageName string) (string, error) {
+	digestSplit := strings.Split(imageName, "@sha256:")
+	switch len(digestSplit) {
+	case 0:
+		return "", errors.New("invalid image name")
```
aleskandro: can we change the two errors so that we can be able to classify whether the error happens when spliiting by digest or tag?

pkg/image/inspector.go
```diff
@@ -196,6 +202,29 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
+	case 1:
+		return imageName, nil
```
aleskandro, line 216: For completeness, can we define `case 2:` (no action) and the `default:` one (returning an error)?

aleskandro (commented):
/lgtm
/approve

PR #678, opened 2025-09-08
MULTIARCH-5418: ENoExec Metrics

aleskandro:
This PR implements the metrics and the Prometheus Operator objects lifecycle for the ExecFormatError plugin.

In particular:
- A metric is for the enoexec-handler to record successful broadcasting of ExecFormatErrror events, mto_enoexecevents_total. This is user-facing and reports the number of exec format errors detected in the cluster.
- A metric tracking invalid ENoExecEvent objects
- An AlertRule to detect when either the enoexec-handler or the enoexec-daemon are not fully rolled-out, available and ready.
- An AlertRule that informs the user about the rate of increase of the mto_enoexecevents_total (1st derivative). The alert fires if `rate(mto_enoexec_eventstotal[6h]) > 0`, i.e., if any exec format error occurred in the past 6h.


controllers/operator/enoexecevent_objects.go
```diff
@@ -214,3 +219,87 @@ func buildDeploymentENoExecEventHandler(logVerbosity int) *appsv1.Deployment {
...
+			Namespace: utils.Namespace(),
+		},
+		Spec: monitoringv1.PrometheusRuleSpec{
+			Groups: []monitoringv1.RuleGroup{
+				{
+					Name: "multiarch-tuning-operator-enoexec-detected.rules",
+					Rules: []monitoringv1.Rule{
+						{
+							Alert: utils.ExecFormatErrorsDetected,
+							Expr:  intstr.FromString("rate(mto_enoexecevents[6h]) > 0"),
```
lwan-wanglin:
```suggestion
							Expr:  intstr.FromString("rate(mto_enoexecevents_total[6h]) > 0"),
```
The metric we registered in Go is `mto_enoexecevents_total` above, so the PrometheusRule expression should use this exact name here, right?

aleskandro: Thanks Lin

lwan-wanglin:
infrastructure issue
/retest

AnnaZivkovic:
infrastructure issue
/retest

PR #681, opened 2025-09-18
MULTIARCH-5655: Add DELETE permission to enoexec-event-daemon role for rollback functionality

lwan-wanglin:
This PR fixes the following:

- Add DELETE permission to the enoexec-event-daemon role
- Fix pod nodeSelector to correctly use the passed parameter in DescribeTable("it should create an ENoExecEvent CR when a pod triggers an ENOEXEC error on ...")
- Add an e2e test case to ensure no leftover ENoExecEvent CRs
- Add pod logs collection for enoexec when tests fail


lwan-wanglin:
infrastructure issue
/retest

aleskandro (commented):
/lgtm

PR #683, opened 2025-09-22
Optimize func VerifyPodsEvents to reduce loops

lwan-wanglin:
Previously, the function used nested loops to match pods with their corresponding events. This has been refactored to use a single-level loop with a map to track events, improving efficiency and reducing complexity.


AnnaZivkovic:
/lgtm cancel
Sorry did not see that it was a WIP

PR #692, opened 2025-11-05
MULTIARCH-5767: Fix image processing when port is present.

AnnaZivkovic:
When a extra `:`  is present in an image reference that is not related to a tag MTO does not handle the image correctly 
example: 
`image-registry.openshift-image-registry.svc:5000/ci-op-g7gzwqin/pipeline@sha256:d69e35f73b1e5c0bc2e12a95440a0ddb7976b5a69eb0b7c5a0772f0d954b1a57`
Is processed as 
`image-registry.openshift-image-registry.svc@sha256:d69e35f73b1e5c0bc2e12a95440a0ddb7976b5a69eb0b7c5a0772f0d954b1a57`


pkg/image/inspector.go
```diff
@@ -213,25 +213,48 @@ func parseImageReference(imageName string) (string, error) {
...
-		// Since the length is 1, no tag has been found: imageName is a digest-only image.
+	// Since the length is 2, imageName has a digest.
+	// We need to check if there's also a tag to remove.
+	// Format: [registry[:port]/][namespace/]image[:tag]@sha256:digest
+
+	beforeDigest := digestSplit[0]
+	digest := digestSplit[1]
+
+	// Find the last "/" to separate registry/namespace from image name
+	lastSlashIdx := strings.LastIndex(beforeDigest, "/")
```
lwan-wanglin, line 232: Hi @AnnaZivkovic, Could we refactor the code to follow the logic below? I think it would make it more readable — what do you think?
Also, I created a PR to test this logic: [multiarch-tuning-operator#693](https://github.com/outrigger-project/multiarch-tuning-operator/pull/693), I’ll close it after testing.
```
func parseImageReference(imageName string) (string, error) {
	parts := strings.Split(imageName, "@sha256:")
	switch len(parts) {
	case 0:
		return "", errors.New("invalid image name, must not be empty")
	case 1:
		// No digest present, return as-is
		return imageName, nil
	case 2:
		// Since the length is 2, imageName has a digest.
		// We need to check if there's also a tag to remove.
		// Format: [registry[:port]/][namespace/]image[:tag]@sha256:digest
		namePart := parts[0]
		digest := parts[1]
		/// Find last "/" to separate registry/namespace from image name
		lastSlash := strings.LastIndex(namePart, "/")
		lastColon := strings.LastIndex(namePart, ":")
		if lastColon > lastSlash {
			// Last ":" is a tag, remove it
			namePart = namePart[:lastColon]
		}
		return namePart + "@sha256:" + digest, nil
	default:
		return "", errors.New("invalid image name, must only have one digest")
	}
}
```

AnnaZivkovic:
Closing in favor of https://github.com/outrigger-project/multiarch-tuning-operator/pull/693

PR #696, opened 2025-11-05
MULTIARCH-5767: Fix image processing when port is present

AnnaZivkovic:
Fix image processing when port is present, regard to this [job failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_osdctl/811/pull-ci-openshift-osdctl-master-build/1985749621086883840)


pkg/image/inspector.go
```diff
@@ -208,30 +208,30 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
-	tagSplit := strings.Split(digestSplit[0], ":")
-	switch len(tagSplit) {
-	case 0:
-		return "", errors.New("invalid image name, image must not be empty")
-	case 1:
-		// Since the length is 1, no tag has been found: imageName is a digest-only image.
-		return imageName, nil
-	case 2:
-	default:
-		return "", errors.New("invalid image name, image contains more than one digest or tag")
```
AnnaZivkovic, line 232: We can skip validating the image tag format . That's not our component's responsibility, and the pod will fail to schedule or pull the image regardless. We can let the system handle this failure.

Prashanth684: while this is true..it wouldn't hurt to at least add a warning here

Prashanth684:
if we remove the port in cases like `image-registry.openshift-image-registry.svc:5000` won't it just go to default port ? and it might not work?

AnnaZivkovic:
> if we remove the port in cases like `image-registry.openshift-image-registry.svc:5000` won't it just go to default port ? and it might not work?

documentation here: https://docs.docker.com/reference/cli/docker/image/tag/#extended-description
From what I understand we can assume that there will be a / if there is a port so `image-registry.openshift-image-registry.svc:5000` would pass the logic. `image-registry.openshift-image-registry.svc:5000@sha256:abcdefg` would  be truncated to `image-registry.openshift-image-registry.svc@sha256:abcdefg` because :5000 would be interpreted as a tag. I think that would be expected behavior. 

@Prashanth684 @lwan-wanglin what do you think?

lwan-wanglin:
> From what I understand we can assume that there will be a / if there is a port so `image-registry.openshift-image-registry.svc:5000` would pass the logic. `image-registry.openshift-image-registry.svc:5000@sha256:abcdefg` would be truncated to `image-registry.openshift-image-registry.svc@sha256:abcdefg` because :5000 would be interpreted as a tag. I think that would be expected behavior.
I am ok with it

PR #706, opened 2025-11-27
MULTIARCH-5643: Remove Conmmand field from ENoExecEvent CR

pkenchap:
Remove Conmmand field from ENoExecEvent CR

### DIFF
--- a/bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
@@ -27,7 +27,7 @@ metadata:
     categories: OpenShift Optional, Other
     console.openshift.io/disable-operand-delete: "false"
     containerImage: registry.ci.openshift.org/origin/multiarch-tuning-operator:main
-    createdAt: "2025-11-07T19:53:09Z"
+    createdAt: "2025-11-27T04:58:59Z"
     features.operators.openshift.io/cnf: "false"
     features.operators.openshift.io/cni: "false"
     features.operators.openshift.io/csi: "false"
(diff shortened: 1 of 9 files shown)


prb112 (commented):
Did the command just not show properly?   It's super helpful when there are multiple processes in a container running

AnnaZivkovic:
> per helpful when there are multiple processes in a container running

The command field is not getting set as it does not seem possible to do. The ENoExecEvent does not need the field so we are removing it.

prb112:
~The ebpf doesn't come back with the pid?~
Right the Pid doesn't exist at the failed time.... makes sense.

prb112:
You'd have to store all the pids running on the system, and save to a map

bundle/manifests/multiarch-tuning-operator.clusterserviceversion.yaml
```diff
@@ -27,7 +27,7 @@ metadata:
     categories: OpenShift Optional, Other
     console.openshift.io/disable-operand-delete: "false"
     containerImage: registry.ci.openshift.org/origin/multiarch-tuning-operator:main
-    createdAt: "2025-11-07T19:53:09Z"
+    createdAt: "2025-11-27T04:58:59Z"
```
AnnaZivkovic, line 30: Since nothing else changed in this file can we revert the time stamp change?

pkenchap: When we run the make manifests this timestamp gets updated . I think this is required , if not I can remove it .

AnnaZivkovic: Then ignore me

AnnaZivkovic:
/override ocp418-e2e-azure ocp419-e2e-azure

These are infrastructure issues that lin is working on

PR #707, opened 2025-12-05
MULTIARCH-5543: Migrate Operator SDK to minimum supported version 1.33

pkenchap:
These are the migration todo checklist : 

- [x] PROJECT: go.kubebuilder.io/v4 layout
- [x] API: apis/ to api/  and api/v1beta1/ with +groupName
- [x] Controllers: controllers/ to internal/controller/
- [x] main.go: root (multi-binary with cmd/enoexec-daemon/)
- [x] Dockerfile: Updated COPY paths
- [x] Webhooks: v1 APIs + modern cert-manager


prb112:
*multiarch-tuning-operator-e2e-olm-kustomize* has issues in [hack/deploy-and-e2e.sh](https://github.com/openshift/multiarch-tuning-operator/blob/main/hack/deploy-and-e2e.sh).

```
+ operator-sdk run bundle registry.build01.ci.openshift.org/ci-op-fcibdpwk/pipeline@sha256:a0d2a9826a16160d25390dcb480ae17d7c7cb1c5cef498f7bfcc6f41a204b432 -n openshift-multiarch-tuning-operator --security-context-config restricted --timeout=10m
Error: no plugin could be resolved with key "go.kubebuilder.io/v4" for project version "3"
```

To resolve, I think you need to update https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/bundle.konflux.Dockerfile

cmd/main.go
```diff
@@ -60,14 +60,14 @@ import (
```
AnnaZivkovic, line 60: Can you put main.go back into cmd/main-binary and update the corresponding files? 
https://book.kubebuilder.io/migration/manually_migration_guide_gov3_to_gov4#now-lets-update-the-paths-in-the-project-file-accordingly

AnnaZivkovic: Sorry to ask you to move this again, but can we put it in cmd/main.go? That way when we try to run `operator-sdk create` it can find the main file

AnnaZivkovic:
Can you update all of the deprecated fields in config/ ? 
You should be able to find some examples here https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/getting-started/testdata/project

cmd/main.go
pkenchap: sure , done .

pkenchap:
@AnnaZivkovic , All comments are addressed , Please review .

AnnaZivkovic:
I noticed that when migrating the apis we changed from[ multi group to single group](https://book.kubebuilder.io/migration/multi-group.html) (`apis/multiarch/v1beta1 -> api/v1beta1`).  Is there any reason for this change away from  multi grouped (`api/multiarch/v1beta1`) ?  If there is a good reason to keep single group then we need to remove `multigroup: true` from the PROJECT file

pkenchap:
The project defines only _`multiarch.openshift.io`_ as its API group , if there a plan of adding additional groups (e.g., storage.openshift.io, networking.openshift.io), then the multiple API groups make sense . 
Current structure: `api/v1beta1/` is more straightforward and moreover Multi-group would add unnecessary nesting: `api/multiarch/v1beta1/ `and its easier navigation and maintenance. 

Single-group layout is the default and recommended approach for projects with one API group. 

Yes it's better to remove the `multigroup: true` from the PROJECT file.

pkenchap:
@AnnaZivkovic , I just made multigroup: `true` to `false` .

PR #712, opened 2025-12-22
MULTIARCH-5367: Deny ClusterPodPlacementConfig deletion when local ppcs exist

lwan-wanglin:
Add webhook validating for clusterpodplacementconfig, deny ClusterPodPlacementConfig delation if any local ppc exists.


internal/controller/operator/clusterpodplacementconfig_controller_test.go
```diff
@@ -688,9 +688,57 @@ var _ = Describe("internal/Controller/ClusterPodPlacementConfig/ClusterPodPlacem
...
+					WithNodeAffinityScoring(true).
+					WithNodeAffinityScoringTerm(utils.ArchitectureAmd64, 50).
+					Build(),
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create local PodPlacementConfig", err)
+			//nolint:errcheck
+			defer k8sClient.Delete(ctx, builder.NewPodPlacementConfig().WithName("test-ppc").WithNamespace(ns.Name).Build())
+			By("Attempting to delete ClusterPodPlacementConfig")
+			err = k8sClient.Delete(ctx, builder.NewClusterPodPlacementConfig().WithName(common.SingletonResourceObjectName).Build())
+			Expect(err).To(HaveOccurred(), "the ClusterPodPlacementConfig should not be able to deleted when local PodPlacementConfigs exsit", err)
```
AnnaZivkovic:
```suggestion
			Expect(err).To(HaveOccurred(), "the ClusterPodPlacementConfig should not be able to deleted when local PodPlacementConfigs exist", err)
```

pkg/e2e/podplacementconfig/pod_placement_config_test.go
```diff
@@ -76,6 +80,8 @@ var _ = Describe("The Multiarch Tuning Operator", func() {
 					Build(),
 			)
 			Expect(err).To(HaveOccurred(), "the PodPlacementConfig should not be accepted", err)
+			//nolint:errcheck
+			defer client.Delete(ctx, builder.NewPodPlacementConfig().WithName("test-ppc-2").WithNamespace(ns.Name).Build())
 		})
 		It("The webhook should deny creation when update a local ppc priority to an exsiting one", func() {
```
AnnaZivkovic:
```suggestion
		It("The webhook should deny creation when update a local ppc priority to an existing one", func() {
```

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -847,6 +847,49 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+					WithNodeAffinityScoring(true).
+					WithNodeAffinityScoringTerm(utils.ArchitectureAmd64, 50).
+					Build(),
+			)
+			Expect(err).NotTo(HaveOccurred(), "failed to create local PodPlacementConfig", err)
+			//nolint:errcheck
+			defer client.Delete(ctx, NewPodPlacementConfig().WithName("test-ppc").WithNamespace(ns.Name).Build())
+			By("Attempting to delete ClusterPodPlacementConfig")
+			err = client.Delete(ctx, NewClusterPodPlacementConfig().WithName(common.SingletonResourceObjectName).Build())
+			Expect(err).To(HaveOccurred(), "the ClusterPodPlacementConfig should not be able to deleted when local PodPlacementConfigs exsit", err)
```
AnnaZivkovic:
```suggestion
			Expect(err).To(HaveOccurred(), "the ClusterPodPlacementConfig should not be able to deleted when local PodPlacementConfigs exist", err)
```

pkg/e2e/podplacementconfig/pod_placement_config_test.go
```diff
@@ -76,6 +80,8 @@ var _ = Describe("The Multiarch Tuning Operator", func() {
 					Build(),
 			)
 			Expect(err).To(HaveOccurred(), "the PodPlacementConfig should not be accepted", err)
+			//nolint:errcheck
+			defer client.Delete(ctx, builder.NewPodPlacementConfig().WithName("test-ppc-2").WithNamespace(ns.Name).Build())
```
AnnaZivkovic, line 84: Do we need the delete when the resource should not have been created? what are the chances that there would be a left over ppc?

lwan-wanglin: This can occur if the function has bugs. The deletion here is not harmful; it is only to ensure that all PodPlacementConfigs are deleted, since we don’t want any remaining PPCs to block the deletion of the ClusterPodPlacementConfig, which could cause other test cases to fail.

PR #714, opened 2025-12-26
MULTIARCH-5382: Complete fallbackArchitecture validation and handling logic

lwan-wanglin:
The node affinity will be set to the configured `fallbackArchitecture` when image
inspection repeatedly fails and the `fallbackArchitecture` field is not empty.

This PR also includes the following changes:
- Add the `multiarch.openshift.io/fallback-arch` label to the Pod
- Introduce the `ArchAwareFallbackPredicateSet` event along with the corresponding
  event message

### DIFF
--- a/internal/controller/podplacement/pod_model.go
+++ b/internal/controller/podplacement/pod_model.go
@@ -123,6 +123,8 @@ func (pod *Pod) SetNodeAffinityArchRequirement(pullSecretDataList [][]byte) (boo
 	}
 
 	pod.setRequiredArchNodeAffinity(requirement)
+	pod.PublishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
+		ArchitecturePredicateSetupMsg+fmt.Sprintf("{%s}", strings.Join(requirement.Values, ", ")))
 	return true, nil
 }
 
@@ -162,8 +164,6 @@ func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequi
 	// if the nodeSelectorTerms were patched at least once, we set the nodeAffinity label to the set value, to keep
 	// track of the fact that the nodeAffinity was patched by the operator.
 	pod.EnsureLabel(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet)
-	pod.PublishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
-		ArchitecturePredicateSetupMsg+fmt.Sprintf("{%s}", strings.Join(requirement.Values, ", ")))
 }
 
 // SetPreferredArchNodeAffinity sets the node affinity for the pod to the preferences given in the ClusterPodPlacementConfig.
@@ -292,6 +292,32 @@ func (pod *Pod) trackAffinitySource(arch string, weight int32, source string, ap
 	} else {
 		pod.EnsureAnnotation(utils.PreferredNodeAffinitySourcesAnnotation, existingAnnotation+","+newEntry)
 	}
+}
+
+// setRequiredNodeAffinityToFallbackArchitecture sets the node affinity for the pod to the fallback architecture.
+func (pod *Pod) setRequiredNodeAffinityToFallbackArchitecture(architecture string) {
+	requirement := corev1.NodeSelectorRequirement{
+		Key:      utils.ArchLabel,
+		Operator: corev1.NodeSelectorOpIn,
+		Values:   []string{architecture},
+	}
+
+	if pod.Spec.Affinity == nil {
+		pod.Spec.Affinity = &corev1.Affinity{}
+	}
+
+	if pod.Spec.Affinity.NodeAffinity == nil {
+		pod.Spec.Affinity.NodeAffinity = &corev1.NodeAffinity{}
+	}
+
+	if pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
+	}
+
+	pod.setRequiredArchNodeAffinity(requirement)
+	pod.EnsureLabel(utils.FallbackArchitectureLabel, architecture)
+	pod.PublishEvent(corev1.EventTypeWarning, ArchitectureAwareFallbackNodeAffinitySet,
+		ArchitectureFallbackSetupMsg+fmt.Sprintf("{%s}", architecture))
 }
 
 func (pod *Pod) getArchitecturePredicate(pullSecretDataList [][]byte) (corev1.NodeSelectorRequirement, error) {
--- a/internal/controller/podplacement/pod_reconciler.go
+++ b/internal/controller/podplacement/pod_reconciler.go
@@ -150,6 +150,11 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 		// Publish this event and remove the scheduling gate.
 		log.Info("Max retries Reached. The pod will not have the nodeAffinity set.")
 		pod.PublishEvent(corev1.EventTypeWarning, ImageArchitectureInspectionError, fmt.Sprintf("%s: %s", ImageInspectionErrorMaxRetriesMsg, err.Error()))
+
+		if cppc != nil && cppc.Spec.FallbackArchitecture != "" {
+			log.Info("Setting the nodeAffinity to the fallback architecture", "fallbackArchitecture", cppc.Spec.FallbackArchitecture)
+			pod.setRequiredNodeAffinityToFallbackArchitecture(cppc.Spec.FallbackArchitecture)
+		}
 	}
 	// If the pod has been processed successfully or the max retries have been reached, remove the scheduling gate.
 	if err == nil || pod.maxRetries() {
--- a/internal/controller/podplacement/pod_reconciler_test.go
+++ b/internal/controller/podplacement/pod_reconciler_test.go
@@ -5,8 +5,6 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/openshift/multiarch-tuning-operator/api/v1beta1"
-	"github.com/openshift/multiarch-tuning-operator/pkg/e2e"
 	"k8s.io/apimachinery/pkg/util/sets"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -18,11 +16,14 @@ import (
 
 	imgspecv1 "github.com/opencontainers/image-spec/specs-go/v1"
 
-	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
-
+	"github.com/openshift/multiarch-tuning-operator/api/common"
+	"github.com/openshift/multiarch-tuning-operator/api/v1beta1"
+	"github.com/openshift/multiarch-tuning-operator/pkg/e2e"
+	"github.com/openshift/multiarch-tuning-operator/pkg/informers/clusterpodplacementconfig"
 	. "github.com/openshift/multiarch-tuning-operator/pkg/testing/builder"
 	. "github.com/openshift/multiarch-tuning-operator/pkg/testing/framework"
 	"github.com/openshift/multiarch-tuning-operator/pkg/testing/image/fake/registry"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
 )
 
 var _ = Describe("Internal/Controller/Podplacement/PodReconciler", func() {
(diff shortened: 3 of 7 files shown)


AnnaZivkovic:
We are missing Conversion Webhook Updates for FallbackArchitecture. The new FallbackArchitecture field added to the v1beta1 API is not being handled in the conversion webhooks between v1alpha1 and v1beta1. This will cause data loss during API version conversions.

Files that need updates:api/v1alpha1/clusterPodPlacementConfig_conversion.go

Can you also add/update the conversion tests to reflect these changes

AnnaZivkovic:
Why not put it in the plugins section?  I could see it being clearer in spec but for consistency it would be a plugin.

internal/controller/podplacement/pod_reconciler.go
```diff
@@ -137,6 +137,11 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 		// Publish this event and remove the scheduling gate.
 		log.Info("Max retries Reached. The pod will not have the nodeAffinity set.")
 		pod.PublishEvent(corev1.EventTypeWarning, ImageArchitectureInspectionError, fmt.Sprintf("%s: %s", ImageInspectionErrorMaxRetriesMsg, err.Error()))
+
+		if cppc != nil && cppc.Spec.FallbackArchitecture != "" {
+			log.Info("Setting the nodeAffinity to the fallback architecture", "fallbackArchitecture", cppc.Spec.FallbackArchitecture)
```
AnnaZivkovic, line 155: Should we also publish an event in the pod? or set a different label to let the users know?

lwan-wanglin: I've already added a new label `multiarch.openshift.io/fallback-arch` and an event `ArchAwareFallbackPredicateSet` for the new fallback architecture feature 
https://github.com/outrigger-project/multiarch-tuning-operator/pull/714/changes#diff-983f09c61cb7d791e8762ba6a4a5fcac3eeb5224b0b682efa532414adf1166ef

lwan-wanglin:
> Why not put it in the plugins section? I could see it being clearer in spec but for consistency it would be a plugin.

Thanks for your suggestion. In my view, fallbackArchitecture is not a plugin—it’s just an auxiliary field for the image inspector to provide a fallback. Making it a plugin would unnecessarily complicate the API and the user experience. Keeping it in spec keeps the design simple, clear, and easy to use. @AnnaZivkovic, what do you think? 
I’d also like to hear @Prashanth684’s thoughts on this.

internal/controller/podplacement/pod_model.go
```diff
@@ -207,6 +207,33 @@ func (pod *Pod) SetPreferredArchNodeAffinity(cppc *v1beta1.ClusterPodPlacementCo
 	pod.PublishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet, ArchitecturePreferredPredicateSetupMsg)
 }
 
+// SetNodeAffinityToFallbackArchitecture sets the node affinity for the pod to the fallback architecture.
+func (pod *Pod) SetNodeAffinityToFallbackArchitecture(architecture string) {
```
AnnaZivkovic: Can we change the name to setRequiredNodeAffinityToFallbackArchitecture? or something that has Required in it so we know what kind of affinity we are setting.

internal/controller/podplacement/pod_reconciler_test.go
```diff
@@ -111,6 +114,80 @@ var _ = Describe("Internal/Controller/Podplacement/PodReconciler", func() {
 				// Polling set to 250ms such that the error count is shown in the logs at each update.
 			})
 		})
+
+		Context("with an image that cannot be inspected and fallback architecture is set", Serial, func() {
+			It("sets the node affinity to the fallback architecture", func() {
+				// Update the ClusterPodPlacementConfig to set the fallback architecture
+				ppc := &v1beta1.ClusterPodPlacementConfig{}
```
AnnaZivkovic:
```suggestion
				cppc := &v1beta1.ClusterPodPlacementConfig{}
```

internal/controller/podplacement/pod_reconciler_test.go
```diff
@@ -111,6 +114,80 @@ var _ = Describe("Internal/Controller/Podplacement/PodReconciler", func() {
...
+												Key:      utils.ArchLabel,
+												Operator: corev1.NodeSelectorOpIn,
+												Values:   []string{utils.ArchitectureAmd64},
+											},
+										},
+									},
+								},
+							},
+						}), "unexpected node affinity")
+					g.Expect(pod.Labels).To(HaveKeyWithValue(utils.FallbackArchitectureLabel, utils.ArchitectureAmd64),
```
AnnaZivkovic, line 171: Can you also check that the max retries?

AnnaZivkovic:
Can you add a test for when fall back is set but it is ""?

AnnaZivkovic:
When a pod is created with a scheduling gate, the reconciler initiates a sequence that begins by applying the preferred architecture node affinity (assuming the plugin is active) before attempting image inspection. If the inspection fails, the system enters a retry loop; upon reaching the maximum retry limit (5), it triggers the fallback mechanism. This final step enforces a required affinity for the fallback architecture, applies the corresponding label, and removes the scheduling gate to allow placement. 

However, I can see a potential issue arises because the preferred affinity is applied so early in the cycle: the resulting pod ends up carrying conflicting instructions—both the originally calculated preferred affinity and the newly enforced fallback affinity. I think this could lead to unpredictable scheduling behavior if the weight on the preferred affinity is high enough and available in the cluster. Would this be desired behavior?

lwan-wanglin:
> When a pod is created with a scheduling gate, the reconciler initiates a sequence that begins by applying the preferred architecture node affinity (assuming the plugin is active) before attempting image inspection. If the inspection fails, the system enters a retry loop; upon reaching the maximum retry limit (5), it triggers the fallback mechanism. This final step enforces a required affinity for the fallback architecture, applies the corresponding label, and removes the scheduling gate to allow placement.
> 
> However, I can see a potential issue arises because the preferred affinity is applied so early in the cycle: the resulting pod ends up carrying conflicting instructions—both the originally calculated preferred affinity and the newly enforced fallback affinity. I think this could lead to unpredictable scheduling behavior if the weight on the preferred affinity is high enough and available in the cluster. Would this be desired behavior?

Good point. In Kubernetes scheduling, required node affinity is always enforced first during the filtering phase, and preferred affinity is only used later for scoring among the already-eligible nodes. Once the fallback applies a required affinity, any existing preferred affinity will not affect nodes outside that constraint, so scheduling remains deterministic.
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity

Prashanth684:
> > Why not put it in the plugins section? I could see it being clearer in spec but for consistency it would be a plugin.
> 
> Thanks for your suggestion. In my view, fallbackArchitecture is not a plugin—it’s just an auxiliary field for the image inspector to provide a fallback. Making it a plugin would unnecessarily complicate the API and the user experience. Keeping it in spec keeps the design simple, clear, and easy to use. @AnnaZivkovic, what do you think? I’d also like to hear @Prashanth684’s thoughts on this.

Right - agree on this - it should not be a plugin, it should be a base functionality which the users should be able to turn on or off as a whole. Plugins are more like add on features on top of the vanilla functionality of the operator.

internal/controller/podplacement/pod_model.go
```diff
@@ -207,6 +207,33 @@ func (pod *Pod) SetPreferredArchNodeAffinity(cppc *v1beta1.ClusterPodPlacementCo
...
+	}
+
+	if pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution == nil {
+		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = &corev1.NodeSelector{}
+	}
+
+	pod.setRequiredArchNodeAffinity(requirement)
+	pod.EnsureLabel(utils.FallbackArchitectureLabel, architecture)
+	pod.PublishEvent(corev1.EventTypeWarning, ArchitectureAwareFallbackNodeAffinitySet,
+		ArchitectureFallbackSetupMsg+fmt.Sprintf(": %s", architecture))
```
AnnaZivkovic: Can we keep consistent notation? 

 
```suggestion
		ArchitectureFallbackSetupMsg+fmt.Sprintf("{%s}"", architecture))
		
```

AnnaZivkovic:
The `setRequiredNodeAffinityToFallbackArchitecture` function calls `pod.setRequiredArchNodeAffinity(requirement)` at line 231, which publishes its own event (line 164-165). Then at line 233, it publishes another event. This     
  results in two events being created for a single operation:                                                                                                                                                                     
                                                                                                                                                                                                                                  
  - Event 1 (Normal): "Architecture predicate setup..." from setRequiredArchNodeAffinity                                                                                                                                          
  - Event 2 (Warning): "Image inspection failed; setting the nodeAffinity to the fallback architecture: amd64"  
  
  Do you think this is confusing for the user? Should we only see the warning?

internal/controller/podplacement/pod_model.go
```diff
@@ -207,6 +207,33 @@ func (pod *Pod) SetPreferredArchNodeAffinity(cppc *v1beta1.ClusterPodPlacementCo
...
 }
 
+// setRequiredNodeAffinityToFallbackArchitecture sets the node affinity for the pod to the fallback architecture.
+func (pod *Pod) setRequiredNodeAffinityToFallbackArchitecture(architecture string) {
+	requirement := corev1.NodeSelectorRequirement{
+		Key:      utils.ArchLabel,
+		Operator: corev1.NodeSelectorOpIn,
+		Values:   []string{architecture},
+	}
+	pod.ensureArchitectureLabels(requirement)
```
AnnaZivkovic: Is called with a requirement containing only the fallback architecture. If fallback is amd64, the pod gets labeled with `SingleArchLabel,` but the actual container images might support multiple architectures and we just failed to inspect them. This could mislead users/monitoring systems into thinking the pod is single-arch when it might not be. Could  we add a specific label like` utils.FallbackUsedLabel ` and not calling `ensureArchitectureLabels` at all, modifying the label setting to indicate uncertainty, or is it OK as is?

lwan-wanglin: I’ll just remove it, as I’ve already added utils.FallbackArchitectureLabel below.

internal/controller/podplacement/pod_model.go
```diff
@@ -162,8 +162,15 @@ func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequi
...
-		ArchitecturePredicateSetupMsg+fmt.Sprintf("{%s}", strings.Join(requirement.Values, ", ")))
+	// Publish event unless explicitly skipped via the optional publishEvent parameter
+	doPublish := true
+	if len(publishEvent) > 0 {
+		doPublish = publishEvent[0]
+	}
+	if doPublish {
+		pod.PublishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,
+			ArchitecturePredicateSetupMsg+fmt.Sprintf("{%s}", strings.Join(requirement.Values, ", ")))
+	}
```
AnnaZivkovic: Insted of a flag can we split the functions?  I think the flag has unclear intent  and using optional parameters to suppress side effects usually indicates the function is doing too much.                                                                                                   
                                                                                                                                                                                                                          
 ```                                                                                                                                                                                                                         
  // Remove publishEvent parameter                                                                                                                                                                                        
  func (pod *Pod) setRequiredArchNodeAffinity(requirement corev1.NodeSelectorRequirement) {                                                                                                                               
      // ... existing logic ...                                                                                                                                                                                           
      pod.EnsureLabel(utils.NodeAffinityLabel, utils.NodeAffinityLabelValueSet)                                                                                                                                           
      // Don't publish event here                                                                                                                                                                                         
  }                                                                                                                                                                                                                       
                                                                                                                                                                                                                          
  // In setRequiredNodeAffinityToFallbackArchitecture:                                                                                                                                                                    
  func (pod *Pod) setRequiredNodeAffinityToFallbackArchitecture(architecture string) {                                                                                                                                    
      requirement := corev1.NodeSelectorRequirement{...}                                                                                                                                                                  
      // Initialize affinity if needed (keep this part)                                                                                                                                                                   
      pod.setRequiredArchNodeAffinity(requirement)                                                                                                                                                                        
      pod.EnsureLabel(utils.FallbackArchitectureLabel, architecture)                                                                                                                                                      
      // Publish only the fallback-specific event                                                                                                                                                                         
      pod.PublishEvent(corev1.EventTypeWarning, ArchitectureAwareFallbackNodeAffinitySet,                                                                                                                                 
          ArchitectureFallbackSetupMsg+fmt.Sprintf("{%s}", architecture))                                                                                                                                                 
  }                                                                                                                                                                                                                       
                                                                                                                                                                                                                          
  // In SetNodeAffinityArchRequirement (original caller):                                                                                                                                                                 
  func (pod *Pod) SetNodeAffinityArchRequirement(...) {                                                                                                                                                                   
      // ... existing logic ...                                                                                                                                                                                           
      pod.setRequiredArchNodeAffinity(requirement)                                                                                                                                                                        
      // Publish the normal event here explicitly                                                                                                                                                                         
      pod.PublishEvent(corev1.EventTypeNormal, ArchitectureAwareNodeAffinitySet,                                                                                                                                          
          ArchitecturePredicateSetupMsg+fmt.Sprintf("{%s}", strings.Join(requirement.Values, ", ")))                                                                                                                      
  }     

```

lwan-wanglin: Great suggestion—implemented the fix.

PR #715, opened 2025-12-30
MULTIARCH-5370: Add/remove ClusterPodPlacementConfig finalizer on PodPlacementConfig changes

lwan-wanglin:
Add the finalizer finalizers.multiarch.openshift.io/no-pod-placement-config when the first PodPlacementConfig is created and remove it when the last pod placement config is removed.

### DIFF
--- a/internal/controller/operator/clusterpodplacementconfig_controller.go
+++ b/internal/controller/operator/clusterpodplacementconfig_controller.go
@@ -27,14 +27,20 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	errorutils "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	ctrllog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/openshift/library-go/pkg/operator/events"
 
@@ -66,6 +72,7 @@ const (
 	waitingForUngatingPodsError         = "waiting for pods with the scheduling gate to be ungated"
 	waitingForWebhookSInterruptionError = "re-queueing to ensure the webhook objects deletion interrupt pods gating before checking the pods gating status"
 	clusterPodPlacementConfigNotReady   = "cluster pod placement config is not ready yet. re-queueing"
+	RemainingPodPlacementConfig         = "cannot delete: namespaced PodPlacementConfig resources still exist"
 )
 
 //+kubebuilder:rbac:groups=multiarch.openshift.io,resources=clusterpodplacementconfigs,verbs=get;list;watch;create;update;patch;delete
@@ -140,6 +147,30 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
 		}
 		return ctrl.Result{Requeue: true}, nil
 	}
+
+	// Handle the no-pod-placement-config finalizer
+	ppcList := &multiarchv1beta1.PodPlacementConfigList{}
+	if err := r.List(ctx, ppcList); err != nil {
+		log.Error(err, "Unable to list PodPlacementConfigs")
+		return ctrl.Result{}, err
+	}
+	shouldHavePPCFinalizer := len(ppcList.Items) > 0
+	hasPPCFinalizer := controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
+	if shouldHavePPCFinalizer != hasPPCFinalizer {
+		if shouldHavePPCFinalizer {
+			log.V(1).Info("Adding no-pod-placement-config finalizer to the ClusterPodPlacementConfig")
+			controllerutil.AddFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
+		} else {
+			log.V(1).Info("Removing no-pod-placement-config finalizer from the ClusterPodPlacementConfig")
+			controllerutil.RemoveFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
+		}
+		if err := r.Update(ctx, clusterPodPlacementConfig); err != nil {
+			log.Error(err, "Unable to update finalizers on the ClusterPodPlacementConfig")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
+
 	if clusterPodPlacementConfig.PluginsEnabled(common.ExecFormatErrorMonitorPluginName) {
 		if !controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.ExecFormatErrorFinalizerName) {
 			// Add the finalizer to the object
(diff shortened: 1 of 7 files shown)


internal/controller/operator/clusterpodplacementconfig_controller.go
```diff
@@ -140,6 +143,35 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
...
 		return ctrl.Result{Requeue: true}, nil
 	}
+
+	// Handle the no-pod-placement-config finalizer
+	ppcList := &multiarchv1beta1.PodPlacementConfigList{}
+	if err := r.List(ctx, ppcList); err != nil {
+		log.Error(err, "Unable to list PodPlacementConfigs")
+		return ctrl.Result{}, err
+	}
+	if len(ppcList.Items) > 0 {
```
AnnaZivkovic: Can you reduce the number of nested if statements?

internal/controller/operator/clusterpodplacementconfig_controller.go
```diff
@@ -140,6 +143,35 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
...
+		if !controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer) {
+			log.V(1).Info("Adding no-pod-placement-config finalizer to the ClusterPodPlacementConfig")
+			controllerutil.AddFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
+			if err := r.Update(ctx, clusterPodPlacementConfig); err != nil {
+				log.Error(err, "Unable to update finalizers in the ClusterPodPlacementConfig")
+				return ctrl.Result{}, err
+			}
+			return ctrl.Result{Requeue: true}, nil
+		}
+	} else {
```
AnnaZivkovic: Maybe we could do an else if?

internal/controller/operator/clusterpodplacementconfig_controller.go
lwan-wanglin: thanks for your suggestion, reduced if statements

internal/controller/operator/clusterpodplacementconfig_controller.go
```diff
@@ -878,6 +906,19 @@ func isDeploymentUpToDate(deployment *appsv1.Deployment) bool {
...
+			&multiarchv1beta1.PodPlacementConfig{},
+			handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, obj client.Object) []reconcile.Request {
+				return []reconcile.Request{
+					{
+						NamespacedName: types.NamespacedName{
+							Name: common.SingletonResourceObjectName,
+						},
+					},
+				}
+			}),
```
AnnaZivkovic, line 940: Could we try to only reconcile on create/delete? I dont think we need to reconcile on an update of the ppc.
Maybe we could use a predicate to filter with? https://book.kubebuilder.io/reference/watching-resources/predicates-with-watch

AnnaZivkovic:
Can you add handling when trying to delete the CPPC? 
```
func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
      clusterPodPlacementConfig *multiarchv1beta1.ClusterPodPlacementConfig) error {
....
    if controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer) {
          log.V(1).Info("Checking for existing PodPlacementConfig resources before deletion")

          if err := r.List(ctx, ppcList); err != nil {
              log.Error(err, "Unable to list PodPlacementConfigs during deletion")
              return err
          }
          ...
          if len(ppcList.Items) > 0 {
           // error message as to why we can not delete cppc

```

internal/controller/operator/clusterpodplacementconfig_controller.go
```diff
@@ -140,6 +143,31 @@ func (r *ClusterPodPlacementConfigReconciler) Reconcile(ctx context.Context, req
...
+		return ctrl.Result{Requeue: true}, nil
+	} else if len(ppcList.Items) == 0 && controllerutil.ContainsFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer) {
+		log.V(1).Info("Removing no-pod-placement-config finalizer from the ClusterPodPlacementConfig")
+		controllerutil.RemoveFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
+		if err := r.Update(ctx, clusterPodPlacementConfig); err != nil {
+			log.Error(err, "Unable to remove finalizers from the ClusterPodPlacementConfig")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{Requeue: true}, nil
+	}
```
AnnaZivkovic:
```suggestion
  if shouldHaveFinalizer != hasFinalizer {
      if shouldHaveFinalizer {
          log.V(1).Info("Adding no-pod-placement-config finalizer to the ClusterPodPlacementConfig")
          controllerutil.AddFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
      } else {
          log.V(1).Info("Removing no-pod-placement-config finalizer from the ClusterPodPlacementConfig")
          controllerutil.RemoveFinalizer(clusterPodPlacementConfig, utils.CPPCNoPPCObjectFinalizer)
      }

      // Common error handling - only written once
      if err := r.Update(ctx, clusterPodPlacementConfig); err != nil {
          log.Error(err, "Unable to update finalizers in the ClusterPodPlacementConfig")
          return ctrl.Result{}, err
      }
      return ctrl.Result{Requeue: true}, nil
  }

```

What would you think about this flow?

internal/controller/operator/clusterpodplacementconfig_controller_test.go
```diff
@@ -733,12 +733,91 @@ var _ = Describe("internal/Controller/ClusterPodPlacementConfig/ClusterPodPlacem
...
+			Eventually(framework.ValidateDeletion(k8sClient, ctx)).Should(Succeed(), "the ClusterPodPlacementConfig should be deleted")
+		})
+		It("should add the finalizer when a namespaced PodPlacementConfig exists and remove it when none exist", func() {
+			By("Create an ephemeral namespace")
+			ns := framework.NewEphemeralNamespace()
+			err := k8sClient.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer k8sClient.Delete(ctx, ns)
+			By("Creating a local PodPlacementConfig with a Prioriry setting")
```
AnnaZivkovic:
```suggestion
			By("Creating a local PodPlacementConfig with a Priority setting")
```

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -885,11 +885,93 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+			Eventually(framework.ValidateCreation(client, ctx)).Should(Succeed())
+		})
+		It("should add the finalizer when a namespaced PodPlacementConfig exists and remove it when none exist", func() {
+			By("Create an ephemeral namespace")
+			ns := framework.NewEphemeralNamespace()
+			err := client.Create(ctx, ns)
+			Expect(err).NotTo(HaveOccurred())
+			//nolint:errcheck
+			defer client.Delete(ctx, ns)
+			By("Creating a local PodPlacementConfig with a Prioriry setting")
```
AnnaZivkovic:
```suggestion
			By("Creating a local PodPlacementConfig with a Priority setting")
```

internal/controller/operator/clusterpodplacementconfig_controller_test.go
```diff
@@ -733,12 +733,91 @@ var _ = Describe("internal/Controller/ClusterPodPlacementConfig/ClusterPodPlacem
...
+			}).Should(Succeed(), "the PodPlacementConfig should be deleted")
+			By("Verifying the cppc removing the no-pod-placement-config finalizer")
+			Eventually(func(g Gomega) {
+				cppc := &v1beta1.ClusterPodPlacementConfig{}
+				err = k8sClient.Get(ctx, crclient.ObjectKey{
+					Name: common.SingletonResourceObjectName,
+				}, cppc)
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(cppc.Finalizers).NotTo(ContainElement(utils.CPPCNoPPCObjectFinalizer))
+			}).Should(Succeed(), "the ClusterPodPlacementConfig should remove the no-pod-placement-config finalizer")
```
AnnaZivkovic, line 818: Can we add a test trying to  delete the CPPC when PPCs exists?

lwan-wanglin: We already have a [test case](https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/internal/controller/operator/clusterpodplacementconfig_controller_test.go#L703)  for this scenario, but in practice the validating webhook will deny the CPPC deletion before the request reaches the Reconcile logic.

PR #724, opened 2026-01-23
MULTIARCH-5800: Images with attestation manifests lead the operator to set a node affinity for the architecture "unknown"

AnnaZivkovic:
```
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:0298f7c3ceec45da42b10f52b8edd4c5ebe5edb8011d62b8e5d66fef749ce124",
      "size": 5436,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:6eba444dd58a8748225335c4dcc53232c53d8252370f4a01ecfa6e692925db73",
      "size": 568,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:0298f7c3ceec45da42b10f52b8edd4c5ebe5edb8011d62b8e5d66fef749ce124",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": {
        "architecture": "unknown",
        "os": "unknown"
      }

```

Some container images (e.g., ghcr.io/llm-d/llm-d-cuda-dev:pr-230) include attestation manifests alongside runnable image manifests. These attestation manifests have:                                                   
  - Platform architecture set to "unknown"                                                                                                                                                                                
  - Platform OS set to "unknown"                                                                                                                                                                                          
  - An [annotation](https://docs.docker.com/build/metadata/attestations/attestation-storage/#attestation-manifest-descriptor) vnd.docker.reference.type: attestation-manifest                                                                                                                                                         
                                                                                                                                                                                                                          
  The operator was processing these attestation manifests as valid image manifests, leading to:                                                                                                                           
  1. "unknown" being added to the set of supported architectures                                                                                                                                                          
  2. Node affinity requirements being set for kubernetes.io/arch=unknown                                                                                                                                                  
  3. Pods failing to schedule since no nodes match this architecture

### DIFF
--- a/pkg/image/inspector.go
+++ b/pkg/image/inspector.go
@@ -146,15 +146,39 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 			return nil, err
 		}
 		for _, m := range index.Manifests {
+			// Skip manifests with Docker reference annotations - they are not runnable platform images
+			// Per Docker spec, these annotations indicate special manifest types (e.g., attestation-manifest)
+			// vnd.docker.reference.type: indicates the manifest type (e.g., "attestation-manifest")
+			// vnd.docker.reference.digest: points to the subject manifest being attested
+			// If either annotation is present, the manifest should be ignored as it's not a platform image
+			if m.Annotations != nil {
+				if refType, exists := m.Annotations["vnd.docker.reference.type"]; exists {
+					log.V(3).Info("Skipping manifest with reference type annotation", "type", refType, "digest", m.Digest)
+					continue
+				}
+				if refDigest, exists := m.Annotations["vnd.docker.reference.digest"]; exists {
+					log.V(3).Info("Skipping manifest with reference digest annotation", "refDigest", refDigest, "digest", m.Digest)
+					continue
+				}
+			}
+			// Skip manifests with unknown architecture
+			if m.Platform == nil || (m.Platform.Architecture == "unknown") {
+				log.V(3).Info("Skipping manifest with unknown platform", "architecture", m.Platform.Architecture, "os", m.Platform.OS, "digest", m.Digest)
+				continue
+			}
 			supportedArchitectures = sets.Insert(supportedArchitectures, m.Platform.Architecture)
+			// Store the first valid manifest digest for bundle image detection
+			if instanceDigest == nil {
+				instanceDigest = &m.Digest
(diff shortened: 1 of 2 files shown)


pkg/image/inspector.go
```diff
@@ -146,15 +146,29 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
 			return nil, err
 		}
 		for _, m := range index.Manifests {
+			// Skip attestation manifests - they are not runnable images
+			if m.Annotations != nil && m.Annotations["vnd.docker.reference.type"] == "attestation-manifest" {
+				log.V(3).Info("Skipping attestation manifest", "digest", m.Digest)
+				continue
+			}
+			// Skip manifests with unknown architecture or OS
```
lwan-wanglin: The comment says we skip manifests with unknown architecture or OS, but the code only checks for Architecture == "unknown".

AnnaZivkovic:
I am not checking os because I dont think it is relevant to getting the arch, but we need to check vnd.docker.reference.type and vnd.docker.reference.digest because both of those tags indicate a non runnable manifests.

pkg/image/inspector.go
```diff
@@ -146,15 +146,39 @@ func (i *registryInspector) GetCompatibleArchitecturesSet(ctx context.Context, i
...
+					log.V(3).Info("Skipping manifest with reference type annotation", "type", refType, "digest", m.Digest)
+					continue
+				}
+				if refDigest, exists := m.Annotations["vnd.docker.reference.digest"]; exists {
+					log.V(3).Info("Skipping manifest with reference digest annotation", "refDigest", refDigest, "digest", m.Digest)
+					continue
+				}
+			}
+			// Skip manifests with unknown architecture
+			if m.Platform != nil && (m.Platform.Architecture == "unknown") {
```
lwan-wanglin:
```suggestion
			if m.Platform == nil || m.Platform.Architecture == "unknown" {
```
if m.Platform == nil, below `supportedArchitectures = sets.Insert(supportedArchitectures, m.Platform.Architecture)` would hit panic

PR #725, opened 2026-01-29
MULTIARCH-5369: Add PodPlacementConfig plugin detection

AnnaZivkovic:
Fix PodPlacementConfig plugin detection and eliminate code duplication                                                                                                                                                          
                                                                                                                                                                                                                                  Problem                                                                                                                                                                                                                         
                                                                                                                                                                                                                                  The namespace-scoped PodPlacementConfig resources with the NodeAffinityScoring plugin enabled would be ignored if the cluster-scoped ClusterPodPlacementConfig had the plugin disabled.                                                                                                                                                                                                                      
                                                                                                                                                                                                                                  
  Bug scenario:                                                                                                                                                                                                                   
  1. ClusterPodPlacementConfig has NodeAffinityScoring disabled                                                                                                                                                                   
  2. Namespace-scoped PodPlacementConfig has NodeAffinityScoring enabled                                                                                                                                                          
  3. Pod matching the PPC's label selector is created                                                                                                                                                                             
  4. Expected: Pod gets preferred node affinity from the PPC                                                                                                                                                                      
  5. Actual: Pod's preferred affinity is skipped entirely because the system only checked CPPC's plugin status

### DIFF
--- a/internal/controller/operator/clusterpodplacementconfig_controller_test.go
+++ b/internal/controller/operator/clusterpodplacementconfig_controller_test.go
@@ -451,7 +451,7 @@ var _ = Describe("internal/Controller/ClusterPodPlacementConfig/ClusterPodPlacem
 					framework.NewConditionTypeStatusTuple(v1beta1.DegradedType, corev1.ConditionTrue),
 					framework.NewConditionTypeStatusTuple(v1beta1.PodPlacementControllerNotRolledOutType, corev1.ConditionTrue),
 					framework.NewConditionTypeStatusTuple(v1beta1.PodPlacementWebhookNotRolledOutType, corev1.ConditionFalse),
-					framework.NewConditionTypeStatusTuple(v1beta1.MutatingWebhookConfigurationNotAvailable, corev1.ConditionFalse),
+					framework.NewConditionTypeStatusTuple(v1beta1.MutatingWebhookConfigurationNotAvailable, corev1.ConditionTrue),
 					framework.NewConditionTypeStatusTuple(v1beta1.DeprovisioningType, corev1.ConditionFalse),
 				)).Should(Succeed(), "the ClusterPodPlacementConfig should have the correct conditions")
 				By("Verify no mutating webhook configuration is not available")
--- a/pkg/e2e/operator/pod_placement_config_test.go
+++ b/pkg/e2e/operator/pod_placement_config_test.go
@@ -1029,7 +1029,7 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
 			}).Should(Succeed(), "the ClusterPodPlacementConfig should remove the no-pod-placement-config finalizer")
 		})
 	})
-	Context("when CPPC has NodeAffinityScoring disabled but PPC has it enabled", func() {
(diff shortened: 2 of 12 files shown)


AnnaZivkovic:
Depends on https://github.com/outrigger-project/multiarch-tuning-operator/pull/676
/hold

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -1029,4 +1029,231 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
...
+
+			By("The pod should NOT have preferred affinities (CPPC disabled, PPC doesn't match)")
+			Eventually(framework.VerifyPodPreferredNodeAffinity(ctx, client, ns, "app", "frontend", nil), e2e.WaitShort).Should(Succeed())
+		})
+		It("should respect PPC priority when multiple PPCs match a pod", func() {
+			By("Creating a ClusterPodPlacementConfig with NodeAffinityScoring disabled")
+			err := client.Create(ctx,
+				NewClusterPodPlacementConfig().
+					WithName(common.SingletonResourceObjectName).
+					WithNodeAffinityScoring(false).
```
lwan-wanglin, line 1184: platforms is always required, even when NodeAffinityScoring is disabled. https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/api/common/plugins/nodeaffinityscoring_plugin.go#L32

internal/controller/operator/clusterpodplacementconfig_controller_test.go
```diff
@@ -818,6 +818,71 @@ var _ = Describe("internal/Controller/ClusterPodPlacementConfig/ClusterPodPlacem
 				g.Expect(cppc.Finalizers).NotTo(ContainElement(utils.CPPCNoPPCObjectFinalizer))
 			}).Should(Succeed(), "the ClusterPodPlacementConfig should remove the no-pod-placement-config finalizer")
 		})
+		It("should remove the finalizer during ClusterPodPlacementConfig deletion when no PodPlacementConfigs exist", func() {
```
lwan-wanglin: We already have the same or a similar test case here.  [It("should add the finalizer when a namespaced PodPlacementConfig exists and remove it when none exist"](https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/internal/controller/operator/clusterpodplacementconfig_controller_test.go#L767-L821)

pkg/e2e/operator/pod_placement_config_test.go
```diff
@@ -1125,5 +1028,305 @@ var _ = Describe("The Multiarch Tuning Operator", Serial, func() {
 				g.Expect(cppc.Finalizers).NotTo(ContainElement(utils.CPPCNoPPCObjectFinalizer))
 			}).Should(Succeed(), "the ClusterPodPlacementConfig should remove the no-pod-placement-config finalizer")
 		})
+		It("should remove the finalizer during ClusterPodPlacementConfig deletion when no PodPlacementConfigs exist", func() {
```
lwan-wanglin: This is the same situation as above; we already have the same or a similar test case here. https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/pkg/e2e/operator/pod_placement_config_test.go#L1068

PR #726, opened 2026-02-03
MULTIARCH-5642:  Remove uporcessed enoexecEvents on error 

AnnaZivkovic:
In the ENoExecEvent reconciler it is possible that an error could occur which would result in the ENoExecEvent CR to not be deleted. During the uninstall/disabling of the execFormatErrorMontior plugin these left over CR will prevent the removal of the deployment. 

When an ENoExecEvent fails to reconcile, it will now be labeled as an error. The cleanup logic for the execFormatErrorMontior will now bypass these errored CRs when the plugin is disabled, ensuring a clean shutdown. Disabling the ENoExecEvent plugin will remove all remaining ENoExecEvent resources, regardless of their state.

### DIFF
--- a/internal/controller/enoexecevent/handler/enoexecevent_controller.go
+++ b/internal/controller/enoexecevent/handler/enoexecevent_controller.go
@@ -68,76 +68,92 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	logger := log.FromContext(ctx)
 	metrics.InitMetrics()
 	// Fetch the ENoExecEvent instance
-	enoExecEvent := &multiarchv1beta1.ENoExecEvent{}
-	if err := r.Get(ctx, req.NamespacedName, enoExecEvent); err != nil {
+	enoExecEventObj := &multiarchv1beta1.ENoExecEvent{}
+	if err := r.Get(ctx, req.NamespacedName, enoExecEventObj); err != nil {
 		// If the resource is not found, we simply return. This is not an error.
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	if enoExecEvent.Namespace != utils.Namespace() {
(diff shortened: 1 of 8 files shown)


internal/controller/enoexecevent/handler/enoexecevent_controller.go
```diff
@@ -68,76 +68,92 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		// Return nil to avoid requeuing - this event will be cleaned up by deleteErroredENoExecEvents
 		return ctrl.Result{}, nil
 	}
 
-	if enoExecEvent.Status.PodName == "" {
+	if eNoExecEvent.Status.PodName == "" {
 		// If the ENoExecEvent does not have a pod name, we ignore it
-		logger.V(5).Info("ENoExecEvent does not have a pod name, skipping reconciliation", "name", enoExecEvent.Name, "namespace", enoExecEvent.Namespace)
+		logger.V(5).Info("ENoExecEvent does not have a pod name, skipping reconciliation", "name", eNoExecEvent.Name, "namespace", eNoExecEvent.Namespace)
+		r.markAsError(ctx, eNoExecEvent, ErrorReasonNodeNotFound)
```
lwan-wanglin:
```suggestion
		r.markAsError(ctx, eNoExecEvent, ErrorReasonPodNotFound)
```

internal/controller/enoexecevent/handler/enoexecevent_controller_test.go
```diff
@@ -476,5 +500,55 @@ var _ = Describe("internal/Controller/ENoExecEvent/Reconciler", func() {
 				Expect(err).To(HaveOccurred(), "Should not update enne status with containerID that has invalid format", err)
 			})
 		})
+		Context("handles reconciliation errors with error labels", func() {
+			It("should mark the ENoExecEvent with error label when node is not found", func() {
```
lwan-wanglin, line 537: I didn’t find any verification of the error label in the test steps below.

lwan-wanglin:
test sast

PR #735, opened 2026-03-19
Remove upsteam Tekton configurations

lwan-wanglin:
Remove upsteam Tekton configurations


AnnaZivkovic:
Should we even have konflux in the upstream version? If we want to keep it should we make try to make it runnable?

PR #737, opened 2026-03-31
MULTIARCH-6069: proposal for CEL expression placement plugin

prb112:


docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+
+## Motivation
+
+The current Multiarch Tuning Operator automatically determines Pod image architecture compatibility by inspecting container images. While this works well for most scenarios, there are cases where administrators need more control at the namespace level:
+
+1. **Workload-specific architecture preferences** Some workloads may benefit from specific architectures based on their component affinnity (e.g., database pods on ppc64le, web servers on amd64)
```
lwan-wanglin: a typo
```suggestion
1. **Workload-specific architecture preferences** Some workloads may benefit from specific architectures based on their component affinity (e.g., database pods on ppc64le, web servers on amd64)
```

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+#### Pod Admission and Processing
+
+1. A pod is submitted to the API server in a namespace with a `PodPlacementConfig` containing `celArchitecturePlacement`
+2. The MTO mutating webhook adds a scheduling gate (existing behavior)
+3. The Pod Placement Controller processes the pod:
+   - Retrieves all `PodPlacementConfig` resources in the pod's namespace
+   - Filters configs whose `labelSelector` matches the pod
+   - For matching configs with `celArchitecturePlacement` enabled, evaluates CEL expressions in priority order
+   - If a rule matches (CEL expression returns `true`):
+     - **Removes any existing architecture constraints** from the pod's `spec.nodeSelector` (removes `kubernetes.io/arch` key if present)
```
lwan-wanglin: If a user creates a namespaced PPC and enables celArchitecturePlacement, it can override all of `spec.nodeSelector` and `spec.affinity.nodeAffinity`, including user-defined settings, not just those added by MTO？

lwan-wanglin: Ah, I see the answer in the following section

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+- System namespaces matching `kube-*` pattern (unless explicitly configured)
+- Namespaces matching `hypershift-*` pattern (unless explicitly configured)
+
+### Default Architectures Rationale
+
+The `defaultArchitectures` field is required for several important reasons:
+
+1. **Limits Rule Explosion and Configuration Burden** Without a default, administrators would need to create rules for every possible pod pattern, leading to complex and hard-to-maintain configurations. The plugin supports _exceptional_ cases in the same namespace.
+
+2. **Provides Fallback Behavior** When no rules match, the system needs a sensible default rather than failing or using arbitrary behavior
```
lwan-wanglin: we now have a field named fallbackArchitecture in Cluster scope ppc https://github.com/outrigger-project/multiarch-tuning-operator/blob/main/api/v1beta1/clusterpodplacementconfig_types.go#L58, If both fallbackArchitecture and defaultArchitectures are set, does defaultArchitectures take precedence?

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+
+The example sets a default architecture for amd64, and a selection of a matching Pods to place on ppc64le and another rule to place on amd64 or arm64.
+
+```yaml
+apiVersion: multiarch.openshift.io/v1beta1
+kind: PodPlacementConfig
+metadata:
+  name: database-rules
+  namespace: production
+spec:
```
AnnaZivkovic, line 192: This is missing the Priority field which will default to 0 if left empty.

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+
+3. **Image Inspection** The `celArchitecturePlacement` plugin takes precedence over image-based detection when enabled. Existing architecture constraints (including those set by image inspection) are removed.
+
+4. **NodeAffinityScoring Plugin** Can coexist with `celArchitecturePlacement` in the same `PodPlacementConfig`:
+   - `celArchitecturePlacement` determines which architectures are eligible (removes old constraints, sets new ones)
+   - `NodeAffinityScoring` determines preferences among the eligible architectures
+
+#### Priority and Conflict Resolution
+
+Only a single `PodPlacementConfig` resource in the same namespace is allowed.
```
AnnaZivkovic: I feel like this disregards the purpose of the Priority field. We would want multiple PPC with different priories. If several configs match the pod via labelSelector, is the winner only by priority, or is there a tie-breaker (name, creation time)?

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+3. The first matching rule from the highest priority config determines the architecture
+4. If no rules match in any config, the `defaultArchitectures` from the highest priority config with `celArchitecturePlacement` enabled is used
+5. In all cases, existing architecture constraints are removed before new ones are applied
+
+#### Performance Considerations
+
+1. **CEL Compilation** Expressions are compiled once at configuration time
+2. **Expression Caching** Compiled expressions are cached to avoid repeated compilation
+3. **Evaluation Overhead** CEL evaluation is fast (microseconds per expression)
+4. **Rule Limit** Maximum of 500 rules per configuration to prevent excessive evaluation time. This limit will be reviewed after use.
```
AnnaZivkovic: There are conflicting numbers for the rule limit. Is it 50, 500. or 1000+?

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+
+### Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| Complex CEL expressions cause evaluation errors | Pods may not be scheduled correctly | Validate expressions at admission time; provide clear error messages; treat evaluation errors as non-matches |
+| Misconfigured rules assign pods to incompatible architectures | Pods fail to start with ENOEXEC errors | Document best practices; recommend testing rules in non-production environments; existing ENOEXEC monitoring will detect issues |
+| Too many rules impact performance | Increased pod scheduling latency | Limit maximum rules per configuration (500); compile and cache expressions; provide performance guidelines |
+| Conflicting rules between multiple PodPlacementConfigs | Unpredictable behavior | Clear precedence rules based on priority field; document evaluation order |
+| CEL expressions access sensitive pod data | Potential information disclosure | CEL expressions only have access to pod metadata (labels, annotations, name, namespace); no access to secrets or container specs |
```
AnnaZivkovic: The risk table says CEL only sees pod metadata and not container specs. Elsewhere the design is “evaluate against a Pod resource” with self. If the real CEL type is a full Pod, expressions could reference images, env, volumes, etc., unless the implementation uses a restricted type or field mask.

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+    affinity:
+      nodeAffinity:
+        requiredDuringSchedulingIgnoredDuringExecution:
+          nodeSelectorTerms:
+          - matchExpressions:
+            - key: kubernetes.io/arch
+                operator: In
+                values:
+                - ppc64le
+                - amd64
```
AnnaZivkovic:
```suggestion
          - key: kubernetes.io/arch
            operator: In
            values:
            - ppc64le
            - amd64
```

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+   - Filters configs whose `labelSelector` matches the pod
+   - For matching configs with `celArchitecturePlacement` enabled, evaluates CEL expressions in priority order
+   - If a rule matches (CEL expression returns `true`):
+     - **Removes any existing architecture constraints** from the pod's `spec.nodeSelector` (removes `kubernetes.io/arch` key if present)
+     - **Removes any existing architecture-based node affinity** from the pod's `spec.affinity.nodeAffinity` (removes node selector terms with `kubernetes.io/arch` match expressions)
+     - **Sets new node affinity** with the rule's target architecture list
+   - If no rules match, uses the default architecture list specified in the plugin configuration (also removing existing constraints)
+   - The scheduling gate is removed
+
+#### Architecture Constraint Removal
```
AnnaZivkovic, line 167: The proposal states that celArchitecturePlacement removes existing arch constraints and sets required affinity, that image-based detection can still apply elsewhere, and that NodeAffinityScoring can coexist and “prefer among eligible architectures.” Could you document an explicit ordering (and which component runs in which stage) for a single reconcile pass? Without that, it is ambiguous whether image-based logic might run before or after CEL and whether scoring sees the final required arch set or an intermediate state, which makes behavior hard to predict and tests hard to specify.

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,795 @@
...
+3. **Evaluation Overhead** CEL evaluation is fast (microseconds per expression)
+4. **Rule Limit** Maximum of 500 rules per configuration to prevent excessive evaluation time. This limit will be reviewed after use.
+5. **Namespace Scope** Only configs in the pod's namespace are evaluated, limiting the search space
+6. **Constraint Removal** Removing existing constraints is a fast operation (map/slice manipulation)
+
+### Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|--------|------------|
+| Complex CEL expressions cause evaluation errors | Pods may not be scheduled correctly | Validate expressions at admission time; provide clear error messages; treat evaluation errors as non-matches |
```
AnnaZivkovic, line 857: Implementation notes say evaluation errors are logged and treated as false. How do we handle validation on PodPlacementConfig create/update (reject bad CEL), and whether any pod-level admission failure is possible, or failures are always soft (fall through to default / next rule)?

prb112: addressed in next commit

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: resolved

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,1207 @@
...
+
+#### Changes to PodPlacementConfig
+
+The [`PodPlacementConfig`](../../api/v1beta1/podplacementconfig_types.go) CRD will be extended to support the new `celArchitecturePlacement` plugin in the `LocalPlugins` struct:
+
+```go
+// LocalPlugins represents the plugins configuration for podplacementconfigs resource.
+type LocalPlugins struct {
+    NodeAffinityScoring *NodeAffinityScoring `json:"nodeAffinityScoring,omitempty"`
+    celArchitecturePlacement *celArchitecturePlacement `json:"celArchitecturePlacement,omitempty"`
```
AnnaZivkovic:
```suggestion
  CELArchitecturePlacement *CELArchitecturePlacement `json:"celArchitecturePlacement,omitempty"`
```

The celArchitecturePlacement struct and its field in LocalPlugins are lowercase (unexported). Go's JSON serializer ignores unexported fields, and kubebuilder can't generate CRD schemas from them. The existing NodeAffinityScoring plugin uses exported names.

docs/enhancements/MTO-0005-architecture-rules-plugin.md
AnnaZivkovic: Can we updater or remove this?

docs/enhancements/MTO-0005-architecture-rules-plugin.md
```diff
@@ -0,0 +1,1207 @@
...
+  plugins:
+    celArchitecturePlacement:
+      enabled: true
+      fallbackArchitectures:
+        - amd64
+      rules:
+        - name: postgres-on-ppc64le
+          expression: |
+            self.metadata.labels.exists(l, l.key == 'app.kubernetes.io/component' && l.value == 'database') &&
+            self.metadata.labels.exists(l, l.key == 'app.kubernetes.io/part-of' && l.value == 'postgresql')
```
AnnaZivkovic: With Kubernetes protobuf types and cel-go, labels are a map<string, string>, so the idiomatic access would be  `app.kubernetes.io/component' in self.metadata.labels && self.metadata.labels['app.kubernetes.io/component'] == 'database` 
Or with the has() macro. The .exists(l, l.key == ...) pattern works on lists, not maps. This should be validated against the actual cel-go Pod type binding and examples corrected accordingly.

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: added all over the doc (next push)

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: Done

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: fixed in latest

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: switched to referencing fallbackArchitecture

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: Yes, I had the thought about spec.containers[].image could be matched and trigger the cel expression. Thoughts?

docs/enhancements/MTO-0005-architecture-rules-plugin.md
prb112: done

PR #740, opened 2026-04-14
MULTIARCH-6236: Check for PodPlacementConfigs before tearing down operand resources

AnnaZivkovic:
Move the PPC existence check before deleting webhook/operand resources in handleDelete to prevent orphaning namespaced PodPlacementConfigs.

Previously, if admission was unavailable or deletion bypassed the webhook, the deletion flow could tear down the webhook and operand while PPCs still existed, leaving them orphaned without the resources they depend on.

Now the deletion is properly gated:
1. Check if CPPCNoPPCObjectFinalizer exists
2. List all PodPlacementConfigs
3. Abort with RemainingPodPlacementConfig error if any exist
4. Only proceed to delete resources after confirming zero PPCs

This enforces ordered deletion as required by the coding guidelines: pods must be ungated before operand removal

### DIFF
--- a/internal/controller/operator/clusterpodplacementconfig_controller.go
+++ b/internal/controller/operator/clusterpodplacementconfig_controller.go
@@ -343,6 +343,32 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 	// We execute the update here because this function returns multiple times before the whole deletion process is completed.
 	// Executing it here ensures that the conditions are updated throughout the deletion process.
 	_ = r.updateStatus(ctx, clusterPodPlacementConfig)
+
+	// Prevent deletion while PodPlacementConfig resources still exist
+	// Check this BEFORE tearing down webhook/operand resources to prevent orphaning PPCs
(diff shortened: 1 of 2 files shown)


internal/controller/operator/clusterpodplacementconfig_controller.go
```diff
@@ -343,6 +343,31 @@ func (r *ClusterPodPlacementConfigReconciler) handleDelete(ctx context.Context,
 	// We execute the update here because this function returns multiple times before the whole deletion process is completed.
 	// Executing it here ensures that the conditions are updated throughout the deletion process.
 	_ = r.updateStatus(ctx, clusterPodPlacementConfig)
+
+	// Prevent deletion while PodPlacementConfig resources still exist
+	// Check this BEFORE tearing down webhook/operand resources to prevent orphaning PPCs
```
jcpowermac, line 348: could there be a unit test?

AnnaZivkovic: There was an integration test. I updated it to look at the webhook as well

AnnaZivkovic:
/override ci/prow/ocp418-e2e-azure

Fix is merged but this pr is not rebased

PR #745, opened 2026-06-24
MULTIARCH-5646: Fix security findings

AnnaZivkovic:


jcpowermac (commented):
## Security Findings Review

_Review by Claude (AI-assisted) on behalf of @jcallen_

Overall, the security hardening is well-directed — RBAC scoping, hostPath narrowing, input validation, and hash upgrade are all good changes. A few items below worth addressing.

### Summary of Inline Comments

1. **Bug — `DockerCertsDir()` caching broken** (`pkg/image/systemconfig.go`): The empty default defeats the double-checked locking cache, causing every call to take the write-lock path. Small fix needed.
2. **Missing test for invalid architecture filtering** (`pod_model.go`): The new `validArchitectures.Has(value)` guard has no test case with an unexpected architecture string.
3. **`HostPathDirectory` for `containers-conf`** (`podplacement_objects.go`): Worth documenting that `/etc/containers/` must exist on nodes — fine for OpenShift, but a constraint for vanilla K8s.
4. **Minor — `roles`/`rolebindings` not scoped with `resourceNames`**: ClusterRoles/Bindings got `resourceNames` scoping but namespace-scoped Roles/RoleBindings did not. Lower priority since blast radius is limited to the operator namespace.
5. **Minor — byte-based error truncation** (`pod_model.go`): `errMsg[:256]` could split a multi-byte UTF-8 rune. Unlikely in practice but a rune-aware truncation would be cleaner.

pkg/image/systemconfig.go
```diff
@@ -41,7 +41,7 @@ func DockerCertsDir() string {
 	defer rwMutex.Unlock()
 	if dockerCertsDir == "" {
 		// avoid race condition in-between rwMutex.RUnlock and rwMutex.Lock
```
jcpowermac: **Bug — caching broken with empty default**

_Review by Claude (AI-assisted) on behalf of @jcallen_

Changing the default from `"/etc/docker/certs.d"` to `""` breaks the double-checked locking cache pattern used throughout this file. The fast-path check is `if dockerCertsDir != ""` — when the default is empty and the env var is unset, `dockerCertsDir` stays `""` forever, so every call takes the write-lock path and re-calls `os.LookupEnv`.

```go
// Fast path never triggers when dockerCertsDir == ""
rwMutex.RLock()
if dockerCertsDir != "" {   // always false with empty default
    defer rwMutex.RUnlock()
    return dockerCertsDir
}
```

This won't cause incorrect behavior (the caller checks `dockerCerts != ""` and skips), but it defeats caching and introduces unnecessary write-lock contention on every image inspection cache miss with `MaxConcurrentReconciles = NumCPU * 4`.

**Suggested fix:** Use `sync.Once` for initialization, or use a sentinel value (e.g. a private `const unset = "\x00"`) initialized in the `var` block, and map it back to `""` on return.

internal/controller/podplacement/pod_model.go
```diff
@@ -416,8 +416,11 @@ func (pod *Pod) ensureArchitectureLabels(requirement corev1.NodeSelectorRequirem
 	default:
 		pod.EnsureLabel(utils.MultiArchLabel, "")
 	}
+	validArchitectures := utils.AllSupportedArchitecturesSet()
 	for _, value := range requirement.Values {
-		pod.EnsureLabel(utils.ArchLabelValue(value), "")
+		if validArchitectures.Has(value) {
```
jcpowermac, line 421: **Missing test for invalid architecture filtering**

_Review by Claude (AI-assisted) on behalf of @jcallen_

The new `validArchitectures.Has(value)` guard is untested. The existing `TestEnsureArchitectureLabels` only uses valid architecture strings (`ArchitectureAmd64`, `ArchitectureArm64`). A test case with an unexpected value like `"riscv64"` or an adversarial string should confirm it gets dropped and no label is created for it.

internal/controller/operator/podplacement_objects.go
```diff
@@ -126,7 +126,7 @@ func buildControllerDeployment(clusterPodPlacementConfig *v1beta1.ClusterPodPlac
 			VolumeSource: corev1.VolumeSource{
```
jcpowermac, line 126: **`HostPathDirectory` may prevent scheduling on some nodes**

_Review by Claude (AI-assisted) on behalf of @jcallen_

Changing from `HostPathDirectoryOrCreate` to `HostPathDirectory` means the pod-placement-controller won't start on nodes where `/etc/containers/` doesn't exist. This is fine on OpenShift (always present), but worth documenting as a known constraint if this operator is intended to work on vanilla Kubernetes.

internal/controller/podplacement/pod_model.go
```diff
@@ -497,10 +500,14 @@ func (pod *Pod) handleError(err error, s string) {
 	}
 	log := ctrllog.FromContext(pod.Ctx())
 	metrics.FailedInspectionCounter.Inc()
+	errMsg := err.Error()
+	if len(errMsg) > 256 {
+		errMsg = errMsg[:256]
+	}
```
jcpowermac, line 507: **Minor — byte-based truncation**

_Review by Claude (AI-assisted) on behalf of @jcallen_

`errMsg[:256]` slices bytes, which could split a multi-byte UTF-8 rune. In practice, K8s annotations accept arbitrary bytes and error messages are typically ASCII, so unlikely to matter. If you want to be precise:
```go
runes := []rune(errMsg)
if len(runes) > 256 {
    errMsg = string(runes[:256])
}
```

internal/controller/operator/clusterpodplacementconfig_controller.go
```diff
@@ -94,10 +107,14 @@ const (
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch;update;patch;create;delete
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts/status,verbs=get
 //+kubebuilder:rbac:groups=core,resources=serviceaccounts/finalizers,verbs=update
```
jcpowermac, line 121: **Minor — `roles`/`rolebindings` not scoped with `resourceNames`**

_Review by Claude (AI-assisted) on behalf of @jcallen_

ClusterRoles and ClusterRoleBindings got `resourceNames` scoping (FIND-003), but namespace-scoped Roles and RoleBindings on line 109 did not. The blast radius is limited to the operator namespace so this is lower priority, but it's an inconsistency that could be addressed in a follow-up for completeness.

PR #746, opened 2026-06-24
MULTIARCH-6069: Refine CEL logic for tuning operator enhancements

pranitaT:


AnnaZivkovic:
Hey @pranitaT  several CI checks are failing (fmt, generate, goimports, lint, security, unit, vendor). These usually indicate that generated files or vendored dependencies are out of sync.

Could you try running the following locally and pushing any changes?

```bash
make manifests
make generate
make bundle
make vendor
make fmt
make goimports
```

Also make sure you're using operator-sdk v1.33

pkenchap:
@pranitaT , Please rebase the PR .

internal/controller/podplacement/cel_controller_integration_test.go
```diff
@@ -0,0 +1,626 @@
...
+			}
+			Expect(archTermCount).To(Equal(1), "should have exactly one architecture term")
+
+			// ResourceVersion should have changed (pod was updated), but affinity should be stable
+			Expect(pod.ResourceVersion).NotTo(Equal(initialResourceVersion), "resource version should change on updates")
+		})
+	})
+})
+
+// Made with Bob
```
pkenchap: remove this line

internal/controller/podplacement/architecture_inplace_update_test.go
```diff
@@ -0,0 +1,313 @@
...
+				foundOS = true
+			}
+		}
+	}
+	if !foundOS {
+		t.Error("OS constraint should be preserved")
+	}
+}
+
+// Made with Bob
```
pkenchap: remove

internal/controller/podplacement/cel_coexistence_test.go
```diff
@@ -0,0 +1,205 @@
...
+			celApplied := true
+			Expect(celApplied).To(BeTrue(), "CEL applied flag should be true")
+
+			// In the actual reconciler, this would prevent image detection:
+			// if !celApplied { /* image detection */ }
+		})
+	})
+})
+
+// Made with Bob
```
pkenchap: remove this line

internal/controller/podplacement/cel_no_fallback_merge_test.go
```diff
@@ -0,0 +1,327 @@
...
+				g.Expect(archExpression.Values).To(Equal([]string{utils.ArchitectureS390x}),
+					"should contain ONLY s390x from CEL rule, no CPPC fallback")
+				g.Expect(len(archExpression.Values)).To(Equal(1),
+					"should have exactly one architecture value")
+			}).WithTimeout(timeout).WithPolling(interval).Should(Succeed())
+		})
+	})
+})
+
+// Made with Bob
```
pkenchap: remove this line

internal/controller/podplacement/cel_stale_arch_regression_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if !foundExclusivePPC64LE {
+		t.Fatalf("expected to find an exclusive %q architecture expression after reapply", utils.ArchitecturePpc64le)
+	}
+	if !foundOSConstraint {
+		t.Fatal("expected existing os constraint to be preserved in merged required term")
+	}
+}
+
+// Made with Bob
```
pkenchap: remove this line

internal/controller/podplacement/webhook_cel_test.go
```diff
@@ -0,0 +1,395 @@
...
+				}
+			}
+		}
+	}
+	if !found {
+		t.Error("Architecture requirement from highest priority PPC not found")
+	}
+}
+
+// Made with Bob
```
pkenchap: remove this line

internal/controller/podplacement/cel_no_fallback_merge_test.go
```diff
@@ -0,0 +1,327 @@
...
+				FallbackArchitectures: []string{
+					utils.ArchitectureAmd64,
+					utils.ArchitectureArm64,
+					utils.ArchitecturePpc64le,
+					utils.ArchitectureS390x,
+				},
+				Rules: []plugins.ArchitectureRule{
+					{
+						Name:       "match-database-ppc64le-only",
+						Expression: `self.metadata.labels.exists(l, l.key == "component" && l.value == "database")`,
```
pkenchap: There is an issue with this line might will fix your Unit test . 
Here is the fix for it .

Expression: `has(self.metadata.labels.component) && self.metadata.labels.component == "database"`,

internal/controller/podplacement/cel_controller_integration_test.go
```diff
@@ -0,0 +1,626 @@
...
+
+			ppc.Spec.Plugins.CelArchitecturePlacement = &plugins.CelArchitecturePlacement{
+				BasePlugin: plugins.BasePlugin{
+					Enabled: true,
+				},
+				FallbackArchitectures: []string{utils.ArchitectureAmd64},
+				Rules: []plugins.ArchitectureRule{
+					{
+						Name:       "match-database",
+						Expression: `self.metadata.labels.exists(l, l.key == "component" && l.value == "database")`,
```
pkenchap: same -
Expression: has(self.metadata.labels.component) && self.metadata.labels.component == "database",

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,244 @@
...
+	if initErr != nil {
+		return nil, initErr
+	}
+	return packageEvaluator, nil
+}
+
+// newCELEvaluator creates a new CEL evaluator with a Pod-aware environment
+func newCELEvaluator() (*celEvaluator, error) {
+	env, err := cel.NewEnv(
+		cel.Variable("self", cel.DynType),
```
AnnaZivkovic, line 70: The enhancement document explicitly requires that CEL expressions may only reference self.metadata.* and that any reference to self.spec or self.status must be rejected at admission time via an `assertMetadataOnly` check. I am not seeing this check in the codebase?

AnnaZivkovic, line 70: With DynType, I dont think the compiler validate field references so expressions like self.spec.containers or self.metadata.labells (which is  a typo) compile without error. At runtime, podToMap() only exposes metadata fields, so these silently fail and get treated as false so the user never finds out their expression is broken. 

The enhancement specifies using a typed binding (cel.Types(&corev1.Pod{}) with cel.ObjectType("k8s.io.api.core.v1.Pod")) so the compiler can reject invalid field references, plus assertMetadataOnly AST walk to reject self.spec.*/self.status.* even though they're valid Pod fields.

What are your thought on changing to Typed CEL environment & AST check . So we would switch to cel.ObjectType(...) so field references are schema-validated at compile time (typos caught), then add assertMetadataOnly to walk the AST and reject self.spec/self.status access.

internal/controller/podplacement/cel_integration.go
```diff
@@ -0,0 +1,96 @@
...
+			"pod", pod.Name,
+			"fallbackArchitectures", result.architectures)
+	}
+
+	log.Info("Applying architecture constraints from CEL",
+		"pod", pod.Name,
+		"architectures", result.architectures)
+
+	// Remove existing architecture constraints and apply new ones
+	applyArchitectureConstraints(pod.PodObject(), result.architectures)
```
AnnaZivkovic: Do pods processed by CEL will have NodeAffinityLabel = "not-set"?  we want the node affinity labels to be set correctly

internal/controller/podplacement/pod_model.go
```diff
@@ -435,11 +469,122 @@ func (pod *Pod) ensureArchitectureLabels(requirement corev1.NodeSelectorRequirem
...
 func (pod *Pod) shouldIgnorePod(cppc *v1beta1.ClusterPodPlacementConfig, matchingPPCs []v1beta1.PodPlacementConfig) bool {
-	return utils.Namespace() == pod.Namespace || strings.HasPrefix(pod.Namespace, "kube-") ||
-		pod.Spec.NodeName != "" || pod.HasControlPlaneNodeSelector() || pod.IsFromDaemonSet() ||
-		pod.isNodeSelectorConfiguredForArchitecture() &&
-			(pod.isPreferredAffinityConfiguredForArchitecture() ||
-				(!cppc.PluginsEnabled(common.NodeAffinityScoringPluginName) && !pod.hasMatchingPPCWithPlugin(matchingPPCs)))
+	log := ctrllog.FromContext(pod.Ctx())
+
+	cppcPresent := cppc != nil
+	log.Info("[SHOULD_IGNORE] ENTER",
```
AnnaZivkovic: Logging changed from V(3) to Info(). This will flood production logs. Can we change this back? for more verbose logs we can change the log level in the CPPC

internal/controller/podplacement/pod_model.go
```diff
@@ -358,32 +358,66 @@ func (pod *Pod) imagesNamesSet() sets.Set[containerImage] {
 func (pod *Pod) intersectImagesArchitecture(pullSecretDataList [][]byte) (supportedArchitectures []string, err error) {
 	log := ctrllog.FromContext(pod.Ctx())
 	imageNamesSet := pod.imagesNamesSet()
-	log.V(1).Info("Images list for pod", "imageNamesSet", fmt.Sprintf("%+v", imageNamesSet))
+	log.Info("Image inspection: Starting", "pod", pod.Name, "namespace", pod.Namespace, "imageCount", imageNamesSet.Len())
```
AnnaZivkovic: Same logging change. Can we change this back?

internal/controller/podplacement/architecture_application.go
```diff
@@ -0,0 +1,115 @@
...
+// is used by tests to verify the function was called, but production callers (webhook
+// and controller) ignore the return value since they always proceed with pod processing
+// after calling this function.
+func applyArchitectureConstraints(pod *corev1.Pod, architectures []string) bool {
+	if len(architectures) == 0 {
+		return false
+	}
+
+	// Remove architecture from nodeSelector (this is safe and doesn't cause API rejections)
+	removeArchitectureFromNodeSelector(pod)
```
AnnaZivkovic: How do we know this is safe? What would this removed user defined fields?

internal/controller/podplacement/architecture_removal.go
```diff
@@ -0,0 +1,115 @@
...
+	return false
+}
+
+// removeArchitectureFromNodeAffinity removes architecture-based match expressions from
+// requiredDuringSchedulingIgnoredDuringExecution node affinity.
+// It removes matchExpressions with key kubernetes.io/arch from all node selector terms.
+// Empty node selector terms are removed after cleanup.
+// preferredDuringSchedulingIgnoredDuringExecution is preserved as per enhancement doc.
+// Returns true if any architecture constraints were removed, false otherwise
+func removeArchitectureFromNodeAffinity(pod *corev1.Pod) bool {
```
AnnaZivkovic, line 46: Could we move or delete removeArchitectureFromNodeAffinity and removeAllArchitectureConstraints? I think they are only for testing? 

Move to pkg/testing/builder/ or a new pkg/testing/helpers/ package. This would require exporting the function names (RemoveArchitectureFromNodeAffinity, etc.) and importing utils.ArchLabel. Downside it spreads architecture-manipulation logic across packages.

internal/controller/podplacement/scheduling_gate_mutating_webhook.go
```diff
@@ -165,6 +171,60 @@ func (a *PodSchedulingGateMutatingWebHook) delayedSchedulingGatedEvent(ctx conte
...
+	for _, ppc := range matchingPPCs {
+		if !ppc.PluginsEnabled(common.CelArchitecturePlacementPluginName) {
+			continue
+		}
+
+		celPlugin := ppc.Spec.Plugins.CelArchitecturePlacement
+		if celPlugin == nil {
+			// This should never happen due to webhook validation, but handle defensively
+			log.Error(nil, "CEL plugin enabled but configuration is nil", "PodPlacementConfig", ppc.Name, "pod", pod.Name)
+			return
```
AnnaZivkovic: The soft failure model is used everywhere else so why use return here?

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,244 @@
...
+	e.cache[expression] = prog
+	e.mu.Unlock()
+
+	return prog, nil
+}
+
+// podToMap converts a Pod to a map structure that CEL can evaluate
+func podToMap(pod *corev1.Pod) map[string]interface{} {
+	if pod == nil {
+		return map[string]interface{}{
```
AnnaZivkovic, line 126: I think pods can have an empty name and a populated generateName at admission time. If a user writes a CEL expression matching by name prefix for these pods, it  won't work because name is empty and generateName isn't in the map. Could we consider adding generateName  to the exposed metadata, or document this?

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,244 @@
...
+	val, _, err := prog.Eval(map[string]interface{}{
+		"self": podMap,
+	})
+	if err != nil {
+		// Evaluation errors are treated as non-matches per enhancement doc
+		return false, fmt.Errorf("CEL evaluation error: %w", err)
+	}
+
+	// Convert result to boolean
+	result := ref.Val(val)
```
AnnaZivkovic: I think val returned by prog.Eval is already a ref.Val?

internal/controller/podplacementconfig/podplacementconfig_webhook.go
```diff
@@ -94,6 +113,31 @@ func (w *PodPlacementConfigWebhook) Handle(ctx context.Context, req admission.Re
 	}
 }
 
+// validateCELExpressionAtAdmission validates a CEL expression at admission time
+// This ensures that invalid CEL expressions are rejected before they can cause runtime errors
+func validateCELExpressionAtAdmission(expression string) error {
```
AnnaZivkovic: Could we move the validation logic onto the plugin type, then call it from the webhook? I think this would be more consistent in keeping all plugin validation self-contained on the type

go.mod
```diff
@@ -84,7 +84,7 @@ require (
 	github.com/go-openapi/swag/typeutils v0.27.0 // indirect
 	github.com/go-openapi/swag/yamlutils v0.27.0 // indirect
 	github.com/go-task/slim-sprig/v3 v3.0.0 // indirect
-	github.com/google/cel-go v0.28.1 // indirect
+	github.com/google/cel-go v0.28.1
```
AnnaZivkovic: if it is not indirect any more can we move it to the first block in the go.mod?

AnnaZivkovic (commented):
The overall structure is solid and the core flow is there. Nice job on the test coverage
  breadth as well.

I've left some inline comments on areas that diverge from the enhancement document or from existing codebase patterns. The main ones to look at are the metadata-only enforcement, the duplicate webhook/controller CEL evaluation, and the logging levels.

Before the next review round, please get the CI checks passing (unit tests, lint, and vet are all currently failing). It's hard to fully evaluate the implementation with red CI. Thanks!

internal/controller/podplacementconfig/podplacementconfig_webhook.go
pranitaT: Implemented. ValidateCELExpressions() now lives on CelArchitecturePlacement and the webhook calls it.

internal/controller/podplacement/pod_model.go
pranitaT: Restored the logging back to the original verbosity (V(3)), so these debug traces no longer appear at the default log level.

internal/controller/podplacement/pod_model.go
pranitaT: Reverted the image inspection logs back to verbose (V) logging to preserve the existing logging behavior.

go.mod
pranitaT: Moved _github.com/google/cel-go_ to the direct dependency block in go.mod.

internal/controller/podplacement/cel_evaluator.go
pranitaT: Removed the unnecessary ref.Val() conversion since prog.Eval() already returns a ref.Val.

internal/controller/podplacement/cel_evaluator.go
pranitaT: Added metadata.generateName to the CEL evaluation context so admission-time pods created via generateName can be matched correctly.

internal/controller/podplacement/scheduling_gate_mutating_webhook.go
pranitaT: Following the existing soft-failure model by continuing processing instead of exiting early.

internal/controller/podplacement/scheduling_gate_mutating_webhook.go
```diff
@@ -77,17 +80,30 @@ func (a *PodSchedulingGateMutatingWebHook) Handle(ctx context.Context, req admis
...
+	// This avoids missing CEL placement decisions during informer startup or cache
+	// propagation, when later reconciliation cannot reliably add new required
+	// NodeSelectorTerms.
 	ppcList := &multiarchv1beta1.PodPlacementConfigList{}
 	if err := a.client.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
 		log.Error(err, "Failed to list existing PodPlacementConfigs in namespace")
 		// On error, proceed without PPC filtering - fail open
 		ppcList.Items = []multiarchv1beta1.PodPlacementConfig{}
 	}
+	if len(ppcList.Items) == 0 && a.apiReader != nil {
```
AnnaZivkovic: What do you think about gating the apiReader to fallback on informer sync state?  I am worried this will cause lots of traffic on a cluster. 

Could we add  the PPC informer's `HasSynced` signal into the webhook and only fall back before the cache has synced:

```
  type PodSchedulingGateMutatingWebHook struct {
        client         client.Client
        apiReader      client.Reader
        ppcCacheSynced func() bool
        // ...
  }

  if len(ppcList.Items) == 0 && a.apiReader != nil && !a.ppcCacheSynced() {
        // ...
  }
```

api/common/plugins/celarchitectureplacement_plugin.go
```diff
@@ -0,0 +1,265 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	Name string `json:"name" protobuf:"bytes,1,opt,name=name"`
+
+	// Expression is a CEL expression that evaluates against a Pod resource.
+	// The expression must return a boolean value.
+	// The expression has access to the pod via the 'self' variable.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
AnnaZivkovic, line 130:
```suggestion
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:MaxLength=1024
```
What do you think about adding a Max length ?

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,260 @@
...
+// Errors from compilation or runtime evaluation are returned to the caller;
+// it is the caller's responsibility to decide how to handle them (e.g. skip or abort).
+func (e *celEvaluator) evaluate(expression string, pod *corev1.Pod) (bool, error) {
+	prog, err := e.compile(expression)
+	if err != nil {
+		return false, err
+	}
+
+	// Convert pod to a map structure for CEL evaluation
+	podMap := podToMap(pod)
```
AnnaZivkovic: Would it be possible to build podToMap once per pod, not once per rule?

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,260 @@
...
+			// block all subsequent rules. The caller receives a non-error return (nil, "", nil)
+			// only when *all* rules are exhausted; a per-rule skip is intentional.
+			continue
+		}
+		if matched {
+			// First match wins; remaining rules are not evaluated.
+			return rule.Architectures, rule.Name, nil
+		}
+	}
+	return nil, "", nil
```
AnnaZivkovic: I think there is a bug here when malformed CEL rules in high-priority PPC silently apply fallback architectures which block lower-priority PPCs.

Reproduction: A namespace has two PPCs with CEL enabled:
  - Priority 200: malformed CEL expression, fallback ["amd64"]
  - Priority 100: valid CEL expression matching the pod, architectures ["ppc64le"]

So what we want is for the malformed PPC is skipped and the pod gets ppc64le from the valid PPC. What is happening is that the pod gets amd64 from the malformed PPC's fallback.

internal/controller/podplacement/cel_evaluator.go
pranitaT: Addressed this by adding an assertMetadataOnly AST validation check during CEL expression validation. It walks the CEL AST and rejects references to self.spec.* and self.status.* during admission.

I also added regression tests covering these cases to ensure metadata-only access is enforced.

pranitaT: Added a typed CEL validation environment using ObjectType/DeclTypeProvider for the supported Pod metadata fields. This provides compile-time validation and catches invalid field references such as self.metadata.labells.

The metadata-only AST validation is also applied on top of the typed environment to explicitly reject valid Pod fields such as self.spec.* and self.status.*, as required by the enhancement.

I added unit tests for both invalid field references and prohibited spec/status access.

internal/controller/podplacement/cel_integration.go
pranitaT: Addressed. CEL-processed pods now set NodeAffinityLabel to overriden when architecture constraints are successfully applied.

This is handled consistently in both the webhook and controller paths, and the label is only updated when architecture constraints are actually applied.

I also added regression coverage for the CEL match, error, and fallback paths to verify the label is set correctly.

internal/controller/podplacement/architecture_application.go
pranitaT: I updated the architecture constraint handling to remove only the architecture-specific constraints while preserving user-defined/non-architecture affinity and node-selector fields.

Added regression coverage to verify that existing user-defined affinity constraints are preserved.

internal/controller/podplacement/architecture_removal.go
pranitaT: I rechecked the call sites. These helpers are part of the production architecture-constraint application flow and are not test-only, so I kept them in the production package.

internal/controller/podplacement/scheduling_gate_mutating_webhook.go
pranitaT: Added the PPC informer HasSynced state to the webhook. The apiReader fallback is now used only while the PPC informer cache has not synced, avoiding unnecessary API-server requests after the cache is ready.

Added tests covering synced, unsynced, nil sync state, and nil apiReader cases.

internal/controller/podplacement/cel_evaluator.go
pranitaT: podToMap() is now called once per pod before evaluating the rules, and the resulting map is reused for all CEL rules for that pod. This avoids rebuilding the same map for every rule.

internal/controller/podplacement/cel_evaluator.go
pranitaT: Updated CEL rule evaluation to distinguish between no matching rule and all rules failing to evaluate.

If all CEL rules in a higher-priority PPC are malformed, that PPC is skipped without applying its fallback architectures, allowing the next lower-priority PPC to be evaluated.

Added a regression test for the scenario where a high-priority malformed PPC must not block a lower-priority valid PPC.

internal/controller/podplacement/pod_reconciler.go
```diff
@@ -141,8 +163,9 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 
 	// Skip preferred affinity processing if the user has already configured architecture-related preferred affinity
 	// or if the reconcile loop has already applied the PPCs/CPPC (e.g., due to a retry or re-reconciliation)
+	celApplied := false
 	if !pod.isPreferredAffinityConfiguredForArchitecture() {
```
AnnaZivkovic:
```suggestion
celApplied := r.applyMatchingPPCs(ctx, matchingPPCs, pod)
if !celApplied && !pod.isPreferredAffinityConfiguredForArchitecture() {
    // NodeAffinityScoring only
}
```

It appears that `applyMatchingPPCs` -> `applyCELArchitecturePlacement` only runs when the pod has no user-defined preferred ` kubernetes.io/arch` affinity. Pods with preferred arch terms skip reconciler CEL entirely, even though we would want to set the `RequiredDuringSchedulingIgnoredDuringExecution` term.

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/distribution/distribution/v3 v3.1.1
 	github.com/go-logr/logr v1.4.3
 	github.com/go-logr/zapr v1.3.0
+	github.com/google/cel-go v0.28.1
 	github.com/google/uuid v1.6.0
```
AnnaZivkovic, line 13:
```suggestion
	github.com/google/uuid v1.6.0
	gomodules.xyz/jsonpatch/v2 v2.5.0
	k8s.io/apiserver v0.36.2
```

can we move all direct dependence to the first block?

internal/controller/podplacement/architecture_application.go
```diff
@@ -0,0 +1,119 @@
...
+func applyArchitectureConstraints(pod *corev1.Pod, architectures []string) bool {
+	if len(architectures) == 0 {
+		return false
+	}
+
+	// Remove all existing architecture constraints first so that stale arch-only
+	// NodeSelectorTerms (set by a previous MTO invocation) are cleaned up before
+	// the new architecture is applied.  This avoids accumulating duplicate terms
+	// across repeated invocations and ensures the fresh arch replaces the stale one.
+	removeAllArchitectureConstraints(pod)
```
AnnaZivkovic, line 111: `NodeSelectorTerms` shrinking  safe at admission,  but seems risky on reconcile if the webhook failed to apply. 

KEP-3838 does not allow changing the length of nodeSelectorTerms on Update it is only for in-place edits to matchExpressions inside existing terms are safe. This path calls applyArchitectureConstraints, which runs removeAllArchitectureConstraints.  That can strip kubernetes.io/arch from each term and deletes any term that becomes empty (e.g. an arch-only OR branch), which can shrink the list from 2 terms to 1. 

Reconcile -> processPod  ->  applyMatchingPPCs  ->  applyCELArchitecturePlacement  ->  applyArchitectureConstraints  ->  removeAllArchitectureConstraints  ->  r.Update(pod)

That is fine when the webhook already applied CEL at admission (pod already has the merged shape), but if the webhook did not run CEL and the persisted pod still has multiple terms (e.g. arch-only + zone), reconcile can try to send fewer terms than etcd has and the Update can fail with HTTP 422, leaving the pod gated.

This is low risk but probably worth fixing 

Could we onsider splitting into:
- `applyArchitectureConstraintsAtAdmission` - current `removeAll + apply` (webhook only)
- `applyArchitectureConstraintsInPlace` - `removeArchitectureFromNodeSelector` +
  `applyArchitectureNodeAffinity` only (reconciler fallback, no term deletion)

internal/controller/podplacement/cel_integration.go
```diff
@@ -0,0 +1,108 @@
...
+			"PodPlacementConfig", ppc.Name,
+			"pod", pod.Name,
+			"architectures", result.architectures)
+	}
+
+	// Remove existing architecture constraints and apply new ones.
+	// If no architectures were produced (empty/nil result), do not claim the plugin
+	// applied — fall through to image-based detection so the pod is not ungated
+	// without any architecture constraint.
+	if !applyArchitectureConstraints(pod.PodObject(), result.architectures) {
```
AnnaZivkovic: The reconciler path calls the same `applyArchitectureConstraints` as the webhook, including
`removeAllArchitectureConstraints` which can shrink `nodeSelectorTerms` in the case that the webhook failed to apply changes. 

Could we gate at the top of `applyCELArchitecturePlacement` (or in `processPod`)?

cmd/main.go
```diff
@@ -269,7 +270,38 @@ func RunClusterPodPlacementConfigOperandWebHook(mgr ctrl.Manager) {
...
+	// so we resolve it on the first call rather than at registration time.
+	// Until the informer is resolved, the function returns false (not yet synced),
+	// which keeps the apiReader fallback enabled — the safe default at startup.
+	var (
+		ppcHasSynced func() bool
+		ppcOnce      sync.Once
+	)
+	ppcCacheSynced := func() bool {
+		ppcOnce.Do(func() {
+			informer, err := mgr.GetCache().GetInformerForKind(
```
AnnaZivkovic: I think it is possible for  `GetInformerForKind` to fail inside `ppcOnce.Do`, `ppcHasSynced` stays nil and `ppcCacheSynced()` always returns false so apiReader will fallback on every pod in namespaces with zero cached PPCs, forever (no retry). Could we consider retrying informer lookup on subsequent calls instead of `sync.Once`, or treating informer failure as synced after logging to avoid permanent fallback?

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,326 @@
...
+	"github.com/google/cel-go/common/types"
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/openshift/multiarch-tuning-operator/api/common/plugins"
+)
+
+// celEvaluator handles CEL expression compilation, caching, and evaluation
+type celEvaluator struct {
+	env   *cel.Env
+	cache map[string]cel.Program
```
AnnaZivkovic: Compiled expressions are cached indefinitely with no eviction. With up to 1000 rules per
PPC and PPC revisions over time, memory grows unbounded in a long-running operator pod.
Can we switch to an LRU?

pranitaT: Addressed this by replacing the unbounded map with a bounded LRU cache. The cache is limited to 1024 compiled CEL programs and remains protected by a mutex for concurrent access. Added regression coverage to verify eviction once the capacity is exceeded.

cmd/main.go
pranitaT: Addressed this by replacing the one-time informer lookup with retryable synchronization. If GetInformerForKind fails transiently, subsequent calls can retry instead of permanently falling back to the API reader. Added a regression test covering the fail-then-success case.

internal/controller/podplacement/cel_integration.go
pranitaT: Updated the CEL reconcile path to use the in-place architecture update logic instead of removeAllArchitectureConstraints. This prevents NodeSelectorTerms from being removed or recreated during reconciliation.

internal/controller/podplacement/architecture_application.go
pranitaT: Addressed this by keeping the admission and reconcile paths separate. The reconcile path now updates the existing NodeSelectorTerms in place and does not remove terms, avoiding the KEP-3838 update restriction. Added regression coverage for the term-count preservation scenario.

go.mod
pranitaT: Moved the directly used dependencies to the direct require block in go.mod, without changing their versions.

internal/controller/podplacement/pod_reconciler.go
pranitaT: Addressed this by ensuring applyMatchingPPCs runs before the preferred-affinity check. CEL can now apply the required architecture affinity even when the pod already has user-defined preferred architecture affinity, while the existing preferred-affinity handling remains unchanged. Added regression coverage for this scenario.

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,336 @@
...
+// evaluateResult represents the result of CEL rule evaluation
+type evaluateResult struct {
+	architectures []string
+	ruleName      string
+	matched       bool
+	// allRulesErrored is true when every rule in the PPC failed to compile or
+	// evaluate.  Both the webhook and the controller skip the malformed PPC
+	// (return false) so that lower-priority PPCs can still claim the pod.
+	// The fallback architectures of the malformed PPC are never applied.
+	allRulesErrored bool
```
AnnaZivkovic, line 272: Is it an issue that webhook and controller handle this field differently?

I think the webhook skips the PPC entirely so no architecture constraints are written and lower-priority PPCs get a chance to claim the pod. The controller applies fallback architectures from the malformed PPC.

api/common/plugins/celarchitectureplacement_plugin.go
```diff
@@ -0,0 +1,304 @@
...
+// selectPath reconstructs the dotted field path for a select expression node
+// by walking the operand chain back to the root identifier.
+// Returns the path segments from root to leaf, e.g. ["self", "spec", "containers"].
+// Returns nil if the root is not an identifier (e.g. a function call result).
+func selectPath(node celast.NavigableExpr) []string {
+	var segments []string
+	cur := celast.Expr(node)
+	for cur.Kind() == celast.SelectKind {
+		sel := cur.AsSelect()
+		segments = append([]string{sel.FieldName()}, segments...)
```
AnnaZivkovic, line 296: How would this handle something like `self.metadata.labels['spec']` ?

I think the select-only AST walk misses Index nodes

internal/controller/podplacement/cel_reliability_test.go
```diff
@@ -0,0 +1,319 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/openshift/multiarch-tuning-operator/api/common/plugins"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
+)
+
+var _ = Describe("CEL Reliability", func() {
+
+	// TestFirstMatchWinsStrictOrdering
+	It("should evaluate rules strictly in order and stop at first match", func() {
```
AnnaZivkovic: There are a lot of duplicate tests  in cel_reliability_test.go and  cel_evaluator_test.go. Can we clean them up?

internal/controller/podplacement/cel_new_critical_tests_test.go
```diff
@@ -0,0 +1,495 @@
...
+		recorder := record.NewFakeRecorder(8)
+
+		original := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{Name: "test-pod", Namespace: "default"},
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{{Name: "c", Image: "nginx:latest"}},
+			},
+		}
+
+		ppcs := []v1beta1.PodPlacementConfig{
```
AnnaZivkovic, line 58: Can we define objects using this style? 

```suggestion
ppc := NewPodPlacementConfig().
					WithName("test-ppc-user-conflict").
					WithNamespace(ns.Name).
					WithPriority(100).
					WithNodeAffinityScoring(true).
					WithNodeAffinityScoringTerm(utils.ArchitecturePpc64le, 50).
					Build()
Expect(k8sClient.Create(ctx, ppc)).To(Succeed())
```

internal/controller/podplacement/cel_new_critical_tests_test.go
```diff
@@ -0,0 +1,495 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+
+	"github.com/openshift/multiarch-tuning-operator/api/common/plugins"
+	"github.com/openshift/multiarch-tuning-operator/api/v1beta1"
+	"github.com/openshift/multiarch-tuning-operator/pkg/utils"
+)
+
+var _ = Describe("CEL New Critical Tests", func() {
```
AnnaZivkovic, line 43: `cel_new_critical_tests_test.go`  Tests Webhook Path, but I am not seeing any for the Controller Path.
The tests `TestApplyCELInWebhook_PluginDisabled_NoModification` and `TestApplyCELInWebhook_PluginDisabled_ExistingAffinityUnchanged` only exercise `applyCELInWebhook`. There are no equivalent tests for the controller path (`applyCELArchitecturePlacement`) verifying that a disabled plugin is a no-op.

pranitaT: Added direct tests for applyCELArchitecturePlacement() with the plugin disabled, covering the no-op behavior, existing affinity preservation, and preservation of unrelated pod fields.

internal/controller/podplacement/cel_new_critical_tests_test.go
pranitaT: Updated the applicable tests to use NewPod() / NewPodPlacementConfig(). Explicit structs were retained only where the builder cannot represent the required fields or where they make the test intent clearer.

internal/controller/podplacement/cel_reliability_test.go
pranitaT: Removed the duplicate reliability tests after confirming equivalent coverage exists in cel_evaluator_test.go. No unique behavior coverage was lost.

api/common/plugins/celarchitectureplacement_plugin.go
pranitaT: Added coverage for bracket notation, including self.metadata.labels['spec'], also verified the CEL type-checking behavior to ensure index access cannot bypass the allowed field restrictions.

internal/controller/podplacement/cel_evaluator.go
pranitaT: Verified that both webhook and controller paths skip a PPC when all CEL rules fail and do not apply its fallback architectures. Lower-priority PPCs can still claim the pod, with coverage at both unit and regression levels.

api/common/plugins/celarchitectureplacement_plugin.go
```diff
@@ -0,0 +1,304 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/architecture_application.go
```diff
@@ -0,0 +1,119 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/architecture_application_test.go
```diff
@@ -0,0 +1,504 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/architecture_removal.go
```diff
@@ -0,0 +1,115 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/architecture_removal_test.go
```diff
@@ -0,0 +1,391 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_evaluator_test.go
```diff
@@ -0,0 +1,1321 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_coexistence_test.go
```diff
@@ -0,0 +1,174 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_controller_integration_test.go
```diff
@@ -0,0 +1,612 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_evaluator.go
```diff
@@ -0,0 +1,336 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_integration.go
```diff
@@ -0,0 +1,115 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_mutation_safety_test.go
```diff
@@ -0,0 +1,467 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_new_critical_tests_test.go
```diff
@@ -0,0 +1,475 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_no_fallback_merge_test.go
```diff
@@ -0,0 +1,293 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_ppc64le_default_test.go
```diff
@@ -0,0 +1,469 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_preservation_test.go
```diff
@@ -0,0 +1,216 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_reliability_test.go
```diff
@@ -0,0 +1,139 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_stale_arch_regression_test.go
```diff
@@ -0,0 +1,192 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_test_helpers_test.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_webhook_handle_test.go
```diff
@@ -0,0 +1,508 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/regression_test.go
```diff
@@ -0,0 +1,651 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/webhook_cel_test.go
```diff
@@ -0,0 +1,485 @@
+/*
+Copyright 2025 Red Hat, Inc.
```
prb112:
```suggestion
Copyright 2026 Red Hat, Inc.
```

internal/controller/podplacement/cel_evaluator.go
prb112: I do think this part is worth documenting as the behavior should be clear to the customer).

prb112 (commented):
/lgtm

PR #752, opened 2026-07-06
MULTIARCH-6239: use direct API reader for PPC listing in PodReconciler

AnnaZivkovic:
The PodReconciler listed PodPlacementConfigs through the informer cache, which has inherent delivery latency (~5-50ms). When a PPC was created right before a pod was reconciled and image inspection completed instantly (cached), the pod could be fully processed and ungated before the PPC appeared in the cache — permanently missing its preferred affinity.

Switch the PPC list call from the cached client (r.List) to a direct API server read (r.APIReader.List) so every reconciliation sees the latest PPCs regardless of informer cache state.

### DIFF
--- a/internal/controller/podplacement/pod_reconciler.go
+++ b/internal/controller/podplacement/pod_reconciler.go
@@ -46,6 +46,7 @@ import (
 // PodReconciler reconciles a Pod object
 type PodReconciler struct {
 	client.Client
+	APIReader client.Reader
 	Scheme    *runtime.Scheme
 	ClientSet *kubernetes.Clientset
 	Recorder  record.EventRecorder
@@ -114,6 +115,15 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 	if err := r.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
 		pod.handleError(err, "failed to list existing PodPlacementConfigs in namespace")
 		return
+	}
+	// The informer cache can lag behind the API server. If the cache shows no PPCs,
+	// verify with a direct API read to avoid a race where a just-created PPC is missed
+	// and the pod is permanently ungated without its preferred affinity.
+	if len(ppcList.Items) == 0 {
+		if err := r.APIReader.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
+			pod.handleError(err, "failed to list PodPlacementConfigs from API server")
+			return
+		}
 	}
 
 	// Filter to only PPCs that match this pod's labels - do this once for efficiency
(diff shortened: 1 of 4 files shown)


internal/controller/podplacement/pod_reconciler.go
```diff
@@ -111,7 +112,7 @@ func (r *PodReconciler) processPod(ctx context.Context, pod *Pod) {
 	cppc := clusterpodplacementconfig.GetClusterPodPlacementConfig()
 	// List existing PodPlacementConfigs in the same namespace
 	ppcList := &multiarchv1beta1.PodPlacementConfigList{}
-	if err := r.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
+	if err := r.APIReader.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
```
jcpowermac: **Suggestion: cache-first, API-fallback**

_Review by Claude (AI-assisted) on behalf of @jcallen_

The informer cache can lag (miss new items) but never shows phantom items — if the cache returns PPCs, they're real. The dangerous case is when the cache shows **zero** PPCs due to lag, which is exactly the race described in this PR. A hybrid approach gives correctness without always paying the API cost:

```suggestion
	if err := r.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
		pod.handleError(err, "failed to list existing PodPlacementConfigs in namespace")
		return
	}
	// The informer cache can lag behind the API server. If the cache shows no PPCs,
	// verify with a direct API read to avoid a race where a just-created PPC is missed
	// and the pod is permanently ungated without its preferred affinity.
	if len(ppcList.Items) == 0 {
		if err := r.APIReader.List(ctx, ppcList, client.InNamespace(pod.Namespace)); err != nil {
			pod.handleError(err, "failed to list PodPlacementConfigs from API server")
			return
		}
	}
```

This limits direct API server calls to only when the race actually matters (no PPCs visible in cache). When the cache already shows PPCs, the API call is skipped entirely.

**Tradeoff**: if a new PPC is created in a namespace that already has other cached PPCs, the newest one could still be missed. But the pod would still get preferred affinity from the existing PPCs — a much less impactful case than missing all PPCs entirely.

jcpowermac (commented):
## Review

_Review by Claude (AI-assisted) on behalf of @jcallen_

The race is real and the fix is correct — once the scheduling gate is removed there's no second chance to apply preferred affinity. However, this call runs on **every gated pod reconciliation**, replacing a zero-cost local cache read with an API server round-trip. With `MaxConcurrentReconciles = NumCPU * 4`, burst pod creation (large rollouts) could generate significant API server load.

See inline comment for a suggested hybrid approach.

